PR #157, opened 2014-10-11
Prompt for tor controller password if needed

thedod:
We use getpass for this (i.e. text mode).
This means that if you have password protection for your
Tor controller (advised) and you want to run onionshare-gui,
you need to do it from a terminal (so that you can enter
the password).


Silur:
so now that onionshare uses ephemeral services, how about giving the ability to handle control port authentication? I won't leave my tor config unauthorized for just one program.

PR #202, opened 2015-06-16
add build instructions for archlinux to BUILD.md

simonsmiley:
just a few lines about building and installing on archlinux and any other pacman based distro


micahflee:
:+1:

PR #217, opened 2015-11-07
Clean up and exit on Ctrl-Q

effigies:
Expected GUI behavior. Particularly helpful if using window managers that don't provide a close button.


micahflee:
Thanks!

PR #226, opened 2015-11-21
Add Czech translation

Venca24:
I've translated OnionShare into Czech language. The translation is not 100 % but it is useful, especially for newbies which use GUI.


micahflee:
Thank you!

PR #229, opened 2015-11-27
OnionShare translated into Esperanto language.

Venca24:
I've translated OnionShare into Esperanto. The translation is not 100 % but it is still useful mainly for GUI users.


micahflee:
Thanks!

PR #244, opened 2015-12-23
Remove unnecessary status_bar.showMessage

### LINKED ISSUE #242: "Start Sharing" on Mac appears to hang for ~35s
# STR

1. Open OnionShare.app
2. Add a file to upload
3. Click "Start Sharing"

## Expected behavior

Some type of progress indicator (messages, etc.) shows us progress as the hidden service is created, detected as being available, etc. The app remains responsive and communicative until sharing starts.

## Actual behavior

The app _appears_ to hang. The "Start Sharing" button stays highlighted for about 35s, with no messages or progress indicator from the app. After a few seconds, the black pointer cursor changes to the "spinning beachball of doom", which usually indicates that an app is hanging.

After about 35s, the beachball changes back into a cursor, the green status icon changes to yellow, and the "Start Sharing" button changes to "Please Wait...". This should probably happen from the beginning, and it would be helpful to display some status messages as OnionShare sets up.

Here's a [video](https://youtu.be/XvGjoab0MDc) that demonstrates the behavior. Unfortunately, due to a limitation of Quicktime Player's screen recording feature, you can't see the cursor change into a beachball (but it does after 5 seconds).

---

I've only observed this behavior on Mac, and have not tested to see a if similar thing happens on other platforms.

garrettr:
status_bar.showMessage is already called with the same message at the
beginning of start_server, so there is no reason to call it again in the
try block.

### DIFF
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -147,7 +147,6 @@ class OnionShareGui(QtGui.QWidget):
         self.status_bar.showMessage(strings._('gui_starting_server1', True))
         self.app.choose_port()
         try:
-            self.status_bar.showMessage(strings._('gui_starting_server1', True))
             self.app.start_hidden_service(gui=True)
         except onionshare.hs.NoTor as e:
             alert(e.args[0], QtGui.QMessageBox.Warning)


garrettr:
I was wrong about this fixing #242, see https://github.com/micahflee/onionshare/issues/242#issuecomment-166793610 for the root cause. However, I still think it should be merged since the second call to `showMessage` is redundant.

I'm rebasing this PR to remove the reference to fixing #242 from the commit message.

garrettr:
Rebased.

PR #246, opened 2015-12-23
Add estimated time remaining to progress indicator

garrettr:
Estimates the time remaining for each download and displays it in the
progress bar. Waits for 10 seconds before showing the progress bar to
allow the download rate to stabilize, which prevents the estimated time
remaining from jumping all over the place at the start of the download
(a.k.a the "Windows copy dialog experience"). If your download takes
less than 10 seconds, you don't really need to see an ETA anyway.

This commit also refactors the Downloads class, splitting out the
download-specific functionality into a new Download class, providing
better encapsulation. As a result, I was able to simplify the call to
`update_download` because it was no longer necessary to pass the
`total_bytes` (which don't change after the download has begun).

Tested on Mac OS 10.9.

### DIFF
--- a/onionshare/helpers.py
+++ b/onionshare/helpers.py
@@ -19,6 +19,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import os, inspect, hashlib, base64, hmac, platform, zipfile, tempfile
 from itertools import izip
+import math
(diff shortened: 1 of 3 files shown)


garrettr:
Micah says: make sure the strings are localized correctly

PR #256, opened 2016-01-06
Update desktop entry as suggested by Debian lintian tool

infinity0:
- Remove UTF-8 entry - it is now redundant; see http://standards.freedesktop.org/desktop-entry-spec/latest/apc.html
- Add some suggested keywords; see http://standards.freedesktop.org/desktop-entry-spec/latest/ar01s05.html


micahflee:
Thanks!

PR #269, opened 2016-04-05
Minor fixes (proposal)

kant:


micahflee:
Merging directly had conflicts, so I just made a new manual commit.

PR #288, opened 2016-05-25
Fix CVE-2016-5026

mscherer:
See http://www.openwall.com/lists/oss-security/2016/05/23/5 for details
on a potential convoluted attack. Basically,  /tmp/onionshare is
a predictable name, which mean that a local attacker
could precreate it on a shared server and later mess with the hidden
service operations in various way.


u451f:
Thanks @mscherer. i'll prepare a patch for the Debian package tomorrow.

u451f:
Also see https://github.com/micahflee/onionshare/issues/258

Issue #290, opened 2016-05-26
building 0.9.1 fails due to python dependancies 


freddymartinez9:
Hi Micah,

The current build instructions on debian jessie fail because python3-nose isn't listed as a dependency. stacktrace below: 

```
dpkg-checkbuilddeps: Unmet build dependencies: python3-nose
dpkg-buildpackage: warning: build dependencies/conflicts unsatisfied; aborting
dpkg-buildpackage: warning: (Use -d flag to override.)
Traceback (most recent call last):
  File "setup.py", line 103, in <module>
    (os.path.join(sys.prefix, 'share/onionshare/html'), html)
  File "/usr/lib/python3.4/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/lib/python3.4/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3.4/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/usr/lib/python3/dist-packages/stdeb/command/bdist_deb.py", line 48, in run
    util.process_command(syscmd,cwd=target_dirs[0])
  File "/usr/lib/python3/dist-packages/stdeb/util.py", line 169, in process_command
    check_call(args, cwd=cwd)
  File "/usr/lib/python3/dist-packages/stdeb/util.py", line 45, in check_call
    raise CalledProcessError(retcode)
stdeb.util.CalledProcessError: 3
```

u451f:
Hi Freddy,

if you want to you can copy the dependencies from the official Debian package: 
https://anonscm.debian.org/cgit/pkg-privacy/packages/onionshare.git/tree/debian/control?h=debian

It builds correctly but it's hit by this bug: https://github.com/micahflee/onionshare/issues/284
That's why the new version is not yet in Debian.

Cheers!

Issue #284, opened 2016-04-18
Make sure tests can find correct paths


micahflee:
It looks like tests fail when running from certain places, e.g.

```

I: pybuild base:184: cd
/build/onionshare-0.9/.pybuild/pythonX.Y_3.5/build; python3.5 -m nose test
..EE..
======================================================================
ERROR: load_strings() loads English by default
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File
"/build/onionshare-0.9/.pybuild/pythonX.Y_3.5/build/test/onionshare_strings_test.py",
line 33, in test_load_strings_defaults_to_english
    strings.load_strings()
  File
"/build/onionshare-0.9/.pybuild/pythonX.Y_3.5/build/onionshare/strings.py",
line 40, in load_strings
    for filename in os.listdir(locale_dir):
FileNotFoundError: [Errno 2] No such file or directory:
'/build/onionshare-0.9/.pybuild/pythonX.Y_3.5/build/resources/locale'

======================================================================
ERROR: load_strings() loads other languages in different locales
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File
"/build/onionshare-0.9/.pybuild/pythonX.Y_3.5/build/test/onionshare_strings_test.py",
line 40, in test_load_strings_loads_other_languages
    strings.load_strings("fr")
  File
"/build/onionshare-0.9/.pybuild/pythonX.Y_3.5/build/onionshare/strings.py",
line 40, in load_strings
    for filename in os.listdir(locale_dir):
FileNotFoundError: [Errno 2] No such file or directory:
'/build/onionshare-0.9/.pybuild/pythonX.Y_3.5/build/resources/locale'
```

u451f:
That's what currently prevents us from uploading the new version to Debian, if anybody wants to help :)

dballard:
So I don't seem to have this problem in Debian 8 Jessie. 
```
$ python3 --version
Python 3.4.2
$ python3 -m nose --version
python3 -m nose version 1.3.4
$ dpkg -k python3-stdeb
...
ii python3-stdeb 0.8.2-4
```

For me pybuild works, but seems to report 

```
....
Ran 0 tests in 0.000s

OK
```
[...]

PR #294, opened 2016-06-17
add python3-nose for ubuntu 16.04 to Build.md

dballard:
on Ubuntu 16.04 python3-nose was required to proceed (on 14.04 python3-stdeb was just missing entierly)
Don't know if you want to amend the instructions. Feel free to reject :) 
(tho seems like it's required for the tests as you mention below so probably required here for Debian as well)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -27,7 +27,7 @@ After that you can try both the CLI and the GUI version of OnionShare:
 A script to build a .deb package and install OnionShare easily is also provided for your convenience:
 
 ```sh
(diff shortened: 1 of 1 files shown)


freddymartinez9:
related (duplicate?) of #284 and #290

dballard:
to #290.
hm in stdeb.cfg python3-nose is listed as a build dep but not a dependancy. maybe adding it to that will fix that problem? or maybe it was fixed but now #284 is the blocker?

Issue #177, opened 2015-04-21
have a set time to turn off web sever


ghost:
Can there be an option to turn off the web server for a set amount of time, for example you only want the download link available for 24 hours.

micahflee:
Seems like a good idea.

Zaxscdvf:
Do you think you'll add in this feature?

micahflee:
I'd like to, but not in the immediate release I'm working on. I'm just gonna focus on bug fixes for now, and decide which new features to build in later for future releases. So, quite possibly, but not yet :). If you'd like to try implementing it yourself though, go ahead and create a pull request. I'm definitely open to merging it in.

Zaxscdvf:
ok thanks I'll see what I can do.

choltz95:
Is there a preference for the gui layout? i.e. replacing the check box with a drop-down list or textfield?

micahflee:
I changed this to the 0.9.3 milestone because I thought the #296 PR would be easy to merge. But it still needs more work, so I'm changing back to 1.0 again.

micahflee:
When #473 is finished, it will resolve this issue

mig5:
\o/

PR #296, opened 2016-07-08
Time to close

choltz95:
Implement enhancement #177 
- added timer daemon in helpers.py (close_after_seconds())
- added thread instance variable in onionshare class (t_cas)
- added QLineEdit field to gui options alongside previous checkbox
- stay_open bool -> integer; false -> 0 & empty gui field + checkbox = no time limit

### DIFF
--- a/onionshare/helpers.py
+++ b/onionshare/helpers.py
@@ -17,7 +17,7 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-import sys, os, inspect, hashlib, base64, platform, zipfile, tempfile, math, time
+import sys, os, inspect, hashlib, base64, platform, zipfile, tempfile, math, time, threading
 from random import SystemRandom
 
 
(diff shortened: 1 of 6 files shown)


micahflee:
(Sorry I'm so late.) Wow thanks for implementing this. Testing it now.

micahflee:
This doesn't quite implement what I have in mind for #177. Normally, onionshare closes the server as soon as the download completes. The `--stay-open` flag means don't do that, keep the server open for longer.

The feature in #177 is basically: close the server as soon as the download completes, or after X time, whichever comes first. So you onionshare a file with a time limit of 2 hours, and then send someone the link. If they don't click it within 2 hours, the server stops. If they do click it, the server closes as soon as the download finishes.

If you run onionshare with `--stay-open 2` using the code in this PR (to stay open for 2 hours), and the person on the other end immediately downloads the file, the server won't automatically close, and will in fact remain open for almost 2 hours more before it closes.

Would you be into changing it to work like that? So `--stay-open` would remain a bool and work how it used to, but adding a new flag `--shutdown-timeout` to allow for automatically shutting down the server after the timeout ends?

Also, I think the GUI can be improved. When I open the GUI the window it looks like this:

snapshot1

I think it should be something like:

```
☑ Stop sharing when download completes
☑ Stop sharing after [120] minutes
```

The checkbox would mean that `shutdown-timeout` is active, and the `[120]` could be a text field. If the server is currently running, that field be going down automatically, so you can see how long before the server stops.

choltz95:
These should be easy things to fix. I will make the changes in the next few days.

micahflee:
Thanks for implementing this @choltz95 but @mig5 has begun a new implementation that's farther along in #473, so I'm going to close this one.

PR #300, opened 2016-08-16
Add Docker support

pataquets:
Submitted Dockerfile builds an onionshare-cli image. If no argument is given at startup, help is displayed.

To test, start a Tor client. Here I'm using my own dockerized image:

```
docker run --rm -it --name tor -v /root/ pataquets/tor-client
```

I make the container's `/root/` dir to be a volume, since we'll need later from onionshare to find the control files. If you need to expose Tor's ports to the host, map them using ```-p 9051:9051``` option when launching (not needed for this test to work). Use as many ```-p``` as needed to expose more ports.

Now I run the onionshare container (here I'm using my public Docker Hub repo as an example). I mount previously started Tor client's root directory into the onionshare container (via `--volumes-from`). Everything after the image name (pataquets/onionshare) will be passed to the onionshare script. In the example below I just share the container's `/etc/hosts` file. To share files from the Docker host, you should mount them inside the container using Docker's `-v /host/dir/:/container/dir/` option.
Notice that I use `--net=container:tor` to use previously started tor container's networking stack, since I can't figure out how to specify a Tor server other than `localhost` from onionshare. 

```
docker run --rm -it --name onionshare --volumes-from tor --net=container:tor pataquets/onionshare --stay-open /etc/hosts
```

Here I'm using the `--rm` (remove container on finish) and interactive mode (`-it`) Docker launch flags. Stop containers by `CTRL+C`'ing them.


pataquets:
Bump.

pataquets:
@micahflee : Ping

pataquets:
Anything else needed for this to be reviewed/accepted? TIA.

georgeowell:
I'm assuming this would make it possible to run onionshare-cli on a headless server without installing x11 er al? If so, this would be a worthwhile addition.

pataquets:
Yes, that's the purpose of this PR. After merging, the next natural step would be to publish an 'official' image on Docker Hub via linking an Automated Build to this Github repo (see https://docs.docker.com/docker-hub/builds/).
Ping me if I can be of any help.

pataquets:
@georgeowell : Ping.

georgeowell:
Hehe there is no need to ping me. Let's just wait and see what Micah thinks.

pataquets:
Just making sure @micahflee is aware of recent discussion :)

pataquets:
@micahflee: any feedback?

b-meson:
Hi @pataquets thank you for your pull request. I am reviewing your PR and I think I am unsure about this comment here. 

`Notice that I use --net=container:tor to use previously started tor container's networking stack, since I can't figure out how to specify a Tor server other than localhost from onionshare`

The newer release of Onionshare changed how network Tor worked so I'm not sure this implementation is correct. Will take this PR for a spin against master.  More feedback TK

b-meson:
I ran this docker build and it hung on here for me ( I suspect it needs another Tor docker build but I think onionshare needs only _one_). It is unclear, see more below:

```
Jan 14 00:11:34.000 [notice] I learned some more directory information, but not enough to build a circuit: We need more microdescriptors: we have 0/6127, and can only build 0% of likely paths. (We have 0% of guards bw, 0% of midpoint bw, and 0% of exit bw = 0% of path bw.)
Jan 14 00:11:37.000 [notice] Bootstrapped 50%: Loading relay descriptors
Jan 14 00:11:51.000 [notice] Bootstrapped 57%: Loading relay descriptors
Jan 14 00:11:54.000 [notice] Bootstrapped 65%: Loading relay descriptors
Jan 14 00:12:03.000 [notice] Bootstrapped 70%: Loading relay descriptors
Jan 14 00:12:09.000 [notice] Bootstrapped 78%: Loading relay descriptors
Jan 14 00:12:17.000 [notice] Bootstrapped 80%: Connecting to the Tor network
Jan 14 00:12:18.000 [notice] Bootstrapped 90%: Establishing a Tor circuit
Jan 14 00:12:19.000 [notice] Tor has successfully opened a circuit. Looks like client functionality is working.
Jan 14 00:12:19.000 [notice] Bootstrapped 100%: Done
```

I think it could use a rebase on top of the latest onionshare and testing against that. Right now it just connects to Tor on docker, but I don't think this is a full correct implementation. Happy to talk further or help in re-design @pataquets !

micahflee:
Hey @pataquets sorry I never got around to reviewing this. I'm into supporting docker, but like @b-meson said, would you be able to remake this using the latest master (because a lot has changed), and also change it so that it only uses one Docker container that contains both OnionShare and bundled Tor?

pataquets:
Hi guys.
@b-meson : I'll try to figure out the new connection method. It will make things more easy. Are these the up-to-date docs? https://github.com/micahflee/onionshare/wiki/Connecting-to-Tor
@micahflee : It could be done. But, in general, packaging more than one service in the same container is an antipattern. The Docker way is to orchestrate two containers, one for the OnionShare service and another for the Tor service and link them. Docker provides lots of mechanisms to interconnect containers. The easiest way (and most convenient also for local development and testing) is to use Docker Compose and create a manifest setting everything up. I can provide an appropriate manifest if I find my way around linking OnionShare to Tor.

As of now, I'm on building the new version and figuring how to connect to a remote Tor server. Stay tuned :)

pataquets:
So far, the image builds the latest code (just checked that usage info displays). The AB on Docker Hub [triggered](https://hub.docker.com/r/pataquets/onionshare/builds/), and you'll be able to test if it works with the OP instructions.
Now, I'll try to figure out how to connect OnionShare to an external Tor, ideally to a TCP reachable Tor server via some cmdline switch. Not sure when I'll be able to take a stab at it, though.

b-meson:
@micahflee since I am not a contributer to OnionShare I can't assign myself this issue for review on GH but consider it being QAed by me. Thank you @pataquets

b-meson:
I get the following stacktrace error in OS X 
```
zerohistory:~ bmeson$ docker run --rm -it --name onionshare --volumes-from tor --net=container:tor pataquets/onionshare --stay-open /etc/hosts
  File "./dev_scripts/onionshare", line 28, in <module>
    onionshare.main()
  File "/usr/src/app/onionshare/__init__.py", line 87, in main
    onion.connect(settings=False, config=config)
  File "/usr/src/app/onionshare/onion.py", line 216, in connect
    self.tor_proc = subprocess.Popen([self.tor_path, '-f', self.tor_torrc], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
  File "/usr/local/lib/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/usr/local/lib/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/usr/bin/tor': '/usr/bin/tor'
```

pataquets:
@micahflee : Found a slot to try to make it work, but failed.

I've got the same as @b-meson here. Even provided a ```--config``` switch to a JSON containing:
```
{
  "connection_type": "control_port",
  "control_port_address": "127.0.0.1",
  'socks_address': '127.0.0.1',
  "auth_type": "no_auth",
  'auth_password': '',
  'close_after_first_download': True,
  'systray_notifications': False,
}
```
But OnionShare keeps complaining about not existing ```/usr/bin/tor```.
Also, there is no docs that I could found which deal with connecting OnionShare CLI to a remote, TCP reachable Tor service, so I figured out that the JSON is the way to go, but not sure.
I'm stuck now. Please, advise. TIA.

mig5:
Yeah the paths to the Tor binary are hardcoded per operating system, see [get_tor_paths()](https://github.com/micahflee/onionshare/blob/master/onionshare/common.py#L91) - may take some refactoring to change this, or, despite the anti-pattern, maybe it would be simpler to let the OnionShare container indeed contain its own 'bundled' Tor to mirror non-container installations more closely.

pataquets:
As far as I can understand, if I configure a remote Tor instance, there's no point in trying to start a local daemon and complaining about a not found binary might even be considered a bug in such case. The fact that it's hardcoded is not a problem here.
Also, starting a local daemon to end up using a remote Tor sounds to me as wasting resources.
But perhaps I'm making some incorrect assumption here on the remote Tor usage here, since I'm not 100% sure on how to make it work, anyway. I need help here to fully understand how it should be used.

Making more than one process run inside the same container has several drawbacks and makes the container behave different. Also, it adds unnecessary complexity to build recipe.
A lot of Docker users would end not using such an image, especially for production use.

mig5:
No you’re right, I somehow missed that you were trying to connect via control port, which indeed should work without needing the Tor binary.

I’ll do some debugging and see what the issue might be.

mig5:
So this is working for me, using your pataquets/tor-client container as per the original comment in the PR

```
docker run --rm -it --name onionshare --volumes-from tor --net=container:tor onionshare --config ./docker/onionshare.json  /etc/hosts
Onionshare 1.2 | https://onionshare.org/
Configuring onion service on port 17638.
Settings saved to ./docker/onionshare.json
Preparing files to share.
 * Running on http://127.0.0.1:17638/ (Press CTRL+C to quit)
Give this URL to the person you're sending the file to:
http://6cv6hs7cmsul5bp5.onion/shallow-tyke

Press Ctrl-C to stop server
```

My Dockerfile:
```
FROM python:3

ADD ./install/requirements.txt /usr/src/app/install/requirements.txt
WORKDIR /usr/src/app/
RUN \
  pip3 install -r ./install/requirements.txt

ADD . /usr/src/app/

ENTRYPOINT [ "./dev_scripts/onionshare" ]
```

I added /docker/onionshare.json to the repo that looks like this:

```
{"version": "1.2", "connection_type": "control_port", "control_port_address": "127.0.0.1", "control_port_port": 9051, "socks_address": "127.0.0.1", "socks_port": 9050, "socket_file_path": "/var/run/tor/control", "auth_type": "no_auth", "auth_password": "", "close_after_first_download": true, "systray_notifications": true, "use_stealth": false, "use_autoupdate": true, "autoupdate_timestamp": null, "no_bridges": true, "tor_bridges_use_obfs4": false, "tor_bridges_use_custom_bridges": "", "save_private_key": false, "private_key": "", "slug": "", "hidservauth_string": ""}
```

Maybe the issue is because your branch is not using latest master (the ```--config``` parameter was functionality added long after your PR was created).

Do you want to try against latest master just to see if it works for you, then update your branch to add the docker-specific onionshare.json? (I don't know if it's 'good practice' or not to have the config in a 'docker' subfolder of the app, but it has to go somewhere.. 

So this works but it depends on using your tor-client Docker container. I can imagine many users might prefer to use an 'official' Tor Docker image, do you know if such a thing exists? I don't know if we want Onionshare to depend on a 'contrib' Tor container: if something happens to your personal repo, then the Onionshare container won't be usable.

If there's no way around using your tor-client container, at least for now, then we could call this complete but we'd also need documentation so users know they need to run pataquets/tor-client and example commands. Maybe can add a new section to the BUILD.md.

mig5:
Thinking about it some more, I'm not sure we'd want the burden of maintaining official tor/onionshare containers given the existing workload of cross-platform compat.. anyone wanting to use Docker with onionshare could just ```docker build``` from the Dockerfile and use a paired tor-client container if they wish.

I still, however, think it's worthwhile installing the Tor binary in the container. I know it's an anti-pattern but:

a) You don't need to worry about supervisord etc as it doesn't need to be a running service - Onionshare will invoke the execution of the binary once it compiles the torrc per share, as a subprocess of the main onionshare command

b) it allows the user more flexibility e.g they can supply their own onionshare.json config that maybe has bridge settings added. Rather than have to tweak your tor/tor-client containers to support said bridges.

Yes it's an anti-pattern but it is a simpler way that offers the user more flexibility and it doesn't complicate the container design at all, it's just one more package included alongside all the other python dependencies.

mig5:
Here's my Dockerfile now

```
FROM python:3

RUN \
  DEBIAN_FRONTEND=noninteractive \
    apt-get update && apt-get -y install lsb-release

RUN \
  apt-key adv --keyserver hkp://hkps.pool.sks-keyservers.net --recv-keys 886DDD89 && \
  echo "deb http://deb.torproject.org/torproject.org $(lsb_release -cs) main" >> /etc/apt/sources.list && \
  apt-get update && \
  DEBIAN_FRONTEND=noninteractive \
    apt-get -y install \
      deb.torproject.org-keyring \
      obfs4proxy \
      tor \
  && \
  apt-get clean && \
rm -rf /var/lib/apt/lists/*

ADD ./install/requirements.txt /usr/src/app/install/requirements.txt
WORKDIR /usr/src/app/
RUN \
  pip3 install -r ./install/requirements.txt

ADD . /usr/src/app/

ENTRYPOINT [ "./dev_scripts/onionshare" ]
```

Then I can run it with

```
docker run --rm -it --name onionshare onionshare  /path/to/file/to/share
```

or, if I want to add a custom json file 'docker.json' in the same dir as the app (perhaps one that uses obfs4 bridges or something, due to censorship):

```
docker run --rm -it --name onionshare onionshare  --config ./docker.json /etc/hosts
```

If you would be willing to make it like this, as well as an update to BUILD.md to cover:

1) Basic instructions on how to build and run the container
2) Instructions on how to add a custom .json settings file (per my above example, by placing it in the app codebase somewhere and adding the --config flag to the `docker run` command). Because tor would be bundled, we wouldn't *have* to add a custom json file by default which is nice.

.. then I think this could be merged. Maybe after that we could link the repo to DockerHub to build or at least tag a release when we release new OnionShare versions.

I feel like this is a decent middle ground - 'bundled' tor isn't relying on a separate service, it's just invoking a command via Python's subprocess module, so it doesn't really break Docker design (IMO) (no more than your own tor container's tor binary already has to subprocess out to obfsproxy when using obfs, for example) and this way we avoid depending on third party containers.

Issue #306, opened 2016-09-08
0.9.1 requires 10.11.x


BenjaminHCCarr:
It is very odd in standard FOSS numbering to abandon an OS version in a point release, particularly a minor point release. 0.9 runs fine but 0.9.1 requires 10.11. Not everyone who uses this software will have updated, or has reason or concern for upgrading.
![screen shot 2016-09-07 at 11 45 05 pm](https://cloud.githubusercontent.com/assets/8225968/18336472/80c6c89c-7555-11e6-9d67-258671eb713c.png)

micahflee:
Oops, I had no idea that 0.9.1 no longer works in OS X 10.10. This must have been a byproduct of switching from PyInstaller to cx_Freeze.

micahflee:
Did you get that popup while running the `OnionShare.pkg` installer or while running `/Applications/OnionShare.app`?

BenjaminHCCarr:
Sadly it's on first run not on install. Did you move from python 2.7? I know a lot of stuff is breaking as they migrate. There is zero warning in the install, but on first launch you get that error. So people are using ~20+ MB that may be expensive and installing over 0.9 only to find it doesn't run. I tried toggling back and forth. Attaching a slew of logs for you, I don't have time to look at them, but I love the app.

[onionshare_2016-08-25-091422_tyr.crash.txt](https://github.com/micahflee/onionshare/files/462758/onionshare_2016-08-25-091422_tyr.crash.txt)
[onionshare_2016-08-25-091422-1_tyr.crash.txt](https://github.com/micahflee/onionshare/files/462759/onionshare_2016-08-25-091422-1_tyr.crash.txt)
[onionshare_2016-09-07-154226_tyr.spin.txt](https://github.com/micahflee/onionshare/files/462755/onionshare_2016-09-07-154226_tyr.spin.txt)
[onionshare_2016-09-07-154811_tyr.spin.txt](https://github.com/micahflee/onionshare/files/462757/onionshare_2016-09-07-154811_tyr.spin.txt)
[onionshare_2016-09-07-232831_tyr.spin.txt](https://github.com/micahflee/onionshare/files/462754/onionshare_2016-09-07-232831_tyr.spin.txt)
[onionshare.log.txt](https://github.com/micahflee/onionshare/files/462756/onionshare.log.txt)

micahflee:
We switched from python2 to python3 a few releases ago. The major change that I think affected this is switching to cx_Freeze #303.

[...]

PR #307, opened 2016-09-08
Update Readme to show OS Requirements

BenjaminHCCarr:
Indicate that 0.9 works on 10.10.x, but 0.9.1 requires 10.11.x or higher, as I imagine future version will.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ You can download OnionShare to install on your computer from <https://onionshare
 
 You can set up your development environment to build OnionShare yourself by following [these instructions](/BUILD.md).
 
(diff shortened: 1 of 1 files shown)


micahflee:
Since #306 is just a bug and I'm hoping 0.9.2 will work again in OS X 10.10.x, I'm just going to close this. If you'd like to help figure out how to fix #306 though I'll merge that PR :).

BenjaminHCCarr:
Yeah, I update #306 with a ton of logs, I am guessing python related, but I haven't had time to look. If we can fix it in 0.9.2 and call it a regression, we're all good.

PR #309, opened 2016-09-18
Import adjustment

tpltnt:
Hi micahflee,

executing the scripts fails if you follow the BUILD.md instructions due to missing modules.

``` sh
$ ./install/scripts/onionshare
  File "./install/scripts/onionshare", line 21, in <module>
    import onionshare
ImportError: No module named 'onionshare'
```

My changes adjust the import path to enable the user to run both the CLI and the GUI version of OnionShare as described.
Cheers,
tpltnt

### DIFF
--- a/install/scripts/onionshare
+++ b/install/scripts/onionshare
@@ -18,5 +18,24 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-import onionshare
+try:
+    # try to use module installed on system first
+    import onionshare
+except ImportError:
+    # use module provided by source repository
+    import os, sys, inspect
+    # path used to call the executable
+    relative_exec_path = inspect.getfile(inspect.currentframe())
+    # split off path to executable
+    path_to_executable = os.path.split(relative_exec_path)[0]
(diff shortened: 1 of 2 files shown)


micahflee:
I'm getting an error:

```
user@dev:~/code/onionshare$ install/scripts/onionshare
  File "install/scripts/onionshare", line 23, in <module>
    import onionshare
ImportError: No module named 'onionshare'

During handling of the above exception, another exception occurred:

  File "install/scripts/onionshare", line 39, in <module>
    import onionshare
ImportError: No module named 'onionshare'
```

If I add `print(sys.path)` right before `import onionshare`, it looks like this: `['install/scripts', '/home/user/code/onionshare/install/scripts', '/usr/lib/python35.zip', '/usr/lib/python3.5', '/usr/lib/python3.5/plat-x86_64-linux-gnu', '/usr/lib/python3.5/lib-dynload', '/usr/local/lib/python3.5/dist-packages', '/usr/lib/python3/dist-packages']`

However there's another issue here too. The location isn't the only path that onionshare needs to know about to run in a not-installed situation. It also needs access to everything in the `resources` directory, via the `helpers.get_resource_path()` function: https://github.com/micahflee/onionshare/blob/master/onionshare/helpers.py#L31

Issue #171, opened 2014-12-10
Write AppArmor profile for OnionShare


micahflee:

u451f:
At Tails we've started to write some profiles. I'm gonna review them and will propose them upstream when I'm done.

See profiles for onionshare{,-gui} and an abstraction they can share:
* https://git-tails.immerda.ch/tails/tree/config/chroot_local-includes/etc/apparmor.d/usr.bin.onionshare?h=feature/7870-include_onionshare
* https://git-tails.immerda.ch/tails/tree/config/chroot_local-includes/etc/apparmor.d/usr.bin.onionshare-gui?h=feature/7870-include_onionshare
* https://git-tails.immerda.ch/tails/tree/config/chroot_local-includes/etc/apparmor.d/abstractions/onionshare?h=feature/7870-include_onionshare

u451f:
If you want to, you can assign this issue to me.

PR #320, opened 2016-11-19
Apparmor profiles

u451f:
Hi Micah,

as you know, we've started to work on including OnionShare in Tails and we'd like to add some AppArmor profiles to it. Here they are. This would fix https://github.com/micahflee/onionshare/issues/171.

I'll take care of integrating them into the official Debian package.

What do you think?

Cheers!
u.

(Tails ticket: https://labs.riseup.net/code/issues/7870 )


micahflee:
This looks great, thank you!

Issue #321, opened 2016-12-18
Current wordlist uses unusual words, proper nouns, and literal numbers; could benefit from more dictionary words


commandnotfound:
Whenever `onionshare` generates a slug to use in the URL, it makes use of words under `resources/wordlist.txt`, likely based of one of the [original `diceware` wordlists](http://world.std.com/~reinhold/diceware.html).

The current wordlist makes use of "words" that include `a`, `92nd`, and 'moran`. While it appears the wordlist avoids using vulgar words, it could still benefit from improvement.

The Electronic Frontier Foundation [recently generated their own diceware wordlist](http://world.std.com/~reinhold/diceware.html) that provides a reasonable list. While these words are, on average, longer, it would be provide more usability. One might communicate an `onionshare` URL over an audio call, perhaps over Signal. Someone else might access an `onionshare` URL on their mobile device, where autocorrect is prevalent and prefers dictionary words.

Thus, a move to take the words from the EFF diceware wordlist and move them into a format appropriate for `onionshare` is called for.

micahflee:
Merged! Thanks this is great.

PR #322, opened 2016-12-18
Replaced wordlist with well researched and more usable wordlist from EFF

commandnotfound:
A modification was made to `resources/wordlist.txt` such that it makes use of [the EFF recommended diceware wordlist](https://www.eff.org/files/2016/07/18/eff_large_wordlist.txt).

Related to #321.


micahflee:
This is awesome, thank you.

Issue #332, opened 2016-12-26
Exception when start sharing with newest version and distro-standard stem


b3sigma:
The exception is as follows:

```
Onionshare 0.9.2 | https://onionshare.org/
 * Running on http://127.0.0.1:17600/ (Press CTRL+C to quit)
Configuring onion service on port 17600.
Staring ephemeral Tor onion service and awaiting publication
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.5/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "/home/fuck/Documents/create/onionshare/onionshare_gui/onionshare_gui.py", line 172, in start_onion_service
    self.app.start_onion_service()
  File "/home/fuck/Documents/create/onionshare/onionshare/onionshare.py", line 89, in start_onion_service
    self.onion_host = self.onion.start(self.port)
  File "/home/fuck/Documents/create/onionshare/onionshare/onion.py", line 133, in start
    res = self.c.create_ephemeral_hidden_service({ 80: port }, await_publication=True, basic_auth=basic_auth)
TypeError: create_ephemeral_hidden_service() got an unexpected keyword argument 'basic_auth'
```
Looks like basic_auth is supported in stem 1.5.0, but I had by default:
$pip3 show stem
Name: stem
Version: 1.4.1

PR #333, opened 2016-12-26
Support stem versions older than 1.5.0 again. Fix for issue #332

b3sigma:
Support stem versions older than 1.5.0 again by not including named keyword argument "basic_auth" unless the previously detected interface succeeded without an exception. Fix for issue #332


micahflee:
Oops, thank you!

PR #337, opened 2016-12-30
Alternative fix for template content injection vulnerability

### LINKED ISSUE #319: Filenames aren't escaped

garrettr:
As I mentioned in my [comment](https://github.com/micahflee/onionshare/issues/319#issuecomment-269794722), I think a better fix for #319 is to fix the insecure default for autoescaping template variables in Flask when rendering templates using the `render_template_string` method. The insecure default was fixed in Flask starting with Flask 0.11, so we only need to fix the behavior in versions of Flask earlier than 0.11.

This PR detects if OnionShare is using an insecure version of Flask, and if so, uses monkeypatching to apply the upstream fix included in versions of Flask >= 0.11. It also reverts the previous attempts to fix this issue.

Unfortunately, it is not easy for me to test this because I do not have ready access to a vulnerable environment (such as Debian Jessie). @micahflee can you test this? If it is too much trouble, let me know, and I will try to find some more of my own free time to test it.

Fixes #319. In addition, this solution prevents **all** content injection in templates rendered using the `render_template_string` method. Since all of OnionShare's templates are HTML templates rendered using `render_template_string`, this is a useful fix that prevents this entire class of content injection vector and is therefore more future-proof than https://github.com/micahflee/onionshare/commit/cff11cd7e48fe4e94ef484ba1f1095362330452d. It also prevents the potential issues with performing HTML auto-escaping in contexts unrelated to rendering HTML, as I discussed in https://github.com/micahflee/onionshare/issues/319#issuecomment-269794722.

Finally, this fix is only necessary if OnionShare can be used with versions of Flask that have insecure defaults for `render_template_string`. In the future, you may require OnionShare use any version of Flask >= 0.11, in which case this code could safely be removed.

### DIFF
--- a/onionshare/web.py
+++ b/onionshare/web.py
@@ -17,7 +17,7 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-import queue, mimetypes, platform, os, sys, socket, logging, html
+import queue, mimetypes, platform, os, sys, socket, logging, re
 from urllib.request import urlopen
 from flask import Flask, Response, request, render_template_string, abort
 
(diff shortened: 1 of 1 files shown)


garrettr:
Sorry about accidentally creating this issue with the first line of the description instead of the title! Was using `hub pull-request` and forgot that the first line is used as the title. I've fixed this PR's title and description!

micahflee:
Thanks @garrettr! I tested on Debian Jessie and this fix does indeed properly escape the filename `` in HTML.

PR #348, opened 2017-01-18
AppArmor profile improvements

u451f:
Hi Micah,

anonym from Tails has updated the profiles and here are the corresponding changes. Could you include them?


micahflee:
I'd be happy to

u451f:
thanks!

PR #388, opened 2017-05-18
#386 - put the connection radio buttons into their own group, which b…

### LINKED ISSUE #386: In Settings, clicking on selected radio buttons unselects them
For example, if "Use Tor that is bundled with OnionShare" is selected, and you click it again, it unselects it so that nothing is selected. It shouldn't unselect the option, it should keep it selected.

mig5:
By putting the radio buttons into their own group widget, autoExclusive is set by default which prevents them from being all untoggled

Fixes #386


micahflee:
Thank you, this is awesome! I'm reviewing it now

PR #389, opened 2017-05-18
#345 - validate the readability of files before adding them

### LINKED ISSUE #345: Trying to share unreadable files results in silently ignoring them
As a user, I've experienced a UX problem, that's especially important when onionshare is run under some sandbox or confinement: if I add, in the GUI, files that onionshare is not allowed to read, then onionshare will simply share an empty archive. Arguably there's already indirect feedback about it to the user: the total size of the to-be-shared files is displayed. Still, it's suboptimal: first, it's _indirect_ feedback (one has to look for it and reason about it to notice there's a problem); second, when sharing a folder that contains both readable and unreadable files, it's hard to notice that the total size is not what it should be.

I'm not a UX expert, so I won't bother suggesting solutions, and will stick to my user hat for today :)

mig5:
Should fix #345 in my tests

Also fixes a perhaps previously undiscovered bug in that the onionshare_gui component contained a call to Alert in its validation, however Alert was not defined (imported)

### DIFF
--- a/onionshare/__init__.py
+++ b/onionshare/__init__.py
@@ -67,6 +67,9 @@ def main(cwd=None):
         if not os.path.exists(filename):
             print(strings._("not_a_file").format(filename))
             valid = False
+        if not os.access(filename, os.R_OK):
+            print(strings._("not_a_readable_file").format(filename))
+            valid = False
     if not valid:
         sys.exit()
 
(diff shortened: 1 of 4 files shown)


micahflee:
Thank you, this is awesome. I just tested it, and it completely fixes the issue in the CLI version, but in the GUI version there are a few places where it was incomplete. When you use the "Add Folder" button and select an unreadable folder it let it through, and also when you drag-and-drop unreadable files and folders it let them through. But I made a simple commit that fixes those issues too.

Issue #385, opened 2017-05-17
Bundled Tor sometimes crashes with "dyld: Library not loaded: @executable_path/libevent-2.0.5.dylib"


micahflee:
I'd like OnionShare to support macOS 10.9 and newer. However, when I try running OnionShare on a 10.9 computer using the bundle Tor setting, it fails with the error:

> [Errno 2] No such file or directory

When I manually run the `tor` binary that OnionShare ships with, it immediately crashes:

```sh
$ /Applications/OnionShare.app/Contents/MacOS/tor 
dyld: Library not loaded: @executable_path/libevent-2.0.5.dylib
  Referenced from: /Applications/OnionShare.app/Contents/MacOS/tor
  Reason: image not found
Trace/BPT trap: 5
```
However, when I run the same binary on a Mac with a newer OS, it doesn't crash (the error messages are expected since I'm not providing a special `torrc` here, and I'm running a tor service in the background):

```sh
$ /Applications/OnionShare.app/Contents/MacOS/tor 
May 17 09:30:51.831 [notice] Tor 0.3.0.6 (git-47d2e4f06ec26a79) running on Darwin with Libevent 2.0.22-stable, OpenSSL 1.0.2k and Zlib 1.2.8.
May 17 09:30:51.832 [notice] Tor can't help you if you use it wrong! Learn how to be safe at https://www.torproject.org/download/download#warning
May 17 09:30:51.893 [notice] Configuration file "/home/debian/install/etc/tor/torrc" not present, using reasonable defaults.
May 17 09:30:51.897 [notice] Opening Socks listener on 127.0.0.1:9050
May 17 09:30:51.897 [warn] Could not bind to 127.0.0.1:9050: Address already in use. Is Tor already running?
May 17 09:30:51.897 [warn] Failed to parse/validate config: Failed to bind one of the listener ports.
May 17 09:30:51.897 [err] Reading config failed--see warnings above.
```

This is what the command line version in macOS 10.9 looks like:

```sh
$ /Applications/OnionShare.app/Contents/MacOS/onionshare --debug code/onionshare/README.md 
Onionshare 1.1.dev1 | https://onionshare.org/
[May 17 2017 09:39:27] Onion.__init__
[May 17 2017 09:39:27] Onion.connect
[May 17 2017 09:39:27] Settings.__init__
[May 17 2017 09:39:27] Settings.load
Traceback (most recent call last):
  File "site-packages/stem/socket.py", line 442, in _make_socket
[...]

PR #390, opened 2017-05-18
Wait a bit longer before attempting to connect to the Tor controller

mig5:
Seems to fix remaining issue in #385 but will it cover all cases of 'slow computer' (whilst you don't want it to sleep for so long as to be annoying on fast computers)?


micahflee:
I just tested this and it totally works. And in fact, even if we sleep 2 seconds, it doesn't really affect much in the UI. It always takes longer than that to connect to Tor anyway, and sleeping 2 seconds looks fine. Hopefully this will be long enough to cover all slower computers (considering on my laptop 0.2 was long enough).

Issue #392, opened 2017-05-19
onionshare-gui symlink pointing to itself in onionshare/ dir


mig5:
Just curious - there's an onionshare-gui symlink in the onionshare/ folder which points to itself.

Introduced in https://github.com/micahflee/onionshare/commit/6d58d5dcfff341a403c796bb1f44abf9e39cfd6f

Looks like a mistake? (I noticed from getting 'too many levels of symbolic links' when trying to run the install/build_deb.sh script)

PR #393, opened 2017-05-19
Remove onionshare-gui symlink pointing to itself

mig5:
Since this breaks execution of the install/build_deb.sh as per #392


micahflee:
Yeah, this must have been a mistake. Good catch.

Issue #176, opened 2015-04-17
Build fails on Ubuntu vivid


pde:
```
./build_deb.sh 
rm: cannot remove ‘deb_dist’: No such file or directory
/usr/lib/python2.7/dist-packages/setuptools/dist.py:282: UserWarning: Normalizing '0.7dev' to '0.7.dev0'
  normalized_version,
running bdist_deb
running sdist_dsc
running egg_info
writing onionshare.egg-info/PKG-INFO
writing top-level names to onionshare.egg-info/top_level.txt
writing dependency_links to onionshare.egg-info/dependency_links.txt
reading manifest file 'onionshare.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching 'onionshare/strings.json'
writing manifest file 'onionshare.egg-info/SOURCES.txt'
running sdist
warning: sdist: standard file not found: should have one of README, README.rst, README.txt

running check
creating onionshare-0.7.dev0
creating onionshare-0.7.dev0/bin
creating onionshare-0.7.dev0/images
creating onionshare-0.7.dev0/locale
creating onionshare-0.7.dev0/onionshare
creating onionshare-0.7.dev0/onionshare.egg-info
creating onionshare-0.7.dev0/onionshare_gui
creating onionshare-0.7.dev0/setup
creating onionshare-0.7.dev0/test
making hard links in onionshare-0.7.dev0...
hard linking BUILD.md -> onionshare-0.7.dev0
hard linking LICENSE -> onionshare-0.7.dev0
hard linking MANIFEST.in -> onionshare-0.7.dev0
hard linking README.md -> onionshare-0.7.dev0
hard linking setup.cfg -> onionshare-0.7.dev0
hard linking setup.py -> onionshare-0.7.dev0
hard linking version -> onionshare-0.7.dev0
hard linking bin/onionshare -> onionshare-0.7.dev0/bin
hard linking bin/onionshare-gui -> onionshare-0.7.dev0/bin
hard linking images/drop_files.png -> onionshare-0.7.dev0/images
hard linking images/logo.png -> onionshare-0.7.dev0/images
hard linking images/server_started.png -> onionshare-0.7.dev0/images
hard linking images/server_stopped.png -> onionshare-0.7.dev0/images
hard linking images/server_working.png -> onionshare-0.7.dev0/images
hard linking locale/de.json -> onionshare-0.7.dev0/locale
[...]

PR #394, opened 2017-05-19
387 build script logic fix

### LINKED ISSUE #387: Dev version: Script ./install/build_deb.sh gies the frong filename at the end
it's
sudo dpkg -i deb_dist/onionshare_1.1~dev1-1_all.deb
not
sudo dpkg -i deb_dist/onionshare_1.1.dev1-1_all.deb

Ubuntu 17.04

mig5:
stdeb renames .dev to ~dev for compatibility with Debian's strict rules.

This fixes #387 and (in case you like a mystery case closed) I dare say it was probably a similar reason for #176 :)

The alternative would be to ensure you always remember the ~ in share/version.txt but since it's Debian specific (and we're human), and that stdeb already fixes it for you, may as well force-fix it when printing the final filename.

This branch also fixes a separate tiny logic bug - you were checking $? but *after* a whitespace echo, meaning that the script technically never ran your else (exit 1) conditional.

### DIFF
--- a/install/build_deb.sh
+++ b/install/build_deb.sh
@@ -12,8 +12,8 @@ rm -r deb_dist >/dev/null 2>&1
 python3 setup.py --command-packages=stdeb.command bdist_deb
 
 # return install instructions if onionshare builds properly
-echo ""
 if [[ $? -eq 0 ]]; then
+	echo ""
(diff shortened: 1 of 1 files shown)


micahflee:
Nice.

Issue #228, opened 2015-11-21
Localization is applied to the download page, which is disclosing the locale of the sender


alexlehm:
When running OnionShare on a localized OS (e.g. Windows with German, French etc localization), the recipient gets information about locale of sender, giving some indication where sender is (even though only a broad language region).

I would prefer it if the download page is not localized or it is configurable if the page is localized or not.

Some of the language files do not contain the localization keys for the words, so the issue does not apply to all languages, it is currently in de, es, fi, fr and nl

micahflee:
This is an excellent point that I never considered. I think the easiest solution is to not localize the download page at all, and just leave it in English. The second easiest (and maybe better) solution is to remove all language from the download page and replace it with icons.

burdges:
Yeah, this could be problematic in Europe where many people work in countries different from their native one.

Just fyi, it's possible to use the recipient's locale in straight JavaScript.  https://taler.net/ does this.  It doesn't really protect the recipient from leaking their locale to the sender, but kinda.  It'd leak the sender's onionshare version by leaking the locales supported by the sender.

PR #395, opened 2017-05-19
More descriptive HTML

mig5:
Be explicit that the the first filesize is that of the compressed zip file and that it contains the files beneath. Use the headers in the table that we've styled for

screen shot 2017-05-19 at 3 43 29 pm

### DIFF
--- a/share/html/index.html
+++ b/share/html/index.html
@@ -57,7 +57,9 @@
       .download-size {
         color: #999999;
       }
-
+      .download-description {
+        padding: 10px;
+      }
       .file-list {
         margin: 50px auto 0 auto;
         padding: 10px;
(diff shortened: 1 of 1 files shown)


micahflee:
Interestingly, the download page used to have text on it that was localized into several different languages, but it turns out this was a privacy leak because it let the recipient know the language on the sender's computer (see #228). I solved it by removing all language.

However, I've been working with a UX designer who made some really good looking download page changes, and they include text. So I think I'll keep the text on the download page, and just make sure it's always in English. So even if, for example, a German-speaking OnionShare user shares something to another German speaker, the download page will be in English. This seems like a reasonable trade-off.

Issue #253, opened 2016-01-02
Add some security headers


blobgo:
It would be nice if the server could automatically send some security headers (like CSP and so on).
Here is a nice overview: https://securityheaders.io

PR #397, opened 2017-05-20
Harden some response headers

mig5:
A tentative improvement for #253. Doesn't seem to cause any problems with browsing the slug/downloading, in my tests.

I also noticed the Server signature header was set to a Werkzeug Server / python string. I thought it could be perhaps set to the OnionShare version string, although perhaps it would be better security to not show the version at all.

Thoughts welcome

### DIFF
--- a/onionshare/web.py
+++ b/onionshare/web.py
@@ -21,7 +21,7 @@ from distutils.version import StrictVersion as Version
 import queue, mimetypes, platform, os, sys, socket, logging
 from urllib.request import urlopen
 
-from flask import Flask, Response, request, render_template_string, abort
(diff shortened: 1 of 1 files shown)


mig5:
tests fail since I added the Onionshare version string as the Server signature. Seems to be a problem with the test itself as it worked fine in my manual tests

FileNotFoundError: [Errno 2] No such file or directory:'/usr/share/onionshare/locale'

micahflee:
I think you're right about the travis tests needing `onionshare_dev_mode` to be set to pass. But in any case, I'm slightly tweaking this to remove the version string from the http response, and remove the need for strings in the tests anyway. No need to disclose any more information about the OnionShare user's software than necessary.

PR #398, opened 2017-05-20
Add help button to settings dialog, which takes user to the wiki

mig5:
There's a slight bug in Python on OS X 10.12.5 that breaks the native webbrowser.open() function. See http://bugs.python.org/issue30392 & https://github.com/jupyter/notebook/issues/2438 for more context.

Using os.system('open ...') works around it - it even opens a new tab in an already-running browser, so it's not bad.

After testing both styles, I thought the Help button looked better for UX on the left, isolated from Save/Cancel... opinions welcome
screen shot 2017-05-20 at 2 03 27 pm

### DIFF
--- a/onionshare_gui/settings_dialog.py
+++ b/onionshare_gui/settings_dialog.py
@@ -18,7 +18,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from PyQt5 import QtCore, QtWidgets, QtGui
-import sys, platform, datetime
+import os, sys, platform, datetime, webbrowser
 
 from onionshare import strings, common
 from onionshare.settings import Settings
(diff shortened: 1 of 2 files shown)


micahflee:
Adding a help button is a great idea. You know how in in the Settings dialog there's "Stealth (advanced)" there's a "More information" link? I didn't write code that opens a link in the default browser for that, I just used a `<a href="">` inside of a `QLabel`, and Qt5 took care of the rest.

So I decided to see how Qt5 opens links in a cross-platform manner, and it looks like it has its own function: http://doc.qt.io/qt-5/qdesktopservices.html#openUrl

Also, `QDesktopServices` is it `QtGui`: http://pyqt.sourceforge.net/Docs/PyQt5/api/qdesktopservices.html

Want to refactor this to use `QDesktopServices.openUrl` instead?

mig5:
Oh, that's so much better, and Just Works (I have only tested on Mac - appreciate some extra OS testing just in case). Refactored.

PR #399, opened 2017-05-20
Remove unneeded modules/unreferenced variables as discovered by pyflakes

mig5:
I ran pyflakes (v1.5.0, as installed with 'pip3 install pyflakes') over the codebase and found a few cases where modules are imported but never used (as well as one or two variables assigned but never used).

Given this cleanup covers a wide number of files, it's probably worth a bit of careful regression testing. Basic test of the app after building on OS X seemed fine for me.

### DIFF
--- a/onionshare/onion.py
+++ b/onionshare/onion.py
@@ -21,9 +21,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 from stem.control import Controller
 from stem import ProtocolError, SocketClosed
 from stem.connection import MissingPassword, UnreadableCookieFile, AuthenticationFailure
-import os, sys, tempfile, shutil, urllib, platform, subprocess, time, shlex
+import os, sys, tempfile, platform, subprocess, time, shlex
 
-from . import socks
 from . import common, strings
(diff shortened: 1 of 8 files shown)


micahflee:
Speaking of needing careful regression testing, I've done a completely abysmal job of writing tests for OnionShare. I just opened a new issue to write more tests: #404. If you have free time, I'd be more than happy to merge PRs full of tests :).

mig5:
Yeah, I'm trying to read up on how to do tests with Travis. It would also be really good for catching things like #401 if there's some way to simulate clicking on the UI.

micahflee:
Yeah, exactly. Right now none of the GUI is tested. (Honestly barely anything is tested.) I think I'm gonna wait until #404 is finished and there's much more test coverage than there currently is before merging this one, just so we don't accidentally introduce more regressions like #401.

Wanna move the discussion about which test frameworks to use and how to simulate clicking buttons to the #404 issue?

micahflee:
Since this is from nearly a year ago, I'm going to close this PR, but we should open another brand new pyflakes clean up PR instead.

PR #400, opened 2017-05-20
Send stay_open variable in the right place, so that it really does stay open

mig5:
'stay_open' was always evaluating to False, even if we had set it to true in settings. So even if you wanted to stay open, it would still close after the first download.

This turned out to be because it was being set in the wrong place - by the time you start the onionshare http thread, you are sending self.app.stay_open which is still False. 

I verified with some common.log action that even if it was being set to True earlier, it would still be False after the start_server_step2.

```
[May 20 2017 16:13:16] OnionShareWeb.stay_open: stay_open=True
[May 20 2017 16:13:16] OnionShareGui.stay_open: stay_open=True
[May 20 2017 16:13:16] OnionShare.start_onion_service
[May 20 2017 16:13:16] Onion.start_onion_service
Configuring onion service on port 17607.
[May 20 2017 16:13:52] OnionShareGui.start_server_step2
[May 20 2017 16:13:52] OnionShareWeb.stay_open: stay_open=False
```

This moves the config check to the place where it matters.

### DIFF
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -221,9 +221,7 @@ class OnionShareGui(QtWidgets.QMainWindow):
 
             # start onionshare http service in new thread
             # First, load settings and configure
-            settings = Settings()
-            settings.load()
-            self.app.stay_open = not settings.get('close_after_first_download')
+            self.app.stay_open = not self.settings.get('close_after_first_download')
             common.log('OnionShareGUI', 'stay_open', 'stay_open={}'.format(self.app.stay_open))
             t = threading.Thread(target=web.start, args=(self.app.port, self.app.stay_open))
             t.daemon = True
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -203,7 +203,6 @@ class OnionShareGui(QtWidgets.QMainWindow):
         settings = Settings()
         settings.load()
         self.app.set_stealth(settings.get('use_stealth'))
-        web.set_stay_open(not settings.get('close_after_first_download'))
 
         # Reset web counters
         web.download_count = 0
@@ -221,6 +220,11 @@ class OnionShareGui(QtWidgets.QMainWindow):
                 return
 
             # start onionshare http service in new thread
+            # First, load settings and configure
+            settings = Settings()
+            settings.load()
+            self.app.stay_open = not settings.get('close_after_first_download')
+            common.log('OnionShareGUI', 'stay_open', 'stay_open={}'.format(self.app.stay_open))
             t = threading.Thread(target=web.start, args=(self.app.port, self.app.stay_open))
             t.daemon = True
             t.start()


micahflee:
Oh, great catch 👍 . If only I had written a test for this...

So it looks like this will fix the issue, but I think there can be a slightly more efficient way of doing it that doesn't involve reloading settings from disk each time you start the server.

The `OnionShareGui` class already has a `self.settings` class variable that's initialized in the constructor: https://github.com/mig5/onionshare/blob/f446f6630dc4c096e17c6816b64ae7d640405bfa/onionshare_gui/onionshare_gui.py#L59

It looks like the only other time that `self.settings` is referenced is when it checks for available updates: https://github.com/mig5/onionshare/blob/f446f6630dc4c096e17c6816b64ae7d640405bfa/onionshare_gui/onionshare_gui.py#L318

But now we make a new Settings object in the `start_server` function and also in the `start_onion_service` function within the `start_server` function, so this code is repeated twice, and it's not using the class variable `self.settings` either:

```python
# First, load settings and configure
settings = Settings()
settings.load()
```

I think it would make sense to refactor this to only use `self.settings` every time we need to reference settings. It's also possible that the user might open the SettingsDialog and change something, in which case the settings should be reloaded. So, maybe here, right after opening the settings dialog in the `open_settings` function: https://github.com/mig5/onionshare/blob/f446f6630dc4c096e17c6816b64ae7d640405bfa/onionshare_gui/onionshare_gui.py#L187

mig5:
Yeah I had a feeling this wasn't right. I, too, noticed yesterday (in #403) that changing a setting after stopping a share, and then starting the share again, didn't take into account the changed setting. 

I've modified this to simply use self.settings.get() and it still seems to work for me.

I need to get my head around the latter part of open_settings - far as I can tell the settings are *saved* but not reloaded after the settings dialog is *closed*.. seems to make sense to re-load *after* the Settings.save_clicked event is triggered?

micahflee:
The issue here is that the `OnionShareGui` object has its own `Settings` class variable called `self.settings`, but when you open the `SettingsDialog` it doesn't pass it in -- instead, the `SettingsDialog` creates a brand new temporary `Settings` object and works with it there.

So if you add `settings.load()` to at the end of the `Settings.save_clicked` function, you still don't modify the `OnionShareGui.settings` object -- that's still the old version of the object, with the old settings.

One way to fix this is to use [signals and slots](http://doc.qt.io/qt-5/signalsandslots.html). We can create a "settings have changed" signal in the `SettingsDialog` object, and emit it as soon as the settings have changed. Then the `OnionShareGui.open_settings` function can connect a slot (a function) to that signal, that reloads settings when its emitted.

Another, simpler, way to fix it to to simply reload the settings after open the `SettingsDialog`, like this:

```python
def open_settings(self):
    """
    Open the SettingsDialog.
    """
    SettingsDialog(self.onion, self.qtapp)
    self.settings.load()
```

This will actually work only because `SettingsDialog` is a modal dialog, which basically means it blocks everything in the window that launches it until the dialog is closed.

However, the former is actually the better solution. With the latter, the settings will _always_ reload as soon as you close the Settings dialog, but with the former it will only reload settings when you click Save, but not when you click Cancel.

micahflee:
Alright, I've implemented this in #409, mostly in this commit: https://github.com/micahflee/onionshare/pull/409/commits/5846b06478994caa71b1fffe43574a27ef064563

mig5:
Thanks - now it all makes sense, it was the syntax in open_settings() that I wasn't sure exactly how to write. Working great

PR #401, opened 2017-05-20
Reinstate the GUI method of shutting down the web server

mig5:
I discovered that clicking 'Stop Sharing' from the UI was not actually stopping the Onion web server! Eek! (First I noticed that a reload of the onion site still loaded in my Tor Browser, and sure enough, lsof confirmed that the port was still open despite the debug message saying it had stopped and cleaned up)

I think the method that allows the GUI to send that request (out of context of the web server environment) was accidentally removed re: removing transparent torification in https://github.com/micahflee/onionshare/commit/bb990ff57484129db26cf43127a58e4e252c9f5f

This reinstates it in the 'except' catch, so that the GUI can still have a method of shutting down the server when we tell it to :)

### DIFF
--- a/onionshare/web.py
+++ b/onionshare/web.py
@@ -371,4 +371,7 @@ def stop(port):
         s.connect(('127.0.0.1', port))
         s.sendall('GET /{0:s}/shutdown HTTP/1.1\r\n\r\n'.format(shutdown_slug))
     except:
-        pass
+        try:
+            urlopen('http://127.0.0.1:{0:d}/{1:s}/shutdown'.format(port, shutdown_slug)).read()
+        except:
+            pass


micahflee:
Like #399, this is blocked by #404.

mig5:
can I recommend adding this to the 1.1 milestone :)

micahflee:
Oh, does automatically stopping not currently work? I think that broke in #399, which I didn't merge, but I didn't manually test it myself. If this is currently broken, then yes definitely it should be in 1.1 :).

mig5:
Automatically is OK, the service shuts down..  but if you explicitly press the 'Stop sharing' button, the UI states that the server is stopped, but the onion hidden service keeps running :) so rather severe.

See the first comment above, I think it's a regression from removing transparent torification in https://github.com/micahflee/onionshare/commit/bb990ff57484129db26cf43127a58e4e252c9f5f - a chunk of logic was taken out of the shutdown function, so as of 1.1/master branch, the web server never gets to make its Flask shutdown request if the UI said to stop.

#399 shows the removal of urllib only because that's the part that was accidentally removed, so pyflakes picked up that the module 'isn't required' (but the logic should go back in). So #399 probably needs a code update if this one here is merged.

Happy to chat about it if I've confused you!

mig5:
Or to put another way: you removed the 'if transparent_torification' condition, leaving this as the default shutdown attempt:

```
        s = socket.socket()
        s.connect(('127.0.0.1', port))
        s.sendall('GET /{0:s}/shutdown HTTP/1.1\r\n\r\n'.format(shutdown_slug))
```

All I know is that that doesn't seem to work. The 'except' side of the 'try' catch was simply 'pass', so it didn't try anything else.

The fix in this PR puts the original alternative method of shutting down the Flask server into the except catch:

```
   urlopen('http://127.0.0.1:{0:d}/{1:s}/shutdown'.format(port, shutdown_slug)).read()		
```

... which *does* seem to work. I dunno, maybe it should be reversed so that this urlopen() method is the type that is tried first? (or the only type?)

micahflee:
Totally makes sense now

Issue #285, opened 2016-04-25
Notification when download begins


anseljh:
Feature suggestion: do a desktop notification when a download begins (and maybe when it finishes). This would be nice when you're waiting for someone to download, and don't want to watch the UI.

freddymartinez9:
Just to clarify @anseljh do you think the person initiating the download should get the notification? If I'm running a long standing .onion that is popular, I wouldn't want to get notifications every single time a user started downloading my file.

anseljh:
The person serving would get a notification. I agree this wouldn't be useful when you leave the share open, but for one-off shares (OnionShare's default), it would be.

micahflee:
I think this is a cool idea.

micahflee:
Implemented!

PR #403, opened 2017-05-22
Add desktop notification messages on download events (via systray)

mig5:
This implements the desktop notification feature as per #285 

A notification appears when the user begins, cancels, or finishes the download.

As this might become annoying if you share a file to many users, I added a share setting that can optionally disable the notifications. It's enabled by default, on the assumption that most people are sending one-to-one shares and sporadically.

In order to implement this, a systray icon gets added to the user's systray. Right now, I added only a single 'Quit' item, just so it didn't seem weird to have an icon that 'does' nothing. (The user probably wouldn't realise it's responsible for sending the notifications)

I wrap the message in a supportsMessages() check, but according to the docs [1], most systems should support this (MacOS < 10.8 would need Growl). Testing on Windows or Linux is appreciated, I've only done so on Mac.

screen shot 2017-05-22 at 4 14 13 pm

[1] http://doc.qt.io/qt-5/qsystemtrayicon.html#details


micahflee:
Nice, this is great! I made a few changes FYI:

- Clicking Quick in the systray menu now calls `self.close()` instead of `sys.exit()`, because this lets OnionShare shutdown gracefully, and also warn the user if they're currently running a server
- In macOS, the icon is grayscale, to fit the macOS systray icon convention
- I updated the wording of the text

Issue #404, opened 2017-05-22
Write way more tests


micahflee:
I've done a terrible job at writing tests for OnionShare. There should be way more tests, both in the `onionshare` module (especially the `onionshare.Onion` class and the `onionshare.web` module) and in `onionshare_gui`.

I haven't written Qt tests before, but it looks like these might be helpful resources:

- http://doc.qt.io/qt-5/qtest-tutorial.html
- https://github.com/pytest-dev/pytest-qt
- http://johnnado.com/pyqt-qtest-example/

micahflee:
Related to #164. FYI, I'm not married to nosetests, I'm happy with using whatever framework works best.

mig5:
Hi @micahflee,

I've made limited progress with tests. I have to confess I'm too much of a noob to know what I'm doing, so I thought I'd leave something that hopefully gets you or someone better than me started.

1) Download this gist and put it in the root of the onionshare dir (this is how much of a noob I am - I couldn't figure out how to import the relative modules properly by putting it in the test dir :) )

https://gist.github.com/mig5/90986df2d5ec10d60fc9514b5d5ae501

2) in onionshare/common.py I had to hack something up for the get_resource_path() prefix in order to load locales. This was done on a Linux machine, and the if/else conditional was just not matching

```
 prefix = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))), 'share')
```

probably the test could be modified to pass the onionshare_dev_mode attribute to workaround this...

3) Finally, you can run 'python3 oniontest.py', which tests:

a) the gui loaded (which also will imply the tor connection worked, as it's part of the setUpClass that runs before any tests)

b) the window title is 'OnionShare'

c) the download container is hidden

d) the settings dialog button is enabled

e) the status bar is visible

f) the number of files in the file dialog (our test.txt) is 1

I could not for the life of me work out how to get the QPushButtons to be pressed via QtTest. That will be the real magic so that we can then tests ports are open/closed etc on share start/end events

Hopefully it inspires someone to take this further/do it properly for the UI tests.

(Side note - I get a lot of 'Resourcewarning' on unclosed files, generally locale and torrc files.. seems mostly harmless)

I hope that this isn't completely wrong and a waste of your time, but I was pleased at least that I could get some passing tests.

mig5:
[...]

PR #411, opened 2017-05-23
[WIP] Add a series of basic qTest GUI tests

mig5:
This script from #404 is now able to be run from the root of the repo with 

```python3 test/onionshare_gui_test.py```

```
.....Configuring onion service on port 17646.
 * Running on http://127.0.0.1:17646/
...
----------------------------------------------------------------------
Ran 8 tests in 96.044s

OK
```

It needs more work, I have not been able to make it stop the service and check if the port is still open (e.g a test for #401)

They could also arguably be broken up into individual files if you prefer, but having an easy entry point into running a bunch of tests seems useful for automation. You can apparently 'group' tests with unittest but I haven't looked into it (I learnt everything done here this afternoon :))

### DIFF
--- a/test/onionshare_gui_test.py
+++ b/test/onionshare_gui_test.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python3
+import os, sys, unittest, inspect, time
+from PyQt5 import QtCore, QtWidgets, QtGui, QtTest
+
+sys.path.append('../onionshare')
+sys.path.append('../onionshare_gui')
+
+from onionshare import onion, strings, common
+from onionshare_gui import *
+
+app = QtWidgets.QApplication(sys.argv)
+
+class OnionShareGuiTest(unittest.TestCase):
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: test/onionshare_gui_test.py [hunk 1]
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
+import os, sys, unittest, inspect, time
+from PyQt5 import QtCore, QtWidgets, QtGui, QtTest
+
+sys.path.append('../onionshare')
+sys.path.append('../onionshare_gui')
```
mig5: I know these appends are ugly but I could not get the test to run otherwise, unless the test file itself was *also* in the root of the repo

mig5:
Lots of trial and error commits buried in here, and can't get this into good shape for Travis etc either. Closing, the tests will remain in my fork if there's anything of use.

PR #413, opened 2017-05-25
Minor update to `common.get_available_port`

delirious-lettuce:
* Use context manager inside of `get_available_port`
* [Rearrange imports, one per line](https://www.python.org/dev/peps/pep-0008/#imports)
* Reuse original `import random` for `random.SystemRandom` (instead of a separate `from random import SystemRandom`)
* [Two blank lines above each function definition](https://www.python.org/dev/peps/pep-0008/#blank-lines)


mig5:
:+1:

PR #415, opened 2017-05-25
[WIP] 91% test coverage of `onionshare/common.py`

delirious-lettuce:
@micahflee ,

This PR is still a work-in-progress but I wanted to give you an idea of what I've been doing to help with #404 . I will continue to work on it to try and achieve 100% coverage.

Let me know what you think!

PR #416, opened 2017-05-26
attempt to run headless pytests on a tor-enabled TravisCI environment

mig5:
Curious to see if this change gets Travis to run the tests successfully

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
 - echo "deb http://deb.torproject.org/torproject.org trusty main" | sudo tee -a /etc/apt/sources.list.d/torproject.list
 - gpg --keyserver keys.gnupg.net --recv A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
 - gpg --export A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89 | sudo apt-key add -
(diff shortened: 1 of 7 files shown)


mig5:
This PR contains a 'local_only' variant of the main GUI test. For the tests that require Tor for their full GUI test, they have @pytest markers that say to skip - but only when it's Travis that would otherwise try and run them (it checks for the TRAVIS environment variable that Travis sets). 

In other words, if you run the dev_scripts/tests script in your local env, they'll still run there because you're not Travis (unless of course you run ```TRAVIS=true dev_scripts/tests```)

As a result, this test has finally passed because it ran:

1) just the local version of the main start/stop GUI test
2) the 'settings' GUI test
3) the original pre-Qtest tests (strings, common etc)

I note that even if you pass --local-only, if Onionshare's default Tor connection method is 'bundled', it will still attempt a tor bootstrap.. as it does here. Maybe it should be using the control port etc of the tor that Travis installs. But I don't think that can be passed as an arg, probably just presupplied json settings.

https://travis-ci.org/micahflee/onionshare/jobs/236244923

micahflee:
Nice!

Like with #415, I think I'm going to wait until after the 1.1 release to merge in these tests. Also, at some point I'd like to merge the two divergent test branches from you and @delirious-lettuce :).

mig5:
No prob - I saw it still fails - failing to bootstrap tor even on a local_only test is enough to break it.

I want to see if we can pass args for tor settings on start up (which override the json) so it can use a control socket from the outset.

PR #419, opened 2017-05-27
Adds Settings and Help menu items to the systray

mig5:
The systray looked a little bare to me with just the Quit button. Added a couple more.

Not sure if you are happy with re-using the same locale string for these, rather than create new systray-specific ones. These aren't technically 'buttons' but 'actions'. Let me know, or feel free to reject if this is a pointless idea :) (also wasn't sure about instantiating a new SettingsDialog for the sake of clicking Help - maybe that Help action needs abstracting out of settings altogether and into the main OnionShareGui)

### DIFF
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -150,6 +150,10 @@ class OnionShareGui(QtWidgets.QMainWindow):
         system = common.get_platform()
 
         menu = QtWidgets.QMenu()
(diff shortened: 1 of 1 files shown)


micahflee:
Nice, I like it! It might be slightly cleaner to abstract out the help button from Settings, but on the other hand this is simple and works, and it doesn't even seem to run `SettingsDialog.__init__` or load settings from disk or anything. So I think I'm just going to merge this, and if you'd like to refactor how clicking help works, I'll gladly merge that too.

PR #421, opened 2017-05-27
Style a prettier progress bar

mig5:
I thought the progress bar could be improved from the matte blue.

I made a purple (Tor inspired) progress bar with a gradient which feels a bit more pleasant. Your feedback welcome. The only thing I wasn't sure about was the implications for colorblind/vision impaired users.

Edited for 'Crunching files' too, but it's a tighter progress bar, doesn't look quite as good, but maybe better to be consistent with this than leave it as it is.

purpleprogress

Mac OS

screen shot 2017-05-27 at 9 03 39 pm

### DIFF
--- a/onionshare_gui/downloads.py
+++ b/onionshare_gui/downloads.py
@@ -32,14 +32,24 @@ class Download(object):
         self.downloaded_bytes = 0
 
         # make a new progress bar
+        cssStyleData ="""
+        QProgressBar {
+            border: 2px solid grey;
+            border-radius: 5px;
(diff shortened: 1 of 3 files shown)


micahflee:
Ooh, so pretty! I love the colors.

mig5:
I thought about doing the tor connection dialog too (for consistency) but decided that that actually looks better with the slim grey bar on Mac, and it felt wrong to add another 'platform' conditional for such a superficial thing.

PR #422, opened 2017-05-27
Remove border around items in the QStatusBar

mig5:
On Mac, the version label and Settings cog icon sit neatly in the status bar seamlessly.

On Linux, I noticed that each widget had a border around it, which is much uglier.

This change forces the border to 0px which fixes it on Linux. I checked my Mac and it doesn't seem to change anything.

Not sure about Windows but if it was an issue there, hopefully this fixes it too.

before:
os_statusbar_border_linux

after:
os_statusbar_noborder_linux

the change in icon in top left doesn't seem related to this change (after noticing it, I reverted my change and it was still missing).  Something about focus on startup. Blame Linux/Qubes for this wtf! :)

### DIFF
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -101,6 +101,8 @@ class OnionShareGui(QtWidgets.QMainWindow):
         # Status bar
         self.status_bar = QtWidgets.QStatusBar()
         self.status_bar.setSizeGripEnabled(False)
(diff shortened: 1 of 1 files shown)


micahflee:
Interesting. I actually don't see the border on my linux (fedora), but in any case this commit can't cause any issues, so I'm merging it.

mig5:
Curious, it occurred on my Fedora24 too.

PR #426, opened 2017-05-28
`hmac.compare_digest` instead of `constant_time_compare`?

delirious-lettuce:
@micahflee ,

I noticed [this link](http://www.levigross.com/2014/02/07/constant-time-comparison-functions-in...-python-haskell-clojure-and-java/) inside the function definition of `constant_time_compare` and saw the comment above it mentioning that this function was built-in to Python. Searching the `cpython` repository showed me this function was accessible through `hmac.compare_digest`.

[hmac.compare_digest(a, b)](https://docs.python.org/3.6/library/hmac.html?highlight=compare_digest#hmac.compare_digest)

> Return a == b. This function uses an approach designed to prevent timing analysis by avoiding content-based short circuiting behaviour, making it appropriate for cryptography. a and b must both be of the same type: either str (ASCII only, as e.g. returned by HMAC.hexdigest()), or a bytes-like object.
> 
> Note If a and b are of different lengths, or if an error occurs, a timing attack could theoretically reveal information about the types and lengths of a and b—but not their values.
> 
> New in version 3.3.

The only usage I could find of `constant_time_compare` in `onionshare` was in `web.py` where both `slug_compare` and `slug_candidate` have to be encoded before comparison.

```python
if not common.constant_time_compare(slug_compare.encode('ascii'), slug_candidate.encode('ascii')):
    abort(404)

```

`hmac.compare_digest` seems to be a bit more flexible since you do not have to encode a Python 3 `str` to `bytes` before using it. Also, it seems to be implemented in C which might be faster than a Python version.

```python
>>> import hmac
>>> constant_time_compare(b'bytes', b'bytes')
True
>>> hmac.compare_digest(b'bytes', b'bytes')
True
>>> constant_time_compare('py3 str', 'py3 str')
  File "<input>", line 1, in <module>
  File "<input>", line 17, in constant_time_compare
TypeError: unsupported operand type(s) for ^: 'str' and 'str'
>>> hmac.compare_digest('py3 str', 'py3 str')
True

```

I'm not sure if there is a particular reason not to use `hmac.compare_digest` so I made this PR just in case using a function from the Python standard library was a better option.

### DIFF
--- a/onionshare/common.py
+++ b/onionshare/common.py
@@ -99,26 +99,6 @@ def get_version():
         return version
 
 
-def constant_time_compare(val1, val2):
-    """
-    Returns True if the two strings are equal, False otherwise.
-
-    The time taken is independent of the number of characters that match.
-
-    For the sake of simplicity, this function executes in constant time only
-    when the two strings have the same length. It short-circuits when they
-    have different lengths.
-
-    From: http://www.levigross.com/2014/02/07/constant-time-comparison-functions-in...-python-haskell-clojure-and-java/
-    """
-    if len(val1) != len(val2):
-        return False
-    result = 0
-    for x, y in zip(val1, val2):
-        result |= x ^ y
-    return result == 0
-
-
 def random_string(num_bytes, output_len=None):
     """
     Returns a random string with a specified number of bytes.
--- a/onionshare/web.py
+++ b/onionshare/web.py
@@ -18,7 +18,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from distutils.version import StrictVersion as Version
-import queue, mimetypes, platform, os, sys, socket, logging
+import queue, mimetypes, platform, os, sys, socket, logging, hmac
 from urllib.request import urlopen
 
 from flask import Flask, Response, request, render_template_string, abort, make_response
@@ -162,7 +162,7 @@ def check_slug_candidate(slug_candidate, slug_compare = None):
     global slug
     if not slug_compare:
         slug_compare = slug
-    if not common.constant_time_compare(slug_compare.encode('ascii'), slug_candidate.encode('ascii')):
+    if not hmac.compare_digest(slug_compare, slug_candidate):
         abort(404)
 
 


garrettr:
@micahflee asked me to take a look at this. Here are my thoughts:

## Backwards compatibility with Python < 3.3

The main reason why you might not want to use `hmac.compare_digest` is if you need compatibility with Python versions < 3.3. A common idiom is to prefer `hmac.compare_digest` if it's available, and otherwise fall back on a bundled Python constant-time compare function, e.g.:

```py
try:
    from hmac import compare_digest as constant_time_compare
except ImportError:
    # Python version < 3.3
    from common import constant_time_compare
```

Of course, if you don't need compatibility with earlier Python versions, you can ignore this backwards compatibility concern; however, since Onionshare on Linux uses the Python version provided by the distribution, it might be difficult to guarantee that you don't need backwards compatibility with earlier Python versions.

## Other benefits of `hmac.compare_digest`

I think the main benefits of using `hmac.compare_digest` are:

1. Results in less code for OnionShare's developers to maintain.
2. It's included in the standard library, so is more likely to have been scrutinized for potential bugs that could compromise its security than a custom version only used by OnionShare.
3. Is written in C and uses a couple of low-level techniques that are only available in C to make it "even more constant time." I'm not sure if these really make a difference.

## Non-benefits of `hmac.compare_digest`

> hmac.compare_digest seems to be a bit more flexible since you do not have to encode a Python 3 str to bytes before using it.

That's only because it's handling the conversion for you - the conversion still happens, you just don't have to do it. However, the auto-conversion is [limited to Unicode strings that contain only ASCII characters](https://github.com/python/cpython/blob/6f0eb93183519024cb360162bdd81b9faec97ba6/Modules/_operator.c#L815-L820), so you don't really gain much here in my opinion. If anything, I personally prefer such transformations on security-sensitive data to be explicit and up-front as much as possible, rather than buried in library code.

Fundamentally, both `common.constant_time_compare` and `hmac.compare_digest` are using the same "xor/or" trick, so the underlying values being compared need to be bytes (or easily converted into bytes) so that they may be `xor`ed.

> Also, it seems to be implemented in C which might be faster than a Python version.

That's possible, but would only be consequential if the code were frequently comparing very large strings, which is not the case in OnionShare.

delirious-lettuce:
@garrettr ,

Great points, thank you for taking the time to explain! I am totally new to `hmac.compare_digest` and I just came across it as I was reading through the code and found that attribution link in the comments.

You make a good point about it not being available in versions < 3.3. When I looked at `.travis.yml`, I noticed `onionshare` was only being tested for `3.4+` and that's part of the reason I brought this up. Also, many of the files (except `check_lacked_trans.py` which doesn't seem to be used?) use the `print` function without any `from __future__ import print_function` so I did make a bit of an assumption about this being a Python 3.4+ project.

As for the conversion, it seemed like encoding to ascii had the same result whether using `common.constant_time_compare` or `hmac.compare_digest` so I'm unsure if this is really a non-benefit on `hmac.compare_digest`. Isn't it the same behaviour for both?

```python
>>> 'abc'.encode('ascii')  # as used in `constant_time_compare` function call
b'abc'
>>> 'aÄc'.encode('ascii')
  File "<input>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character '\xc4' in position 1: ordinal not in range(128)

```

Anyways, I really appreciate the explanation and I am by no means an expert. My thinking was just along the lines of your point # 2:

> "It's included in the standard library, so is more likely to have been scrutinized for potential bugs that could compromise its security than a custom version only used by OnionShare."

Should I close this PR?

micahflee:
It's not really formalized, but I've basically decided I'd like OnionShare to work with Debian stable, and not worry about software older than that. The current version of python3 in jessie is 3.4.2, so I'm happy to not support older versions of python.

So in this case, backwards compatibility isn't an issue. That said, I actually think this is fine to merge as-is. It's cleaner to have less code included in OnionShare if the same thing exists in the standard library.

delirious-lettuce:
@micahflee ,

Ah ok, makes sense. I'll keep that in mind (aiming for Debian stable) for future ideas. I'll adjust the test cases to reflect this new merge.

Also, I noticed that I was using `pytest-3.1.0` for the tests but that you were using `pytest-2.9.2`. I'm going to downgrade my `pytest` version and ensure that the tests run for you.

Thanks again for your time and input! I've learned a bunch of new things reading your code in this project (and blog for that matter)!

PR #427, opened 2017-05-29
Make adding of files/folders a single button

mig5:
This change makes a single 'Add' button which works for adding either just files, just folders, or both at once.

This also has a nice side-effect: the non-native 'add folders' dialog introduced in #402 will no longer look different to the native 'add files' dialog - it's the one dialog. *Edit* didn't realise you actually made that consistent in e828b2f9cdf85b597c3ca244033f2df6b74ae82c so maybe this is not so significant an idea :)

Just an idea, up to you :)

onionshare_single_add


mig5:
Sorry, nevermind, this breaks choosing an individual file (sigh)

mig5:
Think I have this fixed now for selecting either an individual file, individual folder, or multiples of either/both. Hope someone can test!

mig5:
The theory for this was sourced from

https://stackoverflow.com/questions/6484793/multiple-files-and-folder-selection-in-a-qfiledialog
http://widequestion.com/question/multiple-files-and-folder-selection-in-qfiledialog-revisited/

micahflee:
Nice, this is awesome! I just tested it and it appears to work fine. This is actually what I originally wanted to do, but eventually I gave up and made two separate buttons for adding files and folders.

micahflee:
And good job on the locales :)

mig5:
I made a 'Translating' Howto page in the wiki after that ;)

PR #429, opened 2017-05-30
Hide the Onion URL behind a 'Reveal' button that can reveal/hide the Reveal URL when toggled

mig5:
Just an idea.

To avoid risk of passive collection, the Onion URL could be hidden behind a QPushButton which, when toggled, reveals the URL (and hides it again if toggled once more).

This also has the effect of not stretching the window (and the button widgets) automatically (at least, not so dramatically) once the service is ready (I had another tentative change to set a max width on the buttons so they don't get so long, but it's not perfect).

The idea mainly came from working on #427 - once there are only two buttons, that stretch is a bit more obvious, and then I thought perhaps we can hide that quite sensitive URL by default in any case.

### DIFF
--- a/onionshare_gui/server_status.py
+++ b/onionshare_gui/server_status.py
@@ -57,17 +57,15 @@ class ServerStatus(QtWidgets.QVBoxLayout):
         server_layout.addWidget(self.server_button)
 
         # url layout
-        url_font = QtGui.QFont()
-        self.url_label = QtWidgets.QLabel()
-        self.url_label.setFont(url_font)
-        self.url_label.setWordWrap(False)
-        self.url_label.setAlignment(QtCore.Qt.AlignCenter)
+        self.reveal_url_button = QtWidgets.QPushButton(strings._('gui_reveal_url', True))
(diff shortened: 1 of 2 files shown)


micahflee:
This is an interesting idea. I've been working with a UX designer who's done a bunch of excellent work on redesigning the UI (including doing a full user study to gather data). I haven't implemented many of his changes yet (only a few minor ones) -- I'm waiting until I release 1.1 before starting on that.

But since this is a bigger UX change, I think I'm going to wait to merge it until I'm working on other UX stuff, and decide if this is actually the most usable thing.

mig5:
No probs! Please feel free to reject more of my UI stuff if you are planning a more major refactor that would render it obsolete. (Not trying to rack up commits in your project :) )

micahflee:
Thanks for the work on this PR, but I think I'm going to reject it. I'm not considering shoulder surfing part of the threat model of OnionShare -- especially since, in order to effectively use OnionShare, you're going to have to copy and paste that URL somewhere at some point, and when you do that the URL will be on your screen again.

Issue #428, opened 2017-05-30
incorrect files being hosted


Baccount:
When I add 4 files and start hosting them and then click the stop hosting, then remove one file and start hosting the files again. The files that are hosted are not he files that I have in onionshare.
Here is a photo showing the bug
![issue](https://cloud.githubusercontent.com/assets/10781882/26566225/5de3aa18-44a6-11e7-9702-463f135d4f44.png)

mig5:
Hi @Baccount, thanks for reminding me about this, I saw it the other day too.

I have made a potential fix at https://github.com/micahflee/onionshare/pull/430 - if you are in a position to test it (e.g you are able to build on OS X as per the BUILD.md), that would be great! (No problem if not)

mig5:
Actually - confused - my change doesn't do anything, in fact I don't think I reproduce this on master branch anymore?

1) Add 4 files
2) Share
3) Stop Sharing
4) Remove 1 file
5) Start Sharing
6) See only the remaining 3 files in the new share

Strange. Sorry for the confusion, I didn't actually check if I still reproduced this as I recall reproducing it some days ago..

Baccount:
I'm currently using 1.1 dev 2 and I'm able to reproduce it every time following the steps that you mentioned above. I'm not sure if this has been fixed in the master branch or not,  I haven't tested it.

micahflee:
I'm having trouble reproducing this. I'm using the Mac version of OnionShare 1.1.dev2, just like you. I started OnionShare and added four files:

- `2004-01-27_SIDToday_-_The_Role_of_Technical_Directors_in_AP.pdf`
- `2004-07-01_SIDToday_-_Class_for_Grade_14_and_Above_Denial_and_Deception_Executive_Overview.pdf`
- `2004-07-02_SIDToday_-_A_Patriots_Story.pdf`
- `2004-07-02_SIDToday_-_Making_It_Easy_repost.pdf`

Then I shared it. Here's what it looks like in Tor Browser:

<img width="1092" alt="screen shot 2017-05-30 at 12 47 58 pm" src="https://cloud.githubusercontent.com/assets/156128/26601815/2fe5d92a-4536-11e7-9ffd-aedde34bde5e.png">

Then I stopped the server and deleted `2004-07-02_SIDToday_-_A_Patriots_Story.pdf`, and started it again. Here it is in Tor Browser, and also in OnionShare:

[...]

PR #432, opened 2017-05-31
Ensure the self.filenames list is sorted internally.

mig5:
This is important because even though the QListWidget UI is sorted
automatically, the list is not necessarily. Drag-drop events, depending
on the order in which items were highlighted before being dragged,
can result in a different or reversed order (e.g on a Desktop, if you select 5.txt through 1.txt in a vertical list, rather than top-down, the self.filenames list order was 5.txt, 4.txt, 3.txt, 2.txt, 1.txt). This has implications for
popping the list later (e.g on delete events - if you mark 4.txt to be deleted, the row that is actually popped internally is that of 2.txt)

Fixes https://github.com/micahflee/onionshare/issues/428

### DIFF
--- a/onionshare_gui/file_selection.py
+++ b/onionshare_gui/file_selection.py
@@ -136,6 +136,8 @@ class FileList(QtWidgets.QListWidget):
                 return
 
             self.filenames.append(filename)
+            # Re-sort the list internally
+            self.filenames.sort()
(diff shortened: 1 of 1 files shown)


micahflee:
This (and all your other PRs) look great, thank you.

PR #455, opened 2017-07-08
WIP: 40% overall test coverage of `onionshare.onionshare`

delirious-lettuce:
@micahflee ,

I apologize for the delay with this pull request (to help with #404 ). When I saw that message (a few weeks ago) that the Tails guys were writing tests, I just kind of assumed they would do a better job than me so I stopped working on it. When I saw your message a few days ago clarifying what they were actually testing, I thought I would start back up on these tests.

Here is a summary of the coverage so far (inside of PyCharm):

onionshare_coverage

The highlighted files are the ones I've been writing/modifying the tests for. I figured out why that issue was happening to you before. For `pytest <= 2.9`, you have to use `@pytest.yield_fixture` instead of `@pytest.fixture` (for fixtures that use `yield` statements). I also downgraded my Python version and `pytest` version to try and match yours to ensure the tests will pass for you.

```
============================= test session starts ==============================
platform linux -- Python 3.5.3, pytest-2.9.2, py-1.4.34, pluggy-0.3.1
```

Here is a link to the latest Travis build showing the tests passing (and coverage of `onionshare` using the `pytest-cov` plugin) on all desired Python versions, `3.4+`.
* https://travis-ci.org/delirious-lettuce/onionshare

python_34_travis

This is still a work-in-progress and I haven't updated the setup info quite yet but I will do that later today. As you can see, I haven't covered all of the files yet but there are some good reasons.

* `web.py` I'm not exactly sure how to properly test a Flask app and will have to do more research on that.
* `onion.py` I'm also not exactly sure how to properly test the Tor parts (since I'm not very experienced with Tor overall). 
* `socks.py` Again, I'm not exactly sure how to properly test networking and sockets

Overall, it might be easier for me just to finish the setup commands for this pull request. I've been playing with the travis settings using `pip install` instead of system packages because of error relating to this:

* https://docs.travis-ci.com/user/languages/python/#Travis-CI-Uses-Isolated-virtualenvs

    > CI Environment uses separate virtualenv instances for each Python version. System Python is not used and should not be relied on. If you need to install Python packages, do it via pip and not apt.
    >
    > If you decide to use apt anyway, note that Python system packages only include Python 2.7 libraries on Ubuntu 12.04 LTS. This means that the packages installed from the repositories are not available in other virtualenvs even if you use the –system-site-packages option.

Coveralls is not needed (I was only using it for my own reference) and I can easily take it out of `.travis.yml`.

---
I am open to any input/suggestions on how to improve these tests since I am not an expert with `pytest`. I have learned a lot of things while writing them but if you have any questions about why I did something in a particular way, feel free to ask (and/or correct me!).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,5 @@ install:
   - pip install Flask==0.12 stem==1.5.4 pytest-cov coveralls
 # command to run tests
 script: pytest --cov=onionshare test/
-# after_success:
-#   - coveralls
+after_success:
+  - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-sudo: required
+# sudo: required
 dist: trusty
 python:
   - "3.4"
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,6 @@ python:
 install:
   - pip install Flask==0.12 stem==1.5.4 pytest-cov coveralls
 # command to run tests
-script: pytest test/ --cov test/
-after_success:
-  - coveralls
+script: pytest --cov=onionshare test/
+# after_success:
+#   - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,10 @@ python:
   - "3.7-dev"
   - "nightly"
 # command to install dependencies
-before_install: "sudo apt-get update; sudo apt-get install -y python3-nose python3-flask python3-stem python3-pyqt5"
-install: ""
+# before_install: "sudo apt-get update; sudo apt-get install -y python3-nose python3-flask python3-stem python3-pyqt5"
+install:
+  - pip install Flask==0.12 stem==1.5.4 pytest-cov coveralls
 # command to run tests
-script: nosetests3
+script: pytest test/ --cov test/
+after_success:
+  - coveralls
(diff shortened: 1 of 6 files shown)


micahflee:
This is excellent work, thank you so much! I also am not an expert with pytest, but all your tests pass, and I reviewed them and them seem to be solid.

I know there's plenty more work to do, but I'm going to go ahead and merge these. The reason is because @mig5 has made a handful of other PRs that I'd like to merge too, and I just feel better have more solid unit tests around when making any changes.

But please, continue this work and submit more PRs with tests and I'll merge them.

delirious-lettuce:
@micahflee ,

I really appreciate the opportunity to help on a project like this. I have been writing more tests and learning how to structure them in a better way so I will definitely be sending more PRs in the near future (should be a few in the next couple days).

I have also been experimenting with using [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html) which may be an answer to test some of the more advanced parts of the code (like classes, etc). It is in Python 3s standard library so it's not an extra dependency. There is also [`pytest-mock`](https://pypi.python.org/pypi/pytest-mock) (a wrapper for `unittest.mock`) but since that becomes an extra dependency that must be `pip` installed, I thought I'd leave that out for now (mock tests should be somewhat straightforward to convert later, if necessary).

My goal is to get 100% test coverage of `onionshare.onionshare` ASAP.

Thanks again for your time, I'm glad the code was acceptable!

Issue #442, opened 2017-06-03
Flask version check is incorrect


davidism:
`StrictVersion` is not PEP 440 compliant, and fails with our dev tags (`'0.13-dev'`). Arch's onionshare package allows python-flask-git as the flask dep, which has a `-dev` tagged version string.

Use `pkg_resources.parse_version`, which is what pip uses.

~~~python
import flask
import pkg_resources
pkg_resources.parse_version(flask.__version__)

# or by resolving the distribution
pkg_resources.get_distribution('flask').version
~~~

Or if you don't want to depend on setuptools, use `distutils.version.LooseVersion`, which is less correct but still works.

georgeowell:
This is still causing Onionshare AUR package to fail to build.

https://github.com/pallets/flask/issues/2355
https://bugs.archlinux.org/task/54288

Seems it is an Arch upstream issue. Not sure if there is a way we can bump this up their priority list. Quite frustrating.

davidism:
The quick fix is installing `python-flask` instead, which was repackaged with a correct version in Arch recently. It doesn't really make sense to install `python-flask-git` system-wide. But this should still be fixed anyway.

georgeowell:
This still hasn't been fixed or at least it's not building for me even with `python-flask` install.

lothar-mar:
For me neither. Is there any other way to fix this? Also using arch...

livmackintosh:
You can find a patched PKGBUILD over at https://github.com/livmackintosh/onionshare-pkgbuild . More of a quick fix but will make a prospective PR shortly.

georgeowell:
This pull requests will fix this issue for now:
https://github.com/micahflee/onionshare/pull/465

Tested and working with this PKGBUILD:
https://github.com/livmackintosh/onionshare-pkgbuild/blob/master/PKGBUILD

PR #465, opened 2017-09-18
Fix flask version parsing

livmackintosh:
PR chops the '-dev' suffix from the flask version string dynamically. Probably not safe if Flask start changing their usual version formula but should generally be fine for now.

Made a prospective v1.2.dev1 version tag on my side but you can tag it how you like. Would be great if you can get a release out ASAP so we/someone can fixup the Arch Linux PKGBUILD.
Fixes: #442


simonsmiley:
I wouldn't merge that. Use the Looseversion proposed in #442 instead. This is more of a hack...

livmackintosh:
Sure. This was intended to bridge the gap between the actual releases being completely broken (I.e. not compiling at all) and having a "perfect" solution. Feel free to add some commits...

Issue #454, opened 2017-06-29
Tor 0.3.0.9 fixes a path selection bug that would allow a client to use a guard that was in the same network family as a chosen exit relay


Baccount:
Tor 0.3.0.9 fixes a path selection bug that would allow a client to use a guard that was in the same network family as a chosen exit relay. 
https://blog.torproject.org/blog/tor-0309-released-security-update-clients

Issue #448, opened 2017-06-09
An attacker can remotely crash a hidden service with an assertion failure!


Baccount:
Tor 0.3.0.8 fixes a pair of bugs that would allow an attacker to remotely crash a hidden service with an assertion failure. Anyone running a hidden service should upgrade to this version, or to some other version with fixes for TROVE-2017-004 and TROVE-2017-005.
Please update tor to 0.3.0.8.
https://blog.torproject.org/blog/tor-0308-released-fix-hidden-services-also-are-02429-02514-02612-0278-02814-and-02911

PR #469, opened 2017-11-03
Update Tor bundles for OS X and Windows

mig5:
The install/build_osx.sh script is failing because Tor Project has removed the 6.5.2 release from https://dist.torproject.org/torbrowser/ so it 404s fetching the dmg.

This updates the version/filenames and associated signatures.

(I verified the packages against the GPG signing key, in order to have faith in the SHA256 sums, but I have no implicit trust in the key - I see you have signed Tor's signing key, so please double-check along with the SHA256 sums for safety)

This probably takes care of #454 and #448 for non-Linux users too..

### DIFF
--- a/install/get-tor-osx.py
+++ b/install/get-tor-osx.py
@@ -28,9 +28,9 @@ import inspect, os, sys, hashlib, zipfile, io, shutil, subprocess
 import urllib.request
 
 def main():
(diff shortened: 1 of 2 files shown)


mig5:
Hmm, unsure if related to this, but my Tor Connection Dialog doesn't update the progress bar or show the progress messages.. it works on stable release. It does actually connect OK though, just no visual feedback.. weird.

PR #473, opened 2017-11-08
Implements a shutdown timer to stop a share automatically (downloaded or not) after N hours

mig5:
This is a reimplementation of #296

The UI could perhaps do with some work (maybe human-readable seconds/minutes/hours/date) but I made it a float on purpose since I use decimals for time a lot, and find it convenient to reference (say) 0.08 for 5 minutes just as much as 1.25 for an hour 15 minutes.

The functionality works as expected both via the CLI and the UI. Unlike #296 it does not matter whether the share stays open after first download (it will shut down whether any or many downloads have occurred, once the limit is reached).

I welcome changes but maybe it'd be good to get this in for its functionality and then work on improving the UI especially if you had other UI changes planned (which I think you did).


mig5:
Couple of screenshots for posterity showing the current implementation.

The user is presented with a simple checkbox offering to set a timer. 
When clicked, the actual timer settings are shown. 
When the share is started, the timer is made read-only. 
When the share is stopped, the original checkbox is shown again, reset to unchecked (so that future shares are not forced to use a timer).

screen shot 2017-11-08 at 8 29 44 pm
screen shot 2017-11-08 at 8 30 14 pm
screen shot 2017-11-08 at 8 29 28 pm

micahflee:
Very cool. I'm reviewing it now.

micahflee:
This looks great, but there are a couple of issues. Here's what it looks like before checking the "Set auto-stop timer?" box:

screenshot_2017-12-01_10-48-13

And here's what it looks like after checking it:

screenshot_2017-12-01_10-49-15

If I change my mind and decide I don't want to auto-stop after all, there's no UI for unchecking the box again. As long as the server hasn't started yet, the checkbox should remain so the user can choose to uncheck it for their next share.

Another issue is when the server stops because the auto-stop time gets reached, the GUI ends up in an unusable state, different than if you stop the server by clicking Stop Sharing. The Add and Delete buttons are disabled, and you can't drag new files into the window. All you can do is start the same share again.

screenshot_2017-12-01_10-59-14

Finally, the auto-stop timer should auto-stop only if the download hasn't _started_ by time set. I shared a very large file, and began downloading it before the timer ran out. Then, when the timer ran out, the server stopped even though the download was still in progress, and on the client I ended up with a partial download. (The reason I think it should work this way is: let's say I'm sharing a large file with someone, and I want OnionShare to auto-stop if they haven't started the download in 12 hours. But once they start the download, it might take more than 12 hours to finish it.)

So to recap:

* Be able to uncheck "Set auto-stop timer?" after it's checked
* When server is auto-stopped, allow user to add/delete files in OnionShare GUI again
* Don't auto-stop if a download is in progress -- but once it's complete or canceled, auto-stop then

Do you think you can implement these improvements?

I've checked for a few other edge cases, like starting the share after the auto-stop has already expired, and started the share so that by the time it starts it expires, and got the appropriate errors, so 👍 .

And also thanks so much for this PR, this is excellent!

mig5:
@micahflee thanks for the review and especially for picking up the latter two issues.

The first two are fixed - the third is a bit trickier since the logic of the timeout running out is checked 'outside' of the 'event' loop - I'll have a think on what is to be done there.

mig5:
@micahflee think I have the third issue fixed now, seems to work as expected in my tests but appreciate lots more testing of edge cases etc. Note that in the event  that stay_open = False, the share will still stop once 100% of the bytes have been sent *not* because the timeout runs out but simply because stay_open is False (so original behavior is retained).

Also, re: these conflicts which are just due to other recent feature branches having been merged.. do you want me to fix that in this branch (e.g merge master into this branch)?

mig5:
Hmm, sorry, I just realised that fix is a bit premature - it only fixes it for the GUI app but not for the CLI. I'll try and spend some more time on it soon.

mig5:
@micahflee a better fix now applied - much more elegant (uses the existing web.done variable) and works for both CLI and GUI.

Appreciate lots of testing, including that there's no regressions for stay_open (all works for me, but I would hate to break anything)

micahflee:
Awesome thank you, I'm testing this now. I just found one issue. If you set an auto-stop timer, then start the share and wait for the timer to expire, it now isn't auto-stopping, and instead it says, "Waiting for download to complete before auto-stopping" -- but I never started an active download. So, it looks like the new commits introduced this bug.

screenshot_2017-12-04_15-06-14

micahflee:
However if I download it once (with stay-open on), it does auto-stop when the time expires, as expected.

screenshot_2017-12-04_15-11-07

mig5:
Argh - thanks - checking it now.

micahflee:
I also just confirmed that it otherwise works as expected -- if I start a big download before the timer runs out, and it runs out while downloading, it lets the download finish before stopping.

And also, I confirmed that if stay-open is false, and I set an auto-stop timer, and download it before the timer runs out, the server stops, as expected.

mig5:
The issue is that we now stop if web.done (== True), necessary to avoid stopping mid-download. But web.done is False if no attempt to download has occurred.

Looking to see if we can check the count of downloads and also allow to stop if length of downloads < 1

mig5:
@micahflee phew, maybe all sorted.. we now stop if the web.download_count == 0 (no attempts to download occurred during the time window) or web.done (all downloads are completed)


I've been testing the following scenarios:

1) Check that it times out (no download)
2) Check that it times out after download (stay_open = True)
3) Check that it still stops after first download (stay_open = False)
4) Check that it doesn't time out mid-download of larger file, but when download is done
5) Check that it still times out if the user Canceled the download

Likewise for the CLI version...


One one occasion my UI froze on the share stopping on download (stay_open = False). It seemed to get 'caught' between these two debug log messages:

[Dec 05 2017 11:18:00 AM] OnionShareGui.stop_server
[Dec 05 2017 11:18:00 AM] OnionShare.cleanup

I had to Ctrl+c to see that 'cleanup' function run, which immediately unfroze my UI. I don't think this is related to this feature branch though... maybe some existing bug/race condition. I don't reproduce it reliably.. would be interesting if you come across it yourself.

micahflee:
Hurray! This seems to work great now.

mig5:
\o/ Let me know when you are going to do a release, and I'll write up a Wiki page on the new feature. Particularly I anticipate that it would be good for people to know a share won't stop at the time expected, if a download is still running, so would be good to capture this in docs.

mig5:
Documentation on the auto-stop timer added at https://github.com/micahflee/onionshare/wiki/Using-the-Auto-Stop-Timer

PR #477, opened 2017-11-27
Mentions python2 as the interpreter

kushaldas:
`/usr/bin/env python` is  not anymore the preferred way to mention which Python interpreter we want to use. This fixes the corresponding rpmlint error.

```
onionshare.noarch: E: wrong-script-interpreter /usr/share/nautilus-python/extensions/onionshare-nautilus.py /usr/bin/env python
onionshare.noarch: E: non-executable-script /usr/share/nautilus-python/extensions/onionshare-nautilus.py 644 /usr/bin/env python
```


micahflee:
👍

Issue #447, opened 2017-06-07
 Cancel Tor bootstrap bug


Baccount:
When you cancel the tor bootstrap and click open settings then click cancel, the app will allow you to add file and start sharing. but nothing will happen. Here is a video showing the issue.
https://filenurse.com/download/2b2e90d78a8dc2cf2e49c8361bac5f6d.html

mig5:
@micahflee I am happy to work on this (seems quite clear in the code how to react to both events being canceled) but what do you think should happen instead? 

Should we try and:

1) force reboot the Tor connection, or 
2) just disable the add/remove buttons until the matter is dealt with (user opens settings again), or 
3) quit the app entirely?

PR #481, opened 2017-12-04
#447 show alert and exit if the Settings dialog is canceled yet Tor is still not connected

mig5:
Stopgap fix for #447 . 

I initially had it instead re-attempt the Tor connection dialog, but I found that if *that* dialog was canceled, it would return to the original behavior.

In this 'fix', if the user cancels the Tor connection dialog (or somehow Tor is not connected for some other reason), opens Settings, but cancels without making a change (thus not invoking a reload of settings and re-connection of Tor), an alert will be displayed that the user needs to reopen OnionShare. 

The app then exits, so you can't potentially attempt to start a share without a running Tor connection.

There is probably a better long term fix, so feel free to close this.

### DIFF
--- a/onionshare_gui/settings_dialog.py
+++ b/onionshare_gui/settings_dialog.py
@@ -505,7 +505,11 @@ class SettingsDialog(QtWidgets.QDialog):
         Cancel button clicked.
         """
         common.log('SettingsDialog', 'cancel_clicked')
-        self.close()
+        if not self.onion.connected_to_tor:
+            Alert(strings._('gui_tor_connection_canceled', True), QtWidgets.QMessageBox.Warning)
+            sys.exit()
+        else:
+            self.close()
 
     def help_clicked(self):
         """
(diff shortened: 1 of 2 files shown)


micahflee:
I actually think exiting the program is probably the right way to handle this. I mean, what else can we do? For example, disconnect from the internet and try opening OnionShare. If you just can't connect to Tor in any way, then you can't use OnionShare.

I imagine that the main reason people will hit this is because 1) they don't have internet access, or 2) access to Tor is blocked. If Tor is blocked, at the moment the user needs to configure a bridge in Tor Browser then set OnionShare to use Tor Browser settings -- but that's complicated, and probably doesn't happen as frequently as people just not being online.

Maybe we can change the text to say: `"OnionShare cannot connect to Tor.\n\nMake sure you're connected to the internet, then re-open OnionShare to configure the Tor connection."` Then I think this is good.

mig5:
Wording updated as requested.

PR #482, opened 2017-12-05
Don't uncheck the auto-stop timer when the FileList is updated

mig5:
Sorry @micahflee ,

I identified one UX issue with the shutdown timer checkbox.

Previously, I set the checkbox to be unchecked if the SERVER_STOPPED. This was done in the server_status.update function.

This was done so that when the server stops, the checkbox resets back to the default 'unchecked' mode, so that it isn't necessarily enabled by default for the next share.

However, I just discovered that the FileSelection widget triggers the server_status.update() function when the file list changes.

In other words, if you were to either:

1) set the timer *first* and *then* start adding files, or
2) add a file, set the timer, then add or remove a file

.. the timer checkbox would uncheck itself.

This fixes that, but also means that on subsequent new shares, the timer might retain its 'checked' mode - the user would have to uncheck that box if they don't want a timer on the next share. I think the former issue is more of a UX problem, so probably have to live with this extra step.

Hope that makes sense

### DIFF
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -426,7 +426,6 @@ class OnionShareGui(QtWidgets.QMainWindow):
                     # close on finish?
                     if not web.get_stay_open():
                         self.server_status.stop_server()
-                        self.server_status.shutdown_timeout_reset()
                         self.status_bar.showMessage(strings._('closing_automatically', True))
                 else:
                     if self.server_status.status == self.server_status.STATUS_STOPPED:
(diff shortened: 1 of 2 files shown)


mig5:
Actually, just realised I could still force the checkbox to be unchecked when the GUI fires server_status.stop_server(). By moving the call into the timeout_reset() function I was also able to remove some duplicate code.

So, this is now the best of both worlds:

1) Adding or removing files when the timer is checked, doesn't uncheck the timer
2) When the server stops, the timer resets to its default mode of 'unchecked', just like when the OnionShare app is opened for the first time

Let me know if you think this is the optimal UX.

micahflee:
Yeah I think this is better, thanks

Issue #343, opened 2017-01-09
bug, ADD_ONION tries port 0?


adrelanos:
Just observed that.

Incoming Tor ControlPort command.

> ADD_ONION NEW:BEST Port=80,0

This is predestined to fail.

> 512 Invalid VIRTPORT/TARGET

Is this something caused by my Whonix setup or could you check please if there can be a condition under which the VIRTPORT will be zero?

mig5:
I'm going to close this as I can't see anywhere in the code that it could possibly use port 0 anymore. 

Moreover, there are now additional sanity checks that will throw an alert if somehow a proper random port couldn't be obtained, so it should never get that far again. (https://github.com/micahflee/onionshare/pull/484)

PR #484, opened 2017-12-06
Raise an error if we were unable to obtain a random port

mig5:
If all the available ports are in use, attempting to start the server just hangs forever in SERVER_WORKING state.

This fix wraps the attempt to get_available_port() so that it raises an exception if it can't obtain the port.

I think this might've been what led to #343 because I was able to see 'Configuring onion service on port 0' when the 'except' was just a 'pass'. (though I also see there are additional exceptions now to handle that Tor Connection issue, mentioning 'If you are using Whonix...' so that issue can probably just be closed now anyway.)

A simple PoC to use up the ports before attempting to start the share:

```
#!/bin/bash

for port in {17600..17650}; do
  nc -l $port &
done
```

You should now get an Alert in the UI when you attempt to Start Sharing.

### DIFF
--- a/onionshare/common.py
+++ b/onionshare/common.py
@@ -196,8 +196,8 @@ def get_available_port(min_port, max_port):
             try:
                 tmpsock.bind(("127.0.0.1", random.randint(min_port, max_port)))
                 break
-            except OSError:
-                pass
+            except OSError as e:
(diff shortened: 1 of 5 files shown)


micahflee:
Nice, tested and this works great

PR #489, opened 2017-12-07
Patches from the proposed OpenBSD port

StAlphonsos:
Essentially turn Linux assumptions into more general Unix ones.

Thanks for your consideration, -Attila of torbsd.org

### DIFF
--- a/onionshare/common.py
+++ b/onionshare/common.py
@@ -23,6 +23,7 @@ import inspect
 import os
 import platform
 import random
+import re
 import socket
 import sys
 import tempfile
@@ -58,6 +59,11 @@ def get_platform():
     """
     return platform.system()
 
+def platform_is_unixy():
+    """
+    Returns true if the platform we're running on is Linux or BSD
+    """
+    return True if re.match('^(Linux|.*BSD)$', get_platform()) else False
 
 def get_resource_path(filename):
     """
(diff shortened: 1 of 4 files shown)


onionshare/common.py
```diff
@@ -58,6 +59,11 @@ def get_platform():
     """
     return platform.system()
 
+def platform_is_unixy():
+    """
+    Returns true if the platform we're running on is Linux or BSD
+    """
+    return True if re.match('^(Linux|.*BSD)$', get_platform()) else False
```
mike-burns: Could simplify this a bit:

```python
def platform_is_unixy():
    return re.search('(Linux|BSD)', get_platform())
```

Or you could test whether you're on a POSIX system (the `posix` import is only available on such a system):

```python
def platform_is_unixy():
    try:
        import posix
        return True
    except ImportError:
        return False
```

onionshare_gui/__init__.py
```diff
@@ -36,7 +36,7 @@ class Application(QtWidgets.QApplication):
     """
     def __init__(self):
         system = platform.system()
```
mig5: You could consider removing this variable assignment 'system' since you are replacing its use in the __init__.py

micahflee:
Hey @StAlphonsos, thanks for implementing this. I agree with the comments from @mike-burns and @mig5 though. Do you think you could clean this up slightly to make it more universal?

Specifically, do you think you can merge `common.platform_is_unixy` with `common.get_platform`, so when you run `common.get_platform()` it either returns `Windows`, `Darwin`, or `Unixy`? Then, everywhere in the code there's an if statements that checks which platform is being run, make it use the new `get_platform` function?

StAlphonsos:
Sorry this took me a bit to get back to.  The problem with your approach is that we need to distinguish between Linux and BSD in at least one place, so I'm reworking the patches to read more naturally and also redoing them against the 1.2 release. I'll submit another ticket with the new patches. Thanks for the feedback -Attila

mig5:
Closing since we merged #585

Issue #493, opened 2017-12-09
Onionshare gets stuck when I close TorBrowser


Baccount:
Onionshare gets stuck when I close TorBrowser.
1. Open onionshare and make sure TorBrowser and "start a share in onionshare"
2. Close TorBrowser when the onion link is generated.
3. Click "stop sharing" and then click "start sharing", it will cause a traceback and get stuck at "Please wait" and nothing happens.

```
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/jackimoore/Desktop/untitled folder/onionshare-master/onionshare_gui/onionshare_gui.py", line 250, in start_onion_service
    self.app.start_onion_service()
  File "/Users/"User"/Desktop/untitled folder/onionshare-master/onionshare/onionshare.py", line 79, in start_onion_service
    self.onion_host = self.onion.start_onion_service(self.port)
  File "/Users/"User"/Desktop/untitled folder/onionshare-master/onionshare/onion.py", line 404, in start_onion_service
    res = self.c.create_ephemeral_hidden_service({ 80: port }, await_publication=True)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/stem/control.py", line 2903, in create_ephemeral_hidden_service
    response = self.msg(request)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/stem/control.py", line 627, in msg
    self._socket.send(message)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/stem/socket.py", line 126, in send
    raise stem.SocketClosed()
stem.SocketClosed

```

micahflee:
This is fixed now thanks to #503. I just tested using Tor Browser, and confirmed that if a share is active and you close Tor Browser, it stops the share and alerts the user.

PR #494, opened 2017-12-10
#493 check the Tor connection is still working before attempting to start the server

mig5:
Fix for #493 

New helper function Onion.is_authenticated() (we can't rely on self.onion.connected_to_tor here because it's still set to True from the previously-working connection, even if Tor Browser has been closed).

Previously I just threw a sys.exit() after alerting, but realised we can present the opportunity to the user to fix the situation without closing the app, so added another commit. 

Now, if not connected to Tor when clicking Start Server, we alert and offer the ability to edit the Tor connection settings (open the SettingsDialog). When saving the Settings, check if we're still not connected to Tor, and if not, attempt again to reconnect. The scenario I'm accounting for here is as follows:

1) Start Tor Browser
2) Open OnionShare, configured to rely on Tor Browser rather than bundled
3) Start a share
4) Close Tor Browser
5) Stop share
6) Start a share (rather than hang forever in SERVER_WORKING, you will now receive an alert and the offer to open SettingsDialog)
7) Start Tor Browser
8) Save Settings (we will now try and reconnect to Tor (successfully, because you re-opened Tor Browser))
9) Start a share (this will now succeed)

If you repeat as above but skip step 7, OnionShare will give up, alert the user and exit, since it could still not obtain a connection. (similar to #447)

Other bits of functionality such as UpdateChecker() could probably now also benefit from the Onion.is_authenticated() conditional, and maybe more generally phase out the self.onion.connected_to_tor boolean (or we could re-save that boolean as True in this new function).

### DIFF
--- a/onionshare/onion.py
+++ b/onionshare/onion.py
@@ -375,6 +375,14 @@ class Onion(object):
             # ephemeral stealth onion services are not supported
             self.supports_stealth = False
 
+
+    def is_authenticated(self):
+        """
+        Returns True if the Tor connection is still working, or False otherwise.
+        """
+        return self.c.is_authenticated()
+
+
     def start_onion_service(self, port):
         """
         Start a onion service on port 80, pointing to the given port, and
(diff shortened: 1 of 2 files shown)


mig5:
Also now contains a similar fix for when Tor connection goes AWOL during a *running* share (#500)

mig5:
Sorry, this is premature, there seem to be weird issues with bundled Tor being killed before the share is started. I'll try work on it more sometime soon.

Issue #500, opened 2017-12-20
If Onionshare is set to use TorBrowser to connect and you make start a share and when the share is active in onionshare then you close TorBrowser


Baccount:
If Onionshare is set to use TorBrowser to connect and you make start a share and when the share is active in onionshare then you close TorBrowser, there is no error message in Onionshare even though the onion-servive service is offline.
Steps.
1. Open onionshare and make sure that it is set to connect to TorBrowser.
2. Open Torbrowser and then start a file share in onionshare.
3. While the share is active Close TorBrowser.
4. The onion service will go down but onionshare still shows that the share is still active.

Related to #493

mig5:
I have added a fix for this into the PR at #494 since it's indeed very closely related, maybe you want to test it out @Baccount?

In theory this works for other types of 'Tor got disconnected' e.g not just Tor Browser method but bundled or sockets. It also partially solves #370 e.g if the internet connection is lost, the server will alert and stop. (Fun side fact: if combined with the private-key-saving feature in #485, you would in theory be able to start the server back up again later under the same URL when you fix your Tor/Internet connection)

Baccount:
A error message is correctly displayed when when closing TorBrowser, Thank you mig5 for the fast and Excellent patch.

mig5:
Hmm I think I have found more corner cases with the 'bundled' Tor, I might have to close the PR and work on it some more :( sorry

mig5:
A better PR hopefully https://github.com/micahflee/onionshare/pull/503

micahflee:
Looks like #503 fixes both this and #493.

PR #503, opened 2017-12-20
#493, #500 - detect if the Tor connection (bundled, TorBrowser or otherwise) has been lost while the app is open. Stop a running share if so

mig5:
A re-implementation of https://github.com/micahflee/onionshare/pull/494 which is hopefully accounting for all weird edge cases and the issues #493 and #500.

@Baccount I would love some testing of this (no pressure though) with both Tor Browser and 'Bundled Tor' modes. You can kill Bundled Tor by running 'pkill -f tor.real' from a terminal while the app is open, and you should get the same alert as when you close Tor Browser when using that mode.

### DIFF
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -176,6 +176,7 @@ class OnionShareGui(QtWidgets.QMainWindow):
         quit, or open settings.
         """
         common.log('OnionShareGui', '_tor_connection_canceled')
+        self.timer.stop()
 
         def ask():
             a = Alert(strings._('gui_tor_connection_ask', True), QtWidgets.QMessageBox.Question, buttons=QtWidgets.QMessageBox.NoButton, autostart=False)
@@ -394,11 +395,14 @@ class OnionShareGui(QtWidgets.QMainWindow):
         """
         self.update()
 
-        # Has the Tor Connection dialog finished, but we've since lost connection to Tor somehow?
-        if not self.onion.is_authenticated() and self.tor_con.t.isFinished():
-            self.timer.stop()
-            self.start_server_error(strings._('error_tor_protocol_error'))
-            self._tor_connection_canceled()
+        # Have we lost connection to Tor somehow?
+        if self.tor_con.t.isFinished():
+            if not self.onion.is_authenticated():
+                self.timer.stop()
+                if self.server_status.status != self.server_status.STATUS_STOPPED:
+                    self.server_status.stop_server()
+                    self.status_bar.clearMessage()
+                self._tor_connection_canceled()
 
         # scroll to the bottom of the dl progress bar log pane
         # if a new download has been added
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -176,7 +176,6 @@ class OnionShareGui(QtWidgets.QMainWindow):
         quit, or open settings.
         """
         common.log('OnionShareGui', '_tor_connection_canceled')
-        self.timer.stop()
 
         def ask():
             a = Alert(strings._('gui_tor_connection_ask', True), QtWidgets.QMessageBox.Question, buttons=QtWidgets.QMessageBox.NoButton, autostart=False)
@@ -223,8 +222,10 @@ class OnionShareGui(QtWidgets.QMainWindow):
             # We might've stopped the main requests timer if a Tor connection failed.
             # If we've reloaded settings, we probably succeeded in obtaining a new
             # connection. If so, restart the timer.
-            if not self.timer.isActive():
-                self.timer.start()
+            if self.onion.is_authenticated():
+                if not self.timer.isActive():
+                    self.timer.start()
+                    self.status_bar.clearMessage()
 
         d = SettingsDialog(self.onion, self.qtapp, self.config)
         d.settings_saved.connect(reload_settings)
@@ -401,8 +402,7 @@ class OnionShareGui(QtWidgets.QMainWindow):
                 self.timer.stop()
                 if self.server_status.status != self.server_status.STATUS_STOPPED:
                     self.server_status.stop_server()
-                    self.status_bar.clearMessage()
-                self._tor_connection_canceled()
+                self.status_bar.showMessage(strings._('gui_tor_connection_lost', True))
 
         # scroll to the bottom of the dl progress bar log pane
         # if a new download has been added
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -142,10 +142,10 @@ class OnionShareGui(QtWidgets.QMainWindow):
         self.set_server_active(False)
 
         # Start the "Connecting to Tor" dialog, which calls onion.connect()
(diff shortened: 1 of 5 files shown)


mig5:
Nope sorry, still edge cases here - canceling the Tor connection dialog and then clicking Save in the SettingsDialog, is not rebooting the Onion.

Sorry for the distraction, I'm clearly in over my head :(

mig5:
More fixes for these funky-ass edge cases and UX experience (too many alerts under certain conditions)... might be third time lucky??? Testing welcome..

onionshare_gui/tor_connection_dialog.py
```diff
@@ -86,6 +86,7 @@ def _connected_to_tor(self):
     def _canceled_connecting_to_tor(self):
         common.log('TorConnectionDialog', '_canceled_connecting_to_tor')
         self.active = False
+        self.onion.cleanup()
```
mig5: Sort of a mini-bug fixed here. When the Tor Connection dialog was canceled, the actual Onion process continued to run in the background. In theory this could lead to multiple background tor connections if the user later clicked Save under the SettingsDialog.. or at worst, it meant that Onion.is_authenticated() would evaluate to True instead of False, which affected the fixes being implemented in this PR since they heavily interrogate that function.

Baccount: If you open Onionshare with use Tor bundled with onionshare enabled.
Make sure Torbrowser is closed and enable "Attempt to automatically configure with Tor Browser" and click save, your will get a extra "Would you like to open OnionShare settings to troubleshoot connecting to Tor?" popup even though settings is all ready open. Clicking the exit settings button will open 2 settings dialogs.

mig5: Cool, so I officially give up on this one, it's just too complex for my skill. Thanks for testing.

Gonna close this but leave my branch so someone else can maybe see how I tried to fix it.

mig5:
I couldn't resist - pushed one more attempt at this. This one is less heavy handed: no Alert dialogs (which stops the chance of ending up with multiple Settings Dialogs or an array of error messages). It just displays a simple Disconnected message in the status bar - it's up to the user to open settings to reconfigure, or re-open the app.

It will still stop an active share if the Tor connection dies.

Baccount:
When I add 5c84c27 to the current master, I get a traceback when I start the app's GUI.


```
  File "./dev_scripts/onionshare-gui", line 28, in <module>
    onionshare_gui.main()
  File "/Users/jackimoore/Desktop/onionshare-master/onionshare_gui/__init__.py", line 56, in main
    strings.load_strings(common)
  File "/Users/jackimoore/Desktop/onionshare-master/onionshare/strings.py", line 44, in load_strings
    translations[lang] = json.load(f)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/json/__init__.py", line 268, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/json/__init__.py", line 319, in loads
    return _default_decoder.decode(s)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 7 column 1 (char 6)
```

mig5:
You need more than just that commit, you need the whole branch - you are better off cloning my fork and running ```git checkout 493_500_detect_broken_tor_connection``` and then building as usual from that.

Baccount:
After I have onionshare and Torbrowser open and I have onionshare set to connect to TorBrowser, Then I close tor browser and I get the "Disconnected from Tor" in the status bar. And I click start sharing and it gets stuck in a infinite loop and never gives a error.

mig5:
Thanks, what I will do is simply disable the 'Start Sharing' button so you can't somehow try to start the share until you resolve the connection issue.

mig5:
Pushed a fix for that. 

Note that if the user has started Tor Browser again, the user still has to go to Settings and click Save (or perhaps they've changed Tor connection method via Settings dialog, either way), before the 'Start Sharing' button becomes clickable again. Or, they can close OnionShare and re-open.

It would've been nice to automatically reconnect somehow if Tor Browser has started back up, but that's super hard to do without attempting connection, which will then cause another avalanche of Alert/Settings dialogs. I think this is the cleanest way (or maybe even just one Alert and then force the app to quit entirely after that Alert closes). But at least in the current way, the user gets the opportunity to 'fix' the issue via Settings dialog themselves.

Baccount:
When the disconnected from tor status message is shown, it may be helpful if a system notification is shown.
Just add the code right below line 405, in onionshare_gui.py.
```
if self.systemTray.supportsMessages() and self.settings.get('systray_notifications'):
                 self.systemTray.showMessage(strings._('gui_tor_connection_lost', True), strings._('gui_tor_fix_connection_message', True))
```
And adding a new string to recommend the user to open settings and fix the connection problem.
share/locale/en.json
`"gui_tor_fix_connection_message": "Open Settings and fix the connection problem by changing the way onionshare connects to Tor",`
Improvements on the wording would be helpful.

micahflee:
I'm starting to try to review this one, in case you think it's finished, but when I test it the Tor connection dialog window is totally blank.

screenshot_2018-01-13_16-36-36

mig5:
Weird! I saw blank alert dialogs like this in other work, but only on Debian Jessie, never Fedora. I wondered if it was a PyQt bug.

This PR i felt was otherwise complete.

micahflee:
I'll try to troubleshoot this more on my end and see what's causing the blankness.

onionshare_gui/onionshare_gui.py
```diff
@@ -389,6 +400,17 @@ def check_for_requests(self):
         """
         self.update()
 
+        # Have we lost connection to Tor somehow?
+        if self.tor_con.t.isFinished():
```
micahflee: I'm not sure, but think this might be the issue.

`self.tor_con` is an instance of `TorConnectionDialog`, which is a dialog window that's supposed to open, connect to Tor using the passed in `onion` object and display output, and then close. Before it wasn't a class variable, and the reference to it died as soon as the `OnionShareGui` constructor finished executing. I think, in Linux (but maybe not in Mac for some reason), as soon as you reference `self.tor_con` here, PyQt5 is for some reason showing the window again after it's been hidden, only all the widgets in it are still hidden.

What are you trying to do here? I think `self.tor_con.t` isn't the right object you should be checking, because it's just an old thread that was used to connect to Tor, and doesn't need to exist anymore. I think everything you need should be in `self.onion`, like maybe `self.onion.c` for the stem Controller, or `self.onion.tor_proc` for the actual tor subprocess.

micahflee:
Ok, I think I fixed it! I also tested the rest of this PR, and it seems to work great.

Issue #508, opened 2017-12-25
I have had the error "Connecting to Tor is taking too long." when using the bundled Tor 


Baccount:
I have had the error "Connecting to Tor is taking too long." multiple times when using the bundled Tor.
I opened a bug report on [Trac ](https://trac.torproject.org/projects/tor/ticket/24511#comment:16) that is related to the long bootstrap times and teor said.

["This is normal behaviour. Our target bootstrap time is under 30 seconds, but the network is a bit overloaded right now."](https://trac.torproject.org/projects/tor/ticket/24511#comment:14)

Increasing the timeout should fix the issue. 
https://github.com/micahflee/onionshare/commit/022bc57b353de4956c1e7db352955cb00f702ed2

leonardoporpora:
I have got the same problem.

Baccount:
#leonardoporpora  In the meantime change Onionshare setting `Attempt to automatically configure with TorBrowser` and make sure Torbrowser is open when onionshare is open.

Baccount:
https://github.com/micahflee/onionshare/pull/517/commits/c324cd55da7d40008039462f2a77dfb93776bb0b

PR #517, opened 2017-12-27
Increasing the timeout

Baccount:
I have had times where the bootstrap has taken around 80 seconds so I set it to 90 seconds just to be safe. Please correct the time if you thing that timeout is too long. Thank you @mig5 , lets see if this works. #508


Baccount:
https://github.com/micahflee/onionshare/issues/508

PR #496, opened 2017-12-11
Add bridge support (obfs4 or user-supplied)

mig5:
I took a stab at #435

It works, but I need some help with the widgets in the SettingsDialog. When choosing 'custom' bridges, the textbox doesn't expand the parent widgets, so is hard to read. (If you are able to paste in some contents and hit Save, then next time you open the SettingsDialog, it is much better because the settings were detected and so displayed by default).

Additionally it's possible to mess up the other Tor connection widgets above (control port, socket file) when they show the hidden extra options. I couldn't work out how to make them all place nicely :( but otherwise the feature should work.

I think it's the Buttons HBox that refuses to budge when child widgets in the VBoxes above it, show or hide hidden widgets....

I copied the obfs4 transports that seem to come hardcoded into Tor Browser.

Maybe you can fork what I've done and fix the widget issues if it's something trivial that I'm missing.

PR #521, opened 2017-12-27
#520 trigger browser action properly in QLabel hyperlink

### LINKED ISSUE #520: Click the link "More information" in settings does not do anything
Click the link "More information" in settings does not do anything.
https://github.com/micahflee/onionshare/blob/5c390f1bd99a756682801375cc8c0a1418060f65/onionshare_gui/settings_dialog.py#L81
bug
Tested on macOS 10.13.2 (17C88)

mig5:
Fixes #520 

Taken from https://stackoverflow.com/questions/8427446/making-qlabel-behave-like-a-hyperlink

At this time, the only other widget that contains a hyperlink in strings() is 'update_available' - that's an Alert dialog rather than a Qlabel, and there is of course no update available, so it's hard to test if that dialog needs a similar fix or is immune (e.g whether it's QLabel specific). 

Note that in my 'support bridges' PR there is a similar QLabel with a hyperlink, so will need a similar fix if merged..

### DIFF
--- a/onionshare_gui/settings_dialog.py
+++ b/onionshare_gui/settings_dialog.py
@@ -72,6 +72,8 @@ class SettingsDialog(QtWidgets.QDialog):
         # Stealth
         stealth_details = QtWidgets.QLabel(strings._("gui_settings_stealth_option_details", True))
         stealth_details.setWordWrap(True)
+        stealth_details.setTextInteractionFlags(QtCore.Qt.TextBrowserInteraction)
+        stealth_details.setOpenExternalLinks(True)
         self.stealth_checkbox = QtWidgets.QCheckBox()
         self.stealth_checkbox.setCheckState(QtCore.Qt.Unchecked)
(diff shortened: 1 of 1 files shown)


mig5:
@Baccount testing welcomed

Baccount:
It works on macOS 10.13.2 (17C88).

micahflee:
A good way to test the update available dialog is to edit `share/version.txt` and set it to a value smaller than the current stable version. I'm on Linux right now, which doesn't support checking for updates (because I hope Linux users are using a repository), but I can check for OSX or Windows soon.

Also @mig5, can you make sure #496 contains a similar fix for that QLabel before I merge it?

mig5:
@micahflee just noting that I modified the version.txt as mentioned, and could confirm that on macOS, the 'Click here' link for the new OnionShare version, opened a browser for me. So somehow, when a link is in an Alert Dialog rather than a QLabel, it works as expected.

PR #522, opened 2017-12-27
Fixed issues 506, 453

NulAsh:
I made two changes:
1.) If config file is not specified in command line, onionshare will first check for it in folder where onionshare is located. So it's easy to make it portable - just copy config file to that directory.
2.) Now you can choose, where temporary zip file will be created.


xp499:
Hello @micahflee . I see this necessary in the case you want to share large filesizes. Other day sharing 32 GBs got my /tmp dir fullfilled and my computer almost exploded (lol, jk).

Also in those cases i don't see why those files cannot be shared separately or creating the zip in runtime as of MEGA does/offer.

Greetings.

PR #524, opened 2018-01-01
Check the file is a valid file or dir (not just that it exists)

mig5:
This is super corner-case stuff, but I discovered both the CLI and GUI (via --filenames parameter) allowed the user to try to share, say, /dev/null, /dev/urandom, or a socket file somewhere, all of which will result in invalid empty zip packages

The FileSelection widget already ignores such weird files so it's not really a big deal assuming most people just use the GUI

So this change swaps out os.path.exists for the more stringent os.path.isfile / isdir

There remains a corner-case where you can share /dev (a weird dir which equates to 3080 bytes in Python on Linux, but 68 bytes on OS X, and probably similar issues on Windows) so given the inconsistency, I left out a check for that now... previously I was going to reject a dir that is less than 4096 bytes (which at least on Linux is the minimum size even for an empty dir)


mig5:
Actually, here's another corner-case: trying to share /tmp fails on the socket file that is incidentally created by Onionshare for the Tor connection

```
Configuring onion service on port 17640.
[Jan 02 2018 09:56:08 AM] OnionShareGui.start_server_step2
 * Running on http://127.0.0.1:17640/ (Press CTRL+C to quit)
Exception in thread Thread-4:
  File "/usr/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/user/onionshare/onionshare_gui/onionshare_gui.py", line 290, in finish_starting_server
    web.set_file_info(self.file_selection.file_list.filenames, processed_size_callback=_set_processed_size)
  File "/home/user/onionshare/onionshare/web.py", line 101, in set_file_info
    z.add_dir(info['filename'])
  File "/home/user/onionshare/onionshare/common.py", line 255, in add_dir
    self.z.write(full_filename, arc_filename, zipfile.ZIP_DEFLATED)
  File "/usr/lib64/python3.6/zipfile.py", line 1622, in write
    with open(filename, "rb") as src, self.open(zinfo, 'w') as dest:
OSError: [Errno 6] No such device or address: '/tmp/tmplr7b99nx/control_socket'

```

This causes another one of those IPC 'crashes' where the GUI gets stuck on Crunching files after this trace.

Maybe it is necessary to os.walk into dirs that contain invalid files as well, and refuse to share the parent dir - what do you think?

micahflee:
Hmm, so I'm looking at this feature branch, and I merged in master (which includes commits from #497), but this is what happens when I try sharing /tmp:

```
user@dev ~/code/onionshare (mig5-check_is_valid_file) $ ./dev_scripts/onionshare --debug /tmp/
[Jan 13 2018 19:02:00] Onion.__init__
[Jan 13 2018 19:02:00] Onion.connect
[Jan 13 2018 19:02:00] Settings.__init__
[Jan 13 2018 19:02:00] Settings.load
[Jan 13 2018 19:02:00] Settings.load: Trying to load /home/user/.config/onionshare/onionshare.json
[Jan 13 2018 19:02:12] OnionShare.__init__
[Jan 13 2018 19:02:12] OnionShare.set_stealth: stealth=False
[Jan 13 2018 19:02:12] OnionShare.start_onion_service
[Jan 13 2018 19:02:12] Onion.start_onion_service
Configuring onion service on port 17609.
Preparing files to share.
  File "./dev_scripts/onionshare", line 28, in <module>
    onionshare.main()
  File "/home/user/code/onionshare/onionshare/__init__.py", line 101, in main
    web.set_file_info(filenames)
  File "/home/user/code/onionshare/onionshare/web.py", line 101, in set_file_info
    z.add_dir(info['filename'])
  File "/home/user/code/onionshare/onionshare/common.py", line 255, in add_dir
    self.z.write(full_filename, arc_filename, zipfile.ZIP_DEFLATED)
  File "/usr/lib64/python3.6/zipfile.py", line 1622, in write
    with open(filename, "rb") as src, self.open(zinfo, 'w') as dest:
OSError: [Errno 6] No such device or address: '/tmp/atom-1.21.0-user.sock'
```

It looks like error checking in #497 happens in a different location. I agree that it's very minor, but maybe it would be good to check for OSError where this error gets thrown as well, and then I can merge this.

mig5:
I've pushed a fix that more 'neatly' errors out for the CLI on that OSError now.

```
Miguels-MacBook-Pro:onionshare miguel$ ./dev_scripts/onionshare --debug /tmp
[Jan 14 2018 20:31:09] Onion.__init__
[Jan 14 2018 20:31:09] Onion.connect
[Jan 14 2018 20:31:09] Settings.__init__
[Jan 14 2018 20:31:09] Settings.load
[Jan 14 2018 20:31:09] Settings.load: Trying to load /Users/miguel/Library/Application Support/OnionShare/onionshare.json
[Jan 14 2018 20:31:09] OnionShare.__init__
[Jan 14 2018 20:31:09] OnionShare.set_stealth: stealth=False
[Jan 14 2018 20:31:09] OnionShare.start_onion_service
[Jan 14 2018 20:31:09] Onion.start_onion_service
Configuring onion service on port 17605.
Preparing files to share.
Operation not supported on socket
Miguels-MacBook-Pro:onionshare miguel$
```

For the GUI mode, with the merging of #497, as I expected, it already now throws an Alert dialog and stops the share, instead of the IPC getting 'disrupted' from the backend issue, which was the main issue I hoped was addressed.

micahflee:
Nice, it seems to work now for me

PR #526, opened 2018-01-02
#525 catch error in start_onion_service in case we somehow failed to parse a service_id in Stem controller response

### LINKED ISSUE #525: TypeError on start_server_finished
Somehow I triggered this after clicking Start Sharing - right at the moment where you'd expect the share to become ready after the usual momentary delay e.g the Onion Service finishing its negotiation:

```
Configuring onion service on port 17606.
  File "/home/user/onionshare/onionshare_gui/server_status.py", line 214, in start_server_finished
    self.copy_url()
  File "/home/user/onionshare/onionshare_gui/server_status.py", line 237, in copy_url
    url = 'http://{0:s}/{1:s}'.format(self.app.onion_host, self.web.slug)
TypeError: unsupported format string passed to NoneType.__format__
Aborted (core dumped)
```

I haven't been able to reproduce. Obviously my Tor connection was working in order to even start the share, but it seems like somehow the Onion Service failed to obtain somehow.

I don't know how to reliably reproduce, but wondering if we just need to wrap self.app.onion_host and self.web.slug in some conditionals there to make sure they are valid and if not, fail the share gracefully?

An unrelated note, it seems that start_server_finished is emitted in start_server_step2 but should maybe be done in start_server_step3, don't think that was the cause though.

mig5:
Fixes #525

### DIFF
--- a/onionshare/onion.py
+++ b/onionshare/onion.py
@@ -413,7 +413,10 @@ class Onion(object):
             auth_cookie = res.content()[2][2].split('=')[1].split(':')[1]
             self.auth_string = 'HidServAuth {} {}'.format(onion_host, auth_cookie)
 
-        return onion_host
+        if onion_host is not None:
+            return onion_host
+        else:
+            raise TorErrorProtocolError(strings._('error_tor_protocol_error'))
 
     def cleanup(self):
         """


micahflee:
It looks like CI test failed because of a networking issue, but tests pass locally, so I think this looks good.

Issue #528, opened 2018-01-02
When I have "use Tor bundled with onionshare" enabled, clicking on check for updates gives me this error "Error checking for updates"


Baccount:
When I have "use Tor bundled with onionshare" enabled, clicking on check for updates gives me this error  "Error checking for updates: Maybe you're not connected to Tor, or maybe the OnionShare website is down."

mig5:
I reproduced only once, all other times it works for me. (It checks over the Onion Service, maybe it's just more Tor congestion nothing we can do about?)

The time that it failed for me, this was in the log

```
UpdateChecker.chec: loading http://elx57ue5uyfplgva.onion/latest-version.txt
Onion.get_tor_socks_port
UpdateCHecker.check: Socket error: timed out
```

mig5:
The timeout is only set to 15 seconds, maybe it needs to be a bit longer during periods of congestion

Baccount:
I think the issue is when bundle tor is selected the updates does not have any connection,  there is no delay from when I click check for updates and when I get the error message.

mig5:
ok odd, I am using bundled Tor on OS X and I don't reproduce that at least against a built version of master branch

mig5:
@Baccount I reproduce if I first got the 'Tor is taking too long to connect' message with Bundled Tor, and then am taken to the Settings dialog. Of course in that scenario the 'Check for updates' fails immediately since it relies on Bundled Tor to have connected.

So do you mean that your bundled Tor never connected and then you went straight to Settings Dialog and clicked 'Check for Updates'?

If so maybe we should disable the pushbutton until Tor is connected somehow..

Baccount:
Disabling the check for updates button should work.

PR #529, opened 2018-01-03
#528 disable the 'Check for Updates' button in the SettingsDialog if Tor is not connected

mig5:
No major harm, but no sense in allowing the user to click the Updates button if their Tor connection failed to succeed (which might be why they've landed on the SettingsDialog in the first place)

Probably is the cause of #528

### DIFF
--- a/onionshare_gui/settings_dialog.py
+++ b/onionshare_gui/settings_dialog.py
@@ -96,6 +96,9 @@ class SettingsDialog(QtWidgets.QDialog):
         # Check for updates button
         self.check_for_updates_button = QtWidgets.QPushButton(strings._('gui_settings_autoupdate_check_button', True))
         self.check_for_updates_button.clicked.connect(self.check_for_updates)
+        # We can't check for updates if not connected to Tor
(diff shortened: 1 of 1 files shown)


mig5:
Damn this flaky Travis!

```
The command "curl -s -o python-nightly.tar.bz2 ${archive_url}" failed and exited with 35 during .
```

Baccount:
After I added your patch the bug still seams to occur. Here is a screen recording shoeing the bug occur.
[bug.mov.zip](https://github.com/micahflee/onionshare/files/1599452/bug.mov.zip)

mig5:
So that's slightly different - the 'Test Tor Connection' does not reactivate the 'main' bundled Tor connection, it's a more standalone connection for the purposes of the test.

So that's why it still doesn't work when you click on Check for Updates.

The real issue is that it should not re-enable the Check for Updates button at the completion of the test. The button should remain disabled until the main Tor connection is re-activated (e.g on Save). I'll push a fix for that corner case soon.

Issue #543, opened 2018-01-17
Persistent URL bug (Windows specific?)


mig5:
There is some sort of bug in the persistent URL feature :(

Tested on Windows

I opened OnionShare, configured a bridge in settings (I am on the 435_support_bridges branch), and set the Persistent mode on.

I started a share, stopped it, started again, and the .onion URL was different (but the slug was the same)

Frustratingly I don't reproduce from master branch, but all the persistent URL stuff I think is merged into the bridge branch already due to resolving conflicts

It implies the backend Onion object didn't save the .onion URL back to settings json. Not sure if Windows-specific.. 

Checking the JSON it is seeming to set the 'save_private_key' back to false even though I verified it was true beforehand :/ and the 'slug' is an empty string in the json, yet it is reusing the slug! I definitely don't reproduce this in Linux.

Signing off for the night here, I wonder if you can reproduce this.

I hate you Windows!

mig5:
It could be the way we load Settings in https://github.com/micahflee/onionshare/blob/master/onionshare/__init__.py#L81 and then run settings.save() later in the same file

I could see that that Settings object might use 'default' settings where save_private_key is false

But it isn't an issue on Linux :/

mig5:
I tried disabling the auto-update check in case it's that, and now I get consistent results. So maybe indeed that's all the issue is (which I reproduced on Mac for the first time yesterday)

PR #544, opened 2018-01-17
Don't save settings object from onionshare/__init__.py if running on Windows

mig5:
So I wonder something

1) Do you reproduce #543 (persistent .onion is not persistent on Windows, but the slug is, so long as the app stays open)

2) Does this patch fix it for you too, on Windows, as it seems to for me

I am working on the theory that it's something to do with the lack of the command-line mode in Windows, and that when it tried to save the web.slug back to settings via the Settings object, that object's 'config' parameter is False, so it used the 'default' settings, and saves ```'use_private_key': false``` back to the settings file, overwriting what was previously ```'save_private_key': true``` in the json after you save the SettingsDialog but before you start a share.

And that somehow this is not an issue on Linux (haven't tested MacOS), presumably because somehow the 'config' object when passed to OnionShare is not null, or avoids it in some way I don't fully understand.

With this patch, I get the expected big private key and slug saved to the json in Windows, but without, it it is as if it 'resets' back to false/empty settings as soon as the share starts.

Sorry about these surprise bugs in my features on the eve of a release :( ugh

### DIFF
--- a/onionshare/__init__.py
+++ b/onionshare/__init__.py
@@ -129,10 +129,11 @@ def main(cwd=None):
             app.shutdown_timer.start()
 
         # Save the web slug if we are using a persistent private key
-        if settings.get('save_private_key'):
-            if not settings.get('slug'):
-                settings.set('slug', web.slug)
-                settings.save()
+        if common.get_platform() != 'Windows':
+            if settings.get('save_private_key'):
+                if not settings.get('slug'):
+                    settings.set('slug', web.slug)
+                    settings.save()
 
         if(stealth):
             print(strings._("give_this_url_stealth"))


mig5:
Argh - no, I think this is even worse - or yet another related bug - on macOS, the 'check for updates' is maybe saving another new Settings object to json, overwriting my 'save_private_key' after I set it to true.

Man!

mig5:
Assuming that the fix in this PR is actually bogus and it's just all about the updatechecker thread 'overwriting' settings when it saves its timestamp... (I haven't tested to see if that is what was actually happening in Windows, but quite possibly is)

... maybe better to not save a timestamp and just rely on a QTimer(86400) or something?

mig5:
I'm going to close this, because

1) the fix is incorrect
2) there are multiple issues that are not related

I'll create separate PRs as I think I am getting closer to understanding what's going on here.

Sorry for the noise as I go on my journey!

PR #547, opened 2018-01-17
load the Settings in onionshare/__init__.py later,

mig5:
...so that when saving slug, the private_key/hidservauth string have already been saved by the Onion object

*This* was the cause of the slug being saved to Settings json, but not the private key - something I observed on Windows, and then later when using the 'CLI' mode on Linux, in a scenario when save_private_key = True, but there hasn't yet been a share (so no existing private key to rely on, but it should save it back on the next share)

### DIFF
--- a/onionshare/__init__.py
+++ b/onionshare/__init__.py
@@ -79,7 +79,6 @@ def main(cwd=None):
 
 
     settings = Settings(config)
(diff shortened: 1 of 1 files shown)


micahflee:
Nice, good catch

Issue #550, opened 2018-01-18
Kill tor and its subprocesses


micahflee:
When using Bundled Tor, obfs4proxy, if you switch to using no bridges, it crashes like this:

```
[Jan 17 2018 18:32:02] SettingsDialog.save_clicked: Connected to Tor
[Jan 17 2018 18:32:02] SettingsDialog.save_clicked: rebooting the Onion
[Jan 17 2018 18:32:02] Onion.cleanup
Traceback (most recent call last):
  File "onionshare_gui\settings_dialog.py", line 628, in save_clicked
    self.onion.cleanup()
  File "onionshare\onion.py", line 511, in cleanup
    self.tor_data_directory.cleanup()
  File "tempfile.py", line 811, in cleanup
  File "shutil.py", line 494, in rmtree
  File "shutil.py", line 389, in _rmtree_unsafe
  File "shutil.py", line 387, in _rmtree_unsafe
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\MICAHL~1\\AppData\\Local\\Temp\\tmpv011cdr9\\lock'
```

This is because it can't delete the temporary tor data dir, and I'm pretty sure it's because `obfs4proxy.exe` is running and has an open file in that dir, and it doesn't get killed when `tor.exe` gets killed.

So I think the solution is to kill tor's children before killing the tor process (brutal, I know).

micahflee:
Oh interesting. I was just trying to kill subprocesses like this https://stackoverflow.com/questions/1230669/subprocess-deleting-child-processes-in-windows but failed.

So for the moment I'm going with a simpler approach, and just catching the `PermissionError` exception and then ignoring it, because the temp dir will eventually empty itself, and it shouldn't contain any secrets like onion keys or anything in it.

I figure I can spend more time troubleshooting this one in a later release.

mig5:
:+1: I still can't even get obfs4 to get past 20% on Windows, frustrating because the obfs4 in Tor Browser on the same VM Just Works somehow.

micahflee:
My obfs4proxy is stuck at 76% in Windows right now. I picked 6 random bridges from https://bridges.torproject.org/. Haven't succeeded in connecting yet...

maqp:
Another situation where Tor subprocess stays open is when OnionShare CLI client on Linux is closed by closing terminal by clicking the X button. This leaves orphaned child process running in the background. The only solution currently is to remember to close the application every time with ^C first.

[...]

Issue #367, opened 2017-03-04
[Feature request] No-install version (Windows)


vatterspun:
This is connected to case #366 , but there's an easy workaround: just decompress the program with 7zip and launch _onionshare-gui.exe_.

PhoenixCODE2322:
any update on this ?

Xant1k:
fresh news about it? we need to have a normal practice usage

PR #551, opened 2018-01-18
Try to use a persistent directory for holding tor cache/state data

mig5:
Dramatically speeds up connection speed. Falls back to tmp dir if we can't

Only tested in Linux so far


mig5:
Hmm, causes errors connecting in macOS :( even though the various bits and pieces get created in the config directory.

Don't wait for this for 1.2 then..

mig5:
This now uses the old Temporary directory for storing the ControlSocket, which for some reason can't exist in a directory containing whitespace (it chops off at the first whitespace and creates the socket there - very likely a Tor bug!)

Everything else including the Cookie file is in the 'persistent' directory. Working on Mac and Linux, yet to try Windows but will soon.

Baccount:
Much faster than previous versions. Tested on macOS 10.13.2 (17C205)

mig5:
It's working for me now on Windows too.

I continued to have issues switching to obfs4 bridge mode from 'no bridges' mode and had to manually kill processes still.

But after I re-opened OnionShare (having already set obfs4 mode in the last attempt) it connected quite quickly.

I have also written up some docs at https://github.com/micahflee/onionshare/wiki/Using-Bridges

mig5:
I have just added a commit that adds a 'Clean up on exit?' option to the Settings Dialog and will remove this state dir if checked. Tested on Mac and Linux

micahflee:
I'm testing this now too. And wow, I concur, connecting to Tor is _much_ faster. This is excellent. I'm still testing it in different operating systems.

As far as the clear cache on exit option, I think there is value in a "portable" version of OnionShare (#366, #367), but if we do this I think it should probably be more thought out and holistic, instead of just adding portable pieces here and there. Even with clearing the Tor cache on exit, the `settings.json` file will still be there, and maybe other files, which still leaves evidence that the user used OnionShare on the computer. A portable version maybe should still maintain the Tor cache over time for faster connections, but the cache directory and the `settings.json` would be in subdirectory of where the binary is, for example. So, I like the idea, but I think maybe we should discuss all the ways to implement it in #367 and make this a full feature of a later release.

micahflee:
Hmm.. so I'm running into a few problems with this while testing in Windows. Just like happened with you @mig5, I'm finding orphan `tor.exe` and `obfs4proxy.exe` processes that appear to be breaking switching between different Tor bridge settings. I think this is essentially related to #550.

But there's also another problem. Previously in Windows and Linux, you can open multiple instances of OnionShare, and each instance just had its own separate tor process and data dir. Now, if you open a second instance, it will crash and die because the first instance is using that data dir already. Maybe this is fine and you should only be allowed to run one instance at a time (or somehow multiple instances should share the same tor?), but if this is the case then we need make OnionShare check for other running instances when opening and handle it cleanly.

So while I'd really like to merge this as part of v1.2, I think a little more work is required to fix all the issues. So alas, I'm changing the milestone to 1.3 and I'll release 1.2 without this. But hopefully we can resolve it soon, and maybe make the 1.3 release not too long in the future.

Baccount:
@mig5 this will also add protection against timing correlation attacks. https://www.torproject.org/docs/faq#EntryGuards

onionshare/settings.py
```diff
@@ -81,6 +82,19 @@ def fill_in_defaults(self):
...
+        Returns the path of the directory that holds Tor state/cache data.
+        """
+        p = platform.system()
+        if p == 'Windows':
+            appdata = os.environ['APPDATA']
+            return '{}\\OnionShare\\tor'.format(appdata)
+        elif p == 'Darwin':
+            return os.path.expanduser('~/Library/Application Support/OnionShare/tor')
+        else:
+            return os.path.expanduser('~/.config/onionshare/tor')
```
x80486: Is it possible to use the `XDG_CONFIG_HOME` instead?

Following the **XDG Base Directory Specification** these days should be a must for any Linux application.

mig5:
I'm going to close this ancient PR because persistent tor data has formed part of the recent tabs work.

Hooray, OnionShare will finally benefit from the [research-proven anonymity improvement](https://blog.torproject.org/blog/improving-tors-anonymity-changing-guard-parameters) that is persistent entry guards \o/

PR #557, opened 2018-01-19
Support for obfs4 in custom bridges

### LINKED ISSUE #556: Add support for Obfs4 custom bridges
Users should be able to add a Obfs4 bridge into the custom bridges textfield, This could be useful for countries that censor/block all bridges that ship with TorBrowser/onionshare.

mig5:
Fixes #556

Tested with both normal and obfs4 custom bridges, but I would appreciate some looking at the regex pattern match - it's not my strong suit and I deliberately made the suffix 'lazy' for simplicity.

~~~I think it will also ignore ipv6 addresses for now, bonus points if someone can tweak the regex to support either.~~~ Edit: now supports ipv6 too, please do test.


Baccount:
Thank you @mig5 .
Also I think the link on the settings page `https://bridges.torproject.org/bridges` be changed to
`https://bridges.torproject.org/options` because the new link will show the option for Obfs4

Baccount:
Thank you @mig5 
I also think the link in the settings dialog should also be updated to support obfs4. here is my pull that should do that. https://github.com/micahflee/onionshare/pull/558/commits/72dbf6419a50f2b2e4f95206cf26606eba2851cb

PR #558, opened 2018-01-19
Update bridge link to support obfs4

Baccount:
Update bridge link to support obfs4


Baccount:
Changes the link in the settings dialog to allow users to easily find Obfs4

PR #560, opened 2018-01-19
Eject dmg

Baccount:
When installing The Dmg should be ejected not on unmounted. If you just on unmount the Dmg will still show up inside of Disk Utility.


mig5:
Yep, looks good!

PR #565, opened 2018-01-22
Support deleting an item from the file widget with the delete key on a keyboard

mig5:
This just speeds up the UX a bit, allowing the Delete key on a keyboard to remove items from the list rather than click the Delete qPushbutton.

### DIFF
--- a/onionshare_gui/file_selection.py
+++ b/onionshare_gui/file_selection.py
@@ -127,6 +127,24 @@ class FileList(QtWidgets.QListWidget):
             event.ignore()
         self.files_dropped.emit()
 
+    def keyPressEvent(self, event):
+        """
+        Allows deleting of items in the file list via the Delete key.
+        """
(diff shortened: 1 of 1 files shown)


Baccount:
This does not seam to work on macOS 10.13.2 (17C205).

mig5:
On my macOS it seems to work if I hold down the Fn key and press Delete.

Possibly related bug affecting macOS: https://forum.qt.io/topic/35591/single-key-qaction-shortcuts-don-t-work-on-mac-os-x-qt-5-x

Also see the comparison in keysequences per OS http://doc.qt.io/qt-5/qkeysequence.html#standard-shortcuts - so it's possibly not a bug, just pecularities across OS

mig5:
We could probably do more work to make it select/highlight the next item in the list though, so I'll mark this as needs more work.

mig5:
Ugh somehow this also breaks shift+select so I give in.

PR #566, opened 2018-01-23
#535 do the forced update check via a (non-blocking) QThread

### LINKED ISSUE #535: When I click the "check for updates" button, the app will become unresponsive until a result alert dialog is displayed
When I click the "check for updates" button, the app will become unresponsive until a result alert dialog is displayed. 
 https://stackoverflow.com/questions/3221314/asynchronous-programming-in-python

mig5:
I'm not sure if this is the best method, but this PR changes the 'Forced' update checker to be via a new instance of the UpdateThread QThread, so it doesn't block the app by calling UpdateCheck directly.

I couldn't find a way to do it any other way (e.g messing with socket timeouts), ultimately I think threading or queuing is the best method, and we already have the thread logic, so we may as well reuse it?

The thread only lingers until the check is complete, then it closes the thread.

Introduces a new parameter to UpdateThread 'force', so that it can carry through to the underlying UpdateCheck from the thread.

For me this fixes #535 on both macOS and Windows.

### DIFF
--- a/onionshare_gui/settings_dialog.py
+++ b/onionshare_gui/settings_dialog.py
@@ -557,31 +557,34 @@ class SettingsDialog(QtWidgets.QDialog):
         self._disable_buttons()
         self.qtapp.processEvents()
 
+        def update_timestamp():
+            # Update the last checked label
+            settings = Settings(self.config)
+            settings.load()
+            autoupdate_timestamp = settings.get('autoupdate_timestamp')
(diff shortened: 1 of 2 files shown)


Baccount:
This works great on macOS 10.13.2 (17C205), Great job mig5.

micahflee:
Looks good 👍

micahflee:
Now I've tested in both Mac and Windows

PR #573, opened 2018-01-26
Update onion.py to add support for the Next Generation Onion Service

Baccount:
Update onion.py to add support for the Next Generation Onion Service, right now tor.real needs to be manually replaced with tor.real 0.3.3
**Update.**
this pull request was just a example showing the code I changed to get V3's working, likely many things are broken.

### DIFF
--- a/onionshare/onion.py
+++ b/onionshare/onion.py
@@ -427,12 +427,12 @@ class Onion(object):
             basic_auth = None
 
         if self.settings.get('private_key'):
-            key_type = "RSA1024"
+            key_type = "ED25519-V3"
             key_content = self.settings.get('private_key')
             common.log('Onion', 'Starting a hidden service with a saved private key')
         else:
             key_type = "NEW"
-            key_content = "RSA1024"
+            key_content = "ED25519-V3"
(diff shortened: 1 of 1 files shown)


mig5:
This will break v2 I think, as the Stem controller will probably fail to understand this parameter and not request an RSA1024 key

We also need to do a lot more work to detect the version and if 0.3.3 or higher, disable the Client Auth since that gets ignored in v3

Finally there is no guarantee in your first change, where a private key is being reused, whether or not it's an ED25519-V3 key

So although it's exciting that v3 onions might be soon capable, I'll mark this as needs more work as there is lots yet to be done.

Baccount:
I was just using this pull request is example showing the changes I made to get the V3 working, it's likely broke other things.

mig5:
No worries! I read/reviewed this before I saw you edited your comment in the other ticket.

It's a good start and I'll probably incorporate this into further work to do some version checks and other logic. Hopefully then when OnionShare starts bundling tor 0.3.3 (or users using Tor Browser mode, get a version of TB that uses 0.3.3), v3 shares will 'just work'.

PR #585, opened 2018-02-05
Second attempt at patches for an OpenBSD port

StAlphonsos:
… this time based off of 1.2 and after feedback from the upstream and other contributors at
https://github.com/micahflee/onionshare/pull/489

Instead of the "unixy" stuff I just add 'BSD' as a possible return value for
get_platform() and use it in several places.  The idea of collapsing everything
down to Darwin/Windows/Unixy doesn't quite work out because we need to
distinguish between Linux and BSD in one place, so I thought this way of doing
it made it read more naturally.  I also tried to deal with an issue in setup.py
in a way that might be acceptable to you.

Cheers, -A

### DIFF
--- a/onionshare/onion.py
+++ b/onionshare/onion.py
@@ -316,7 +316,8 @@ class Onion(object):
             # guessing the socket file name next
             if not found_tor:
                 try:
-                    if self.system == 'Linux':
+                    plat = common.get_platform()
+                    if plat == 'Linux' or plat == 'BSD':
                         socket_file_path = '/run/user/{}/Tor/control.socket'.format(os.geteuid())
                     elif self.system == 'Darwin':
                         socket_file_path = '/run/user/{}/Tor/control.socket'.format(os.geteuid())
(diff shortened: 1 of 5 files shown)


onionshare/onion.py
```diff
@@ -316,7 +316,8 @@ def connect(self, settings=False, config=False, tor_status_update_func=None):
             # guessing the socket file name next
             if not found_tor:
                 try:
-                    if self.system == 'Linux':
+                    plat = common.get_platform()
+                    if plat == 'Linux' or plat == 'BSD':
                         socket_file_path = '/run/user/{}/Tor/control.socket'.format(os.geteuid())
```
mig5: A minor nitpick - nothing wrong here - but rather than initialise a new variable 'plat' here, perhaps it would be better to be consistent and change line 134 of onion.py ```self.system = platform.system()``` to ```self.system == common.get_platform()``` ? 

And then you can just make this ```if self.system == 'Linux' or self.system == 'BSD'```? 

What do you think?

StAlphonsos:
I've updated my branch as per @mig5's suggestion - ditch new var, use self.system instead.

mig5:
This worked for me on FreeBSD at least with the 'dev_scripts', so I'm happy to merge this in.

PR #586, opened 2018-02-06
Add [da] translations

scootergrisen:
Perhaps add info about this in https://github.com/micahflee/onionshare/wiki/Translating so translators can know they need to translate this file also.
Or even better add the strings to the .json translations files so alle strings are in one place.

### DIFF
--- a/install/onionshare.desktop
+++ b/install/onionshare.desktop
@@ -1,9 +1,11 @@
 [Desktop Entry]
 Name=OnionShare
 Comment=Share a file securely and anonymously over Tor
+Comment[da]=Del en fil sikkert og anonymt over Tor
 Exec=/usr/bin/onionshare-gui
 Terminal=false
 Type=Application
 Icon=/usr/share/pixmaps/onionshare80.xpm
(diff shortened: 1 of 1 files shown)


mig5:
I'm not sure we can inject strings from the json into the nautilus file, though perhaps it could be template-ized somehow and created on intallation... @micahflee ?

Good idea re: the wiki - anyone can edit the wiki, feel free to add your improvements! :)

scootergrisen:
I don't seem to have permission to edit wiki
Here i see Edit button top right: https://github.com/audacity/audacity/wiki
Here i see no Edit button: https://github.com/micahflee/onionshare/wiki/Translating

mig5:
OK, strange, it used to be editable by any Github user I thought (or at least contributor)

micahflee:
Sorry about that. I recently went through all of the github repo settings and locked them down, including checking "Restrict editing to collaborators only: Public wikis will still be readable by everyone."

I just checked it, so @scootergrisen you can go ahead and edit the wiki if you'd wish.

PR #587, opened 2018-02-07
Added sort links under the table, resolved #491

### LINKED ISSUE #491: The files on the download page should be sorted by size
The files on the download page should be sorted by size, where the larger files will be onto and the smallest files at the bottom of the list.
issue

ImDevinC:
Took a crack at #491 with an idea I figured would work without requiring JS and without forcing a user to be stuck with what the sender provides. Fairly new to Flask and this project as a whole, so feel free to yell at me for anything I've done wrong.

### DIFF
--- a/onionshare/web.py
+++ b/onionshare/web.py
@@ -214,7 +214,7 @@ def index(slug_candidate):
     # Allow the user to sort based on the file parameters and
     # default to searching by ascending basename if the user
     # hasn't chosen an optionTest
-    valid_sorts = ['basename', 'size_human']
+    valid_sorts = ['basename', 'size']
     sort_by = request.args.get('sortby', 'basename')
     if not sort_by in valid_sorts:
         sort_by = 'basename'
--- a/onionshare/web.py
+++ b/onionshare/web.py
@@ -211,15 +211,28 @@ def index(slug_candidate):
             r.headers.set(header, value)
         return r
 
-    # If download is allowed to continue, serve download page
+    # Allow the user to sort based on the file parameters and
+    # default to searching by ascending basename if the user
+    # hasn't chosen an optionTest
+    valid_sorts = ['basename', 'size_human']
+    sort_by = request.args.get('sortby', 'basename')
+    if not sort_by in valid_sorts:
+        sort_by = 'basename'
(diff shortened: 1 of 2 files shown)


mig5:
Hi @ImDevinC , thanks for the PR!

It's an interesting approach. I've just tested it, but it doesn't quite seem to work for me. See the below screenshot where I have sorted alphabetically (the default)


Now sorting by size:

I would expect the MANIFEST.in to appear after the stdeb.cfg, and before the the setup.py.

Also the folders remain at the top alphabetically, but that might be a bit tricky to change with a non-javascript method.

Other than this, because the lack of javascript means that to sort involves another GET request with the query parameters, this can be rather 'slow' to sort given it's an Onion address. That's not 'wrong' but it's a quirk of this method.

@micahflee and I were actually talking about this feature today and decided that maybe it's OK to do this via Javascript. If the user has opted to allow Javascript in their Tor Browser (danger, Will Robinson!), then they can sort. If the user doesn't allow Javascript, they simply see the existing alphabetical method.

Do you want to see if you can accomplish this in Javascript, at least for curiosity's sake? It will be a better experience for the javascript-enabled user, and it might be a more reliable method in terms of sorting the sizes correctly (not to mention therefore being able to sort the folders as well as files since it's all in the one table).

[Various](https://stackoverflow.com/questions/14267781/sorting-html-table-with-javascript) [methods](https://www.w3schools.com/howto/howto_js_sort_table.asp) [seem](https://kryogenix.org/code/browser/sorttable/) to exist for this in Javascript but I haven't tried them with any level of success yet.

Thanks again!

ImDevinC:
I'll see about taking a look at integrating JS into it, but I did find the issue with your sorting issue as I was using the human_size (3.0 KiB) instead of size (3072 b). I updated the PR with that new value for now. I'll try to find some time later this week to look at the JS implementation as well.

micahflee:
Just merged #608 so this isn't needed anymore. But thank you for implementing @ImDevinC and I'm definitely open to merging future PRs, too.

ImDevinC:
Thanks, completely forgot about the javascript implementation. Glad there's a solution in place!

PR #590, opened 2018-02-07
Hide the FileList delete buttons when a share is running, so items can't be deleted mid-share

mig5:
I couldn't find a way to just disable the QPushButtons of each QListWidgetItem, so this was the easy way out.

I hope it isn't bad UX e.g user might be confused why the files are all 'greyed out' once they start the share.

### DIFF
--- a/onionshare_gui/file_selection.py
+++ b/onionshare_gui/file_selection.py
@@ -205,16 +205,16 @@ class FileList(QtWidgets.QListWidget):
                 self.takeItem(itemrow)
                 self.files_updated.emit()
 
-            item_button = QtWidgets.QPushButton()
-            item_button.setDefault(False)
-            item_button.setFlat(True)
-            item_button.setIcon( QtGui.QIcon(common.get_resource_path('images/file_delete.png')) )
-            item_button.clicked.connect(delete_item)
+            item.item_button = QtWidgets.QPushButton()
+            item.item_button.setDefault(False)
+            item.item_button.setFlat(True)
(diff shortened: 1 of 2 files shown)


micahflee:
This is good, and I would merge it, but disabling the whole file list also disables the scroll bar. I can see cases where someone is sharing a lot of files, and after they start the server and share the link they want to double-check the files they're sharing before the person downloads it or something.

screenshot_2018-02-06_21-54-26

Here's the bit of code where the `item` and `item_button` are defined:

```python
# Create a new item
item = QtWidgets.QListWidgetItem(item_name)
item.setToolTip(size)
item.setIcon(icon)

# Item's delete button
def delete_item():
    itemrow = self.row(item)
    self.filenames.pop(itemrow)
    self.takeItem(itemrow)
    self.files_updated.emit()

item_button = QtWidgets.QPushButton()
item_button.setDefault(False)
item_button.setFlat(True)
item_button.setIcon( QtGui.QIcon(common.get_resource_path('images/file_delete.png')) )
item_button.clicked.connect(delete_item)
```

You can instead attach the the `item_button` variable to the `item` object, eg:

```python
item.item_button = QtWidgets.QPushButton()
```

Then you'll be able to look through the items that are inside `FileList` which is a `QListWidget` to access `item.item_button`.

micahflee:
Actually, maybe hide the buttons when the server is active, and show them again when the server stops.

mig5:
Oh duh! I couldn't think of how to access the item_button, thanks for giving me the answer.

Pushed a new commit that does exactly that.

PR #592, opened 2018-02-08
[ux-update] Fix shutdown timer

mig5:
This fixes the shutdown timer.

Whilst working on it, I discovered these certain status bar messages are now too long (because the server status icon/message is on the right) and so get cut off:

'Closing automatically because download finished'
'Closing automatically because timeout was reached'
'Waiting for download to complete before stopping'

You can't word wrap the setMessage() with a QStatusBar. Instead, you have to create a (word wrapped) QLabel widget and then use insertWidget() or addWidget() to put it in the Status Bar. So that's what I've also done here.

However, that introduces another 'squish' problem. Look at the add/delete buttons once the server stops - they are breaking into the FileList widget.

screenshot_2018-02-08_11-30-35

I really feel like we might solve problems like this by just setting a fixed minimum width and then not have to worry about word-wrapping at all... as these keep cropping up. However, maybe there is some easy fix for not allowing the Add/Delete buttons to overlap the File List. I haven't found it yet, but haven't looked too hard. Maybe you can merge this and make a quick fix for it if you know it.


mig5:
Worked around the add/delete button issue, for now, by actually reducing the length of those 'Closing because...' messages themselves, so they only wrap 2 lines on the statusBar.

micahflee:
What if, instead of conveying this sort of information in the status bar at all, we use desktop notifications instead? This way there's no limit to how long the messages can be, and we don't need extra labels in the status bar or word-wrapping?

The only downside I see is that "Show desktop notifications" is a setting, and people who disable it will miss important information... but I mean, they did disable the setting to notify them of things.

@glenn-sorrentino what do you think? When the timeout stops, here's what @mig5's work looks like in the status bar, word-wrapped. Do you think it's better to keep doing things like this, or to switch to desktop notifications only? If you want to see an example of desktop notifications, OnionShare something to yourself, and it will display a notification when the download completes.

screenshot_2018-02-10_10-17-59

glenn-sorrentino:
Two questions:
1. How important is that message to the user? 
2. Does that message mean that the timer expired?

micahflee:
1. I don't think it's _that_ important, but I'm sure sometimes users would feel better knowing that the sensitive file they're sharing is actually no longer on the internet anymore after half an hour, or whatever they set the timer to.

2. Yup, that's exactly what it means. But there are a few other similar messages too, like "Stopped because download finished", or "Waiting for download to complete" -- when the timeout finishes but the download is in progress, we don't want to cancel the download.

mig5:
The only thing with systray notifications is, they disappear after a few seconds. 

I wonder if there’s any risk of confusion in this scenario:

User sets a timer and starts a share
User walks away from computer
Timer expires
User comes back to their computer, finds share has stopped
User can’t remember if they actually started a share or just meant to but got distracted

At least the status bar notification lingers long enough to provide feedback.

(I also like Glenn’s use of the word ‘expired’ rather than ‘reached’ - I might introduce that here either way)

micahflee:
That's a good point @mig5! I think the crux of the UX decision is, what are the desktop notifications for, and what is the status bar for?

Since desktop notifications are basically ephemeral, they're for drawing your attention to something that's happening immediately in case you want to pay attention. And the status bar is for giving you pertinent information about the current status of onionshare, that won't disappear immediately.

If this is the case, maybe we should keep these wrapped messages in the status bar like you've done @mig5, but then also reconsider some of the other uses of the status bar. For example, when the address gets copied to the clipboard it temporarily updates the status bar with the message "Copied address to clipboard". Maybe that would be better as a notification?

mig5:
Good idea.

Note in your screenshot above: just realised, even though I mostly resolved the squish issue with shortening the messages and some CSS, those Add/Delete buttons are *slightly* closer to the File List widget than they usually are. So there is still a widget squish issue here when the status bar messages wrap, that I’d like to resolve here (but haven’t found how!)

For tweaking other messages to only appear in desktop notifications, I think I’ll make a separate PR, to try and keep this one specific to fixing the shutdown timer & it’s associated messages.

micahflee:
Awesome, sounds good.

So how about you post a comment when you think you've resolved the squish issue and I'll do a final review and merge. Let's just not worry about some translations potentially wrapping the status bar to a third row -- we don't know if this will be an issue or not, and if it ends up being an issue we can always solve it by increasing the minimum width of the window so it wraps in only 2 lines.

onionshare_gui/onionshare_gui.py
```diff
@@ -132,13 +132,27 @@ def __init__(self, onion, qtapp, app, filenames, config=False):
...
         self.status_bar.addPermanentWidget(self.server_status_indicator)
         self.status_bar.addPermanentWidget(self.settings_button)
         self.setStatusBar(self.status_bar)
 
         # Status bar, zip progress bar
         self._zip_progress_bar = None
+        # Status bar, other larger messages
+        self._close_on_timeout_label = None
+        self._closing_automatically_label = None
+        self._timeout_download_still_running_label = None
```
micahflee: Is it necessary for this to be three separate labels? Could all three of these things just update the same label? This way, if we have more messages like this in the future, we don't need to continue to add more labels.

mig5: Good call - I don’t know why I didn’t do it that way :)

mig5:
@micahflee I'm sorry, I just can't crack the widget squish/layout resizing issue.

Instead I've set the status label (now a single QLabel) to non-word-wrapped, and I've set the overall app window to 430 instead of 350.

This fixes the issue for the current largest status message, without hiding the Settings button that re-appears, and without other squish issues like the add/delete buttons overlapping the File List.

It possibly also resolves the drag/drop 'total number' overlap issue reported by @Baccount in the other ticket too.

Counting my losses here with this 'easy way out', let me know what you think.

In this commit I accidentally also introduced a pluralisation of your most recent 'Total files / size' top widget: if only 1 file, it reports '1 File', rather than '1 Files'.

I'll now move on to other stuff such as moving certain messages to Desktop notifications etc..

PR #595, opened 2018-02-08
[ux-update] Set the URL label to explicitly mention when persistent. 

mig5:
Remove older Persistence warning, which would otherwise appear below the URL labels, to save clutter (lots of text there already..). 

Add ToolTip to explain what Persistence means. Hovering over the URL label will show the tool tip.

The 'close automatically on download' URL label remains as 'Your One-Time Download Address' without tooltip. And with 'close automatically' off, it remains 'Your Download Address' without tooltip too.

Let me know what you think about this.

screenshot_2018-02-09_07-53-16

### DIFF
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -140,16 +140,10 @@ class OnionShareGui(QtWidgets.QMainWindow):
         # Status bar, zip progress bar
         self._zip_progress_bar = None
 
-        # Persistent URL notification
-        self.persistent_url_label = QtWidgets.QLabel(strings._('persistent_url_in_use', True))
-        self.persistent_url_label.setStyleSheet('font-weight: bold; color: #333333;')
-        self.persistent_url_label.hide()
-
         # Primary action layout
         primary_action_layout = QtWidgets.QVBoxLayout()
         primary_action_layout.addWidget(self.server_status)
         primary_action_layout.addWidget(self.filesize_warning)
-        primary_action_layout.addWidget(self.persistent_url_label)
         primary_action_layout.addWidget(self.downloads_container)
         self.primary_action = QtWidgets.QWidget()
         self.primary_action.setLayout(primary_action_layout)
@@ -409,9 +403,6 @@ class OnionShareGui(QtWidgets.QMainWindow):
                 self.stop_server()
                 self.start_server_error(strings._('gui_server_started_after_timeout'))
 
-        if self.settings.get('save_private_key'):
-            self.persistent_url_label.show()
-
     def start_server_error(self, error):
         """
         If there's an error when trying to start the onion service
@@ -443,7 +434,6 @@ class OnionShareGui(QtWidgets.QMainWindow):
         # Remove ephemeral service, but don't disconnect from Tor
         self.onion.cleanup(stop_tor=False)
         self.filesize_warning.hide()
-        self.persistent_url_label.hide()
         self.stop_server_finished.emit()
 
         self.set_server_active(False)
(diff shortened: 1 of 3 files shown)


micahflee:
This is excellent. The only thing I wonder is if we should have a separate label for Persistent One-Time downloads. The language we're using gets confusing, but this one label is trying to describe these four states:

- Close automatically is FALSE, persistence is FALSE
- Close automatically is TRUE, persistence is FALSE
- Close automatically is FALSE, persistence is TRUE
- Close automatically is TRUE, persistence is TRUE

We only have 3 different labels, and both persistence states are covered with one. I think the idea of a "one-time" download (only let one client download the files, then close when the download completes) is still accurate but is more confusing when you add in the idea of a persistent address (same URL, but potentially downloading different files).

Maybe @glenn-sorrentino has some insights into better wording. But in the meantime, this is better than what's already there so I'm happy to merge it.

glenn-sorrentino:
Agree that the word "persistent" might confuse the user. It might even be redundant to say "persistent". If the language was "Your download address" I think that it's clear the address is only active when you're sharing. 

There's also a general feeling that if you need hover/assistance text, the language you're using is too confusing anyway. 

I'd take away the text in parentheses and just say "Your Download Address:"

micahflee:
So "persistent" is a new feature that has special meaning. By default, every time you share something, you create a _different_ address. If you enable "Use a persistent address" in the settings, then this is no longer true -- every time you share anything, the address will now be _the same_.

This can be useful in very specific situations. For example, I can meet you in person and hand you a piece of paper with an OnionShare address on it and a date and time. At that date and time, you can load that address to download a specific file from me. This way we never have to communicate the address over the internet, which is difficult to do anonymously or without leaving a trace. But this is only possible if I'm able to know the address ahead of time, e.g. by using the persistent feature.

But it can also be a security problem. If I have "Use a persistent address" checked and I use OnionShare to send someone a low security file, and I send them the address over Slack, and then later I send someone else a sensitive file, I wouldn't want to accidentally use the same address that I already posted in Slack -- I'd want a brand new address. This is why I think it's important to let people know if they're using persistent mode or not.

This is a totally separate concept than the "Stop sharing after first download" setting. If that is checked, then only one person can download my file and as soon as they succeed server stops (hence the "one-time" download). But if it's not checked, then several people can download the file and the server doesn't stop until I tell it to. This might be useful if I want to send a file to a group of people.

mig5:
TBH the introduction of ‘one-time’ in the label confused me at first, because I take ‘one-time’ to mean ‘opposite of persistent’ eg ‘changes every time’. I didn’t equate ‘one-time’ with ‘close automatically after first download’. To me, both close automatically/stay open are ‘one-time’ URLs, as you won’t see them again.

Maybe we can just have ‘your download address’ and ‘your persistent download address’?

Glenn makes a good point about ‘if you had to tooltip, you’ve already done it wrong’, but I thought it was important here because of the potential security risk as Micah notes.

Instinctively I feel that the *less* that label can say something different, the better.

micahflee:
There's also a potential (but I think lower) security risk with the "Stop sharing after first download" unchecked. Maybe you share a sensitive file, send someone the address, and then leave your computer for many hours expecting that the files will no longer be accessible on the internet when you return. But even if the user finished downloading the file, the files will still remain accessible.

Maybe it's best to be just exactly clear about the state of the current share, like:

- Your Download Address
- Your Download Address (one-time download)
- Your Download Address (persistent address)
- Your Download Address (one-time download, persistent address)

glenn-sorrentino:
Since the user has explicitly opted into using a persistent address the language is something that they're familiar with already. I might explore removing the "(what's this?)" and replacing it with an ⓘ icon that has a hover state. 

Depending on the severity of persistent addresses being a security risk, we might even consider a more "warning" style treatment. Orange text with a ⚠ symbol?

glenn-sorrentino:
@micahflee I like it... maybe the hover text is what you've put in parenthesis.

mig5:
I've made an updated version in #601, let me know your thoughts there.

PR #596, opened 2018-02-09
Allow the user to cancel a share that is still starting up

mig5:
This is my attempt at allowing the user to Cancel a share that is still in the process of starting up (e.g in SERVER_WORKING state).

As per page 12 of Glenn's UX spec.

I change start_onion_service to run in a QThread instead of threading.Thread, because QThread has ways to kill the thread, whereas the latter doesn't. 

So we get quite brutal and use terminate() to totally hang up that startup process. I first tried with quit() but this didn't work, it still got to the end of the Onion obtaining and started the web server etc, e.g it was allowed to run til the end of the start_onion_service() function, so I went with terminate() instead, which is nasty, but works.

I have verified that the onions are successfully deleted on cancellation and do not linger around.


mig5:
My original words in the PR mentioned that I thought there was some bug in deleting the onions on cancellation - turns out it was just caused by bugs on my behalf, spent too many hours on this to notice.

I have fixed the bugs in deleting the onion and strace confirms successful deletion of the onion as soon as the user 'cancels'.

I now think this PR is working perfectly, appreciate others testing.

micahflee:
Testing now

micahflee:
Actually, do you mind resolving the conflict first? I'm trying to but not completely clear on what has changed with the `button_clicked` slot.

mig5:
Should be all good now. I think we just edited the same spot in code (introducing new 'canceled' slot for this PR, in the same place the button_clicked slot was introduced in another, which was just about clearing statusbar messages)

PR #606, opened 2018-02-14
Reverse the order of the files when adding them

### LINKED ISSUE #604: Files are not sorted alphabetically in the File List (leads to deleting wrong row from list, etc)
When I start a share with a large file then cancel the share and  remove the large file and add a very small file I still get the message, "warning sending large files could take hours".
1. Start a share with a file over 200 Mbs
2. when the share starts cancel it.
3. remove the large file and add a very small file, I used a 8 Kb file
4. start a share and you should get the message "Warning sending large files could take hours", even though the only file being shared is very small.
Tested with "ux-update " branch

mig5:
Fixes #604 

Now that file items are QListWidgetItem widgets, the act of 'addWidget()' is putting them into the list in reverse order alphabetically - at least visually.

This means deleting items from the list is actually popping the wrong item from the list.

I couldn't seem to fix this with QListWidget's self.sortItems() despite trying it in various spots.

This fix is counterintuitive and a bit hacky, but basically reverses the order of the filenames when sending to the QListWidget to be added as item widgets. This ensures addWidget() ends up adding the item widgets to the list in the originally intended order, and the correct item should be deleted from the list.

### DIFF
--- a/onionshare_gui/file_selection.py
+++ b/onionshare_gui/file_selection.py
@@ -336,7 +336,9 @@ class FileSelection(QtWidgets.QVBoxLayout):
         """
         file_dialog = FileDialog(caption=strings._('gui_choose_items', True))
         if file_dialog.exec_() == QtWidgets.QDialog.Accepted:
-            for filename in file_dialog.selectedFiles():
+            selected_files = file_dialog.selectedFiles()
+            selected_files.sort(reverse=True)
+            for filename in selected_files:
                 self.file_list.add_file(filename)
 
         self.file_list.setCurrentItem(None)


Baccount:
@mig5 the bug still seems to be occurring even with this patch.
1. I add a 250Mb file and start the share.
2. stop share and remove the 250Mb file and add a very small file "I used a file around 14KBs"
3. start the share and the bug in https://github.com/micahflee/onionshare/issues/604 will occur.

mig5:
OK, I don't reproduce that at all, but I'll close this, maybe someone smarter than me can fix it.

Baccount:
@mig5  it seams to work correctly, I must of not copied file_selection.py before compiling it.

PR #608, opened 2018-02-14
Make table sortable with javascript (if enabled)

mig5:
I took a stab at #491 and #587 given that the ux-update branch has changed the HTML.

If javascript is enabled, the sorting by size works as expected.

It still defaults to alphabetical order with folders on top.

A curious quirk that I couldn't work out how to resolve, was sorting by filename (the folders get put at the bottom, e.g opposite of being on top, instead of being merged in alphabetically like they are by size). Totally am not a javascript person so maybe someone else can resolve. But you can return to the default sort easily and the main aim was to sort by size, so that works at least..

For sorting the table by header, I used the second example from https://www.w3schools.com/howto/howto_js_sort_table.asp as my guide.

For converting human readable sizes back to bytes in order to sort accurately, I used a modification of the answer at https://stackoverflow.com/questions/6974614/how-to-convert-human-readable-memory-size-into-bytes

### DIFF
--- a/onionshare/web.py
+++ b/onionshare/web.py
@@ -59,7 +59,7 @@ zip_filename = None
 zip_filesize = None
 
 security_headers = [
-    ('Content-Security-Policy', 'default-src \'self\'; style-src \'unsafe-inline\'; img-src \'self\' data:;'),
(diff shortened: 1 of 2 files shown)


mig5:
Oh crap, I was so focused on fixing the 'human readable size' issue, I forgot to account for sorting files that are already in bytes :(

It's not my day.. back to the drawing board, sorry.

micahflee:
Look good

PR #616, opened 2018-02-19
Ensure we don't break out of the loop too early when simultaneous downloads are running and one completes first

### LINKED ISSUE #614: Simultaneous downloads is failing for the second download
With 'close automatically after first download' turned off, there is a bug affecting simultaneous downloads.

While running two downloads at once of the same share - one slightly ahead of the other - once the first one completes, the progress bar of the second download in the OnionShare GUI 'freezes' e.g at 88%.

In the Downloads tab of my browser, the second download continues for a little way (due to the discrepancy between progress sent to Tor and what's come out the other side, as usual), but the download suddenly 'completes' (but really cancels: the file size, in this case 9.9MB, is only 8.5MB, and the zip is 'corrupt', so the download is only partial).

Whilst watching the progress bars of both downloads, in the browser, it seems as though once the first download gets 100% of the file client-side, that's around when the other download fails too - as if its stream is closed when the first one is complete. It happens that way for me every time.

It happens if I am downloading the share twice in the same Whonix browser, but also if I am downloading the share via Whonix on Qubes, and Tor Browser on Mac, at the same time (the Mac download completed only partially in the same way). So it's not unique to using the same Tor connection client-side.

I verified the bug occurs in the develop branch as well as ux-update, so it predates the UX work. It also looks like it's present in the 1.2 release.

It occurs regardless of Tor connection method or whether I'm running OnionShare on Linux or Mac.

My guess is there's some sort of weird collision at the Flask layer when handling simultaneous downloads, but I don't know how best to debug..

mig5:
This seems to fix #614

I think I broke it when I registered 'done' as a global in the web component. This was done so that we could avoid stopping a download that is in progress when the shutdown timer ran out.

But in doing so I think I cause 'done = True' to stick, so that when 1 download finishes, a simultaneous download closes early. Sorry..

### DIFF
--- a/onionshare/web.py
+++ b/onionshare/web.py
@@ -305,6 +305,7 @@ def download(slug_candidate):
                         sys.stdout.flush()
 
                     add_request(REQUEST_PROGRESS, path, {'id': download_id, 'bytes': downloaded_bytes})
+                    done = False
                 except:
                     # looks like the download was canceled
                     done = True


micahflee:
Excellent, I confirmed this fixes the bug.

Issue #626, opened 2018-02-23
[ux-update] The status bar messages are still too long on macOS, cuts off Settings icon


mig5:
It seems macOS adds a lot of padding/spacing in Qt. 

The status bar messages, which we tried to shorten in length word-wise, are still slightly too long on my Mac.

In this case 'Stopped because download finished' means it slightly cuts off the Settings Icon on the right.

<img width="441" alt="screen shot 2018-02-23 at 2 29 10 pm" src="https://user-images.githubusercontent.com/99173/36577535-fa341bc6-18ab-11e8-9226-86c3a8ccc69e.png">

PR #629, opened 2018-02-24
Re-work the UI widgets to make it smaller on Macbook Pro

### LINKED ISSUE #625: [ux-update] The app can grow too long for Macbook Pro 13"
On MacBook Pro 13" the overall app size can grow too long (vertically) for the laptop's screen size.

The most common cases of this are when there are multiple download progress bars, or when the auto-stop timer widget is present.

You can't resize the window to be smaller vertically. So you actually lose all visibility of the status bar.

I think we might have to think about removing some of the verbose URL label/description stuff, or possibly move the Copy URL / Copy HidServAuth buttons so that they are in a HBox alongside the Download info. Even then, that may not be enough.

screen shot 2018-02-23 at 2 36 48 pm
screen shot 2018-02-23 at 2 16 58 pm

mig5:
I think this fixes #625 

Changes in this PR:

1) Removed the extra 'Your Download Address' label - I think the other description ('Anyone with this link can download your...') is explanatory enough?

2) Moved the URL type 'tooltips' over to that longer URL description

3) Hide the shutdown timeout widget when a share starts, rather than just disable editing it. This is actually more consistent with what we've done elsewhere (hide widgets the user can't do anything to, show them only when they can)

4) Because the shutdown time is then 'hidden' from the user, I implemented a 'countdown' timer in the 'Stop Sharing' button so they can see the timer is progressing.

5) There is also a tooltip when hovering over the 'Stop Sharing' button to say 'This share will automatically stop at XXXXXXX'

With all these changes, the UI actually fits on my Macbook Pro 13" in its 'biggest' form (e.g scenario where shutdown timer is set and the download progress bars are shown).

What do you think?

screen shot 2018-02-24 at 3 06 49 pm


mig5:
There is one case I still see when the UI 'overlaps' the dock on a Mac - ironically, that's the 'default' scenario where the share stops after first download.

For some reason (and this is true on Linux too) the Download progress bar container is really big

Yet when you allow multiple downloads, the progress bar container is actually smaller (because the scroll bar kicks into gear)

How I am tired of these Download containers! I wish there was a sane way of getting rid of them and putting the progress data in the 'info' widgets at the top...

screen shot 2018-02-24 at 3 51 14 pm

micahflee:
Nice, I've tested this and it looks good.

But I wish we could solve the underlying problem, that there's way more spacing everywhere on Mac than in Linux and Windows for some reason (which would also help make this look better, as well as #626). I've been attempting to see if I can use style sheets to fix the problem across the board, but no luck so far.

micahflee:
Just opened #630 to address this more in the future. But in the meantime, this PR looks good and I'll merge it.

Issue #632, opened 2018-02-24
Move downloads progress into download info widget


micahflee:
As @mig5 referenced in #629, it would be good to move the download progress out of the main window where it takes up a ton of space, and somewhere nice, probably the download info widget. Or maybe even the systray applet?

mig5:
One idea I had was to somehow write a toolTip for the info widget that writes the progress of the downloads, but appends with a <br> separator for where there are simultaneous downloads in progress. 

The tricky part would be removing a completed download's progress bar from the tooltip.

mig5:
See an implementation of this at #651

PR #636, opened 2018-02-24
When the server stops, resize the window to be small again instead of…

micahflee:
… having a huge downloads box.

Looks much better. Related to #632 but doesn't resolve it.

### DIFF
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -200,6 +200,7 @@ class OnionShareGui(QtWidgets.QMainWindow):
         self.layout.addWidget(self.info_widget)
         self.layout.addLayout(self.file_selection)
         self.layout.addWidget(self.primary_action)
+        self.layout.addStretch()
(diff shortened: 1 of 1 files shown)


mig5:
Unfortunately still looks a bit wonky on my Mac - the downloads container is much smaller, but the overall window is still much too big.

screen shot 2018-02-25 at 10 45 15 am

micahflee:
You know, I was getting that at first too, but then I stopped being able to reproduce it and it was actually resizing the window. Hrm...

micahflee:
I don't think we necessarily need to include this in the 1.3 release, too. I'm ok with having a few UI things that aren't perfect that we can put in a point release.

mig5:
It seems to depend on where the app is placed on the screen - when it's mostly 'above' the dock, it's fine, but if it's to the far right, half 'past' the end of the dock, it will end up overlapping the dock.

mig5:
I'm gonna close this, as I've realised #798 actually resolves this old issue.

PR #640, opened 2018-02-25
Revert cancel feature

### LINKED ISSUE #637: When I start a share, onionshare becomes unresponsive
When I start a share, onionshare becomes unresponsive, When I run it in --debug it gets stuck at
`[Feb 24 2018 14:41:39] Onion.Starting a hidden service with a new private key`
I captured a sample of the frozen process.
I have onionshare set to connect to TorBrowser and "Stop after first share" disabled
[Sample of OnionShare.txt](https://github.com/micahflee/onionshare/files/1755396/Sample.of.OnionShare.txt)

**I built and tested with the latest develop build**

**Update.**
I captured a full log when the app became unresponsive.

[full log.txt](https://github.com/micahflee/onionshare/files/1755404/full.log.txt)
@micahflee 
@mig5

mig5:
Broke my recent number one rule (don't add features, Mig, just try and fix bugs)

This reverts the cancel feature in order to fix #637 til we can work out WTF, QThread

I *think* this is a clean revert (had to fix conflicts same as when I did when we merged it in)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 * Client-side web interfact redesigned
 * New feature: Support for meek_lite pluggable transports (Amazon and Azure)
 * New feature: Support for custom obfs4 and meek-lite bridges
-* New feature: ability to cancel share before it starts
 * Bug fix: the UpdateChecker no longer blocks the UI when checking
 * Bug fix: simultaneous downloads (broken in 1.2)
 * Update Tor to 0.2.3.9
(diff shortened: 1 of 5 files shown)


micahflee:
That's a shame, I liked this feature, so it would be cool to figure out what the bug is and fix it at some point. But let me review this and merge it.

mig5:
Yep hopefully @Baccount can test it properly fixes it too.

It works perfectly on Linux so I can only assume it's something to do with threads working differently on OS X, or some bug in QThread in the version of Qt in use on OS X (didn't test if it's also broken on Windows)

PR #644, opened 2018-02-25
Fix the broken images in Windows

### LINKED ISSUE #641: [Windows] info widget icons and URL description {i} are wrong/missing
On Windows, the download icons in the Info widget (top of screen) as well as the 'information' icon on the URL description ('This share will not stop automatically', etc) show an incorrect icon.

I made sure they were added to the .nsi so I don't know why this is. One common trait other than those being 'new' images is that each of those widgets has a tooltip, so maybe it's tooltip related.

micahflee:
Fixes #641.

The problem is (at least my problem when I ran into it) is that the image path filename had spaces in it, and if you're specifying the src of an `` tag, you need to use quotes. This seems to fix the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ pip-log.txt
 .tox
 nosetests.xml
 .cache
+.pytest_cache
 
 # Translations
 *.mo
--- a/onionshare/common.py
+++ b/onionshare/common.py
@@ -57,7 +57,7 @@ def get_platform():
     """
     Returns the platform OnionShare is running on.
     """
-    plat = platform.system() 
+    plat = platform.system()
     if re.match('^.*BSD$', plat):
         plat = 'BSD'
     return plat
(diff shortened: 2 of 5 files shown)


micahflee:
Hmm, this causes a test to fails. It's kind of a confusing failure, and one that I don't think is simple to fix since it `os.path.dirname(sys.executable)` is definitely going to be different in Windows (the executable in Windows is `onionshare-gui.exe`, but in Linux and Mac while running the tests it's `python3`). I kind of think it might just be more straightforward to delete the test. I like testing for it, but if `common.get_resource_path` breaks in Windows while frozen, we will know because running `onionshare-gui.exe` will immediately break.

But also, I found another issue. Several of the tests fail when run in Windows.

mig5:
@micahflee this resolves the issue for me with the info widgets, but not the {i} icon in the url_description label - might need a similar encapsulation there?

Also do you want to fix the conflict in this branch before it gets merged?

screen shot 2018-02-25 at 2 23 39 pm

micahflee:
Oh I missed that one. Yes, lemme fix that too.

micahflee:
Ok @mig5 try reviewing it now

PR #645, opened 2018-02-25
Cancel share

mig5:
Back in #596 I said that using quit() on the QThread didn't seem to work (the share that was meant to be canceled, still started up). So I used terminate() but I now think that that's what caused the freezes on macOS.

Back to using quit() here and everything seems to work fine on macOS and Linux - I can't reproduce the crash.

I'd love if the always reliable @Baccount could test this out too!

Make sure you allow at least a minute or more after canceling the share to make sure it doesn’t actually end up still starting up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Client-side web interfact redesigned
 * New feature: Support for meek_lite pluggable transports (Amazon and Azure)
 * New feature: Support for custom obfs4 and meek-lite bridges
+* New feature: ability to cancel share before it starts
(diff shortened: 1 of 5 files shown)


micahflee:
I'm testing this now, both in Linux and macOS. I've started a share and waited for it to finish, then stopped it. Then started a share but canceled it. Then started one again and waited for it to finish. I can't reproduce the freeze in macOS, so I think this fixed it.

@Baccount I'd still appreciate your testing if you haven't already, but this appears to have solved it for me so I'll merge it.

Baccount:
@micahflee Testing right now.

Baccount:
I have not had any issues with this feature.

PR #646, opened 2018-02-25
Add ability to abort but not exit SettingsDialog/reconnect to Tor, on invalid settings. Do this for invalid bridges

### LINKED ISSUE #634: Should not revert to connecting to Tor directly when bridge syntax is invalid
According to the onionshare wiki: https://github.com/micahflee/onionshare/wiki/Using-Bridges

> Invalid bridge syntax
> 
> If you supply bridges in an invalid format, they will simply be ignored. If none of your bridges are valid, OnionShare will alert you about this, and revert to connecting to Tor directly.

"[R]evert[ing] to connecting to Tor directly" can be dangerous especially considering most of the people who use a custom bridge because they may want to hide the fact that they are using Tor.

Therefore, do you think it is a good idea to:
1. inform the user which bridge's syntax is invalid
2. do not auto connecting to the Tor network until user fixes the syntax error or explicitly choose to connect to the Tor network without a bridge?

mig5:
Here's something that fixes #634 (please test the hell out of it, it's late here and I've had a glass of wine, so migressions might be rampant)

This diff looks bigger than it really is, but it's just a code indentation.

Previously, no matter what, clicking 'Save' on SettingsDialog always emitted 'save clicked', which invoked the 'save_clicked' function, which always attempted to reconnect to Tor with the settings provided, and closed the SettingsDialog.

All this change does, is first check if the settings_from_fields() function actually returned something other than False, in which case do what it always did (as above). But if it returned False, do nothing, meaning that the SettingsDialog stays open.

And in doing so, we now explicitly return False on invalid custom bridges, which solves #634 . The user receives the same alert dialog, but remains in the SettingsDialog to correct the issue. 

~~Note that I still force it to save 'no_bridges' in the settings before returning False, because otherwise, if you click Cancel on the SettingsDialog, then next time you open OnionShare and the SettingsDialog, *none* of the bridge options are selected (but neither is 'Don't use bridges'), which is 'unnatural' - a radio option should always be selected. So we should still be setting 'no bridges' as the default if the user backs out via Cancel altogether (in doing so, we return to the default mode that OnionShare ships with: bundled mode, no bridges).~~

Edit I realise I'm wrong on the last paragraph, but maybe it's not that big a deal.. the issue being that settings.save() only occurs if settings_from_fields() returned True.. but in a way this is good: if you previously had valid custom bridges, and then you entered bad ones, and then you ignored the Alert and clicked Cancel, at least your settings preserve the original valid bridges..

### DIFF
--- a/onionshare_gui/settings_dialog.py
+++ b/onionshare_gui/settings_dialog.py
@@ -660,56 +660,57 @@ class SettingsDialog(QtWidgets.QDialog):
         common.log('SettingsDialog', 'save_clicked')
 
         settings = self.settings_from_fields()
-        settings.save()
+        if settings:
+            settings.save()
 
-        # If Tor isn't connected, or if Tor settings have changed, Reinitialize
-        # the Onion object
-        reboot_onion = False
-        if self.onion.is_authenticated():
-            common.log('SettingsDialog', 'save_clicked', 'Connected to Tor')
-            def changed(s1, s2, keys):
-                """
-                Compare the Settings objects s1 and s2 and return true if any values
-                have changed for the given keys.
-                """
-                for key in keys:
-                    if s1.get(key) != s2.get(key):
-                        return True
-                return False
+            # If Tor isn't connected, or if Tor settings have changed, Reinitialize
+            # the Onion object
+            reboot_onion = False
+            if self.onion.is_authenticated():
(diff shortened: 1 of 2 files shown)


micahflee:
Just tested it a bunch and it appears to work great.

irykoon:
Thank you so much for fixing and reviewing it! @mig5 @micahflee

PR #650, opened 2018-02-27
Add GUI unit tests

mig5:
Here are my current GUI unit tests to date

They are currently passing on develop branch, see https://jenkins.mig5.net:8443/view/OnionShare/job/Test_onionshare_unit_tests_develop_branch/23/console

I'm sure I'll think of more to add over time, I guess I can make more PRs for those when they come about

### DIFF
--- a/unit_tests/onionshare_autostop_timer_test.py
+++ b/unit_tests/onionshare_autostop_timer_test.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+import os, sys, unittest, socket, pytest, json
+from PyQt5 import QtCore, QtWidgets, QtTest
+
+from onionshare import onion, strings, common
+from onionshare_gui import *
+
+app = QtWidgets.QApplication(sys.argv)
+
+class OnionShareGuiTest(unittest.TestCase):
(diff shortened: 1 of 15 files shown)


micahflee:
Nice! I'm excited about these.

How do you run them? Can you add instructions to `BUILD.md`? `python3 -m unittest unit_tests/` doesn't seem right, and neither does `pytest-3 unit_tests/`.

mig5:
I run them in 'headless' mode on Debian 8 like so:

```
xvfb-run py.test-3 -v unit_tests/foobar.py
```

Where foobar.py is the relevant test. I think it might work if you just point it at the folder too, but I haven't tried.

Do you want to see if that works for you, and if so, I'll update the BUILD.md. I think I mentioned before but I had trouble running these on Fedora 26 in Qubes, but they run fine in 'headless' mode on Debian as above.

mig5:
Don't merge this, as the recent downloads window changes broke some of the tests. I'll update when I get the chance.

mig5:
I'll make a new PR for the GUI tests when they're all refactored for receive-mode and v3 addresses etc.

PR #651, opened 2018-02-27
Move downloads into their own window that can be toggled on/off

mig5:
I had a crack at the planned work to move the downloads into their own window that can be toggled on or off.

This maybe could do with some prettying up, but it works.

First, we have the 'inactive' state where there are no downloads. The toggle button is grey.

download_window_inactive

Then we have some download(s) that have occurred. The toggle button lights up green.

download_window_active

And finally, when you toggle the window, this is what you see

download_window_open

If there are no downloads yet, you'll be told 'No downloads yet'.

Toggle the button again and the window will disappear.

There is also a toolTip on the toggle button 'Show/hide downloads'

On Mac and Windows, the 'close' button on the popup window will be hidden. That way the user needs to 'toggle' via the toggle button, to hide it again. The reason is that otherwise, if they could 'close' the window directly via its own menu bar, the toggle button would remain toggled ('shown'), and would have to be pressed several times to get back to 'normal' state. Micah, note that on Qubes I think the window title is *always* shown because of Qubes' security policy in 'painting' the window so that you know which domain you're in. So there's no completely avoiding it in that case. But it's a minor quirk.

Also, note that I used QToolButton deliberately (rather than QPushButton) because the Qt conventions say to use such a widget when the intention is a 'toggle on/off' action (which is why its default mode is to hold an icon only, as is the case here)


Baccount:
If there are a large amount of downloads, the window will go off the screen.
issue

mig5:
Oh thanks, I will make this scrollable again.

Baccount:
@mig5 After a share is finished and I start a new share, the download window is not resized.
bug

mig5:
Thanks for those pick-ups, they should now be fixed (window is scrollable again once it reaches a few downloads, and it resets back to that original size when you start a new share with 'No downloads yet')

micahflee:
Nice, this looks excellent. I've only tested in Qubes so far, and I'm about to test in Windows. But I've reviewed the code and it looks good. I'm adding one more commit myself that basically adds a stretch to the bottom of the downloads window, which forces all the progress bars to be stacked at the top of the window, which I think looks better.

Issue #652, opened 2018-02-27
Can we get real "close/minimize to tray" action?


ghostlands:
Right now (1.3) there isn't any way to get onionshare out of the window list bar. Even when minimized, it still takes up that space.

Can we do a minimize or close to tray? There may be times when onionshare is running for hours at a time, and it would be nice if the gui could stay out of the way in those cases.

The cli version also employs a window which cannot be gotten out of the way; cli isn't an alternative.

Tray notifications are nice, but I feel like the main reason tray presence is a thing is to conserve real estate and reduce clutter.

mig5:
A great idea.

trymeouteh:
I would like to see a close to tray option were the only way to close Onionshare is to click on the tray icon to close the app.

PR #654, opened 2018-02-28
Add an optional setting to minimise to system tray when closing window

mig5:
An initial pass at #652 

Note the 'close_from_systray()' which is a similar function to the closeEvent. This is because otherwise, when the 'minimize to system tray' option is set, clicking Quit from the systray's menu won't actually quit!

@ghostlands would be great if you are in a position to [test this](https://github.com/micahflee/onionshare/wiki/Developing-OnionShare#running-onionshare-without-making-a-package), let us know if you need help doing so.


ghostlands:
I'll go for it later tonight. Thanks as always for working on it.

ghostlands:
Weird, re: 'close_from_systray()'/not closing.

Tested and working. As-is, it will definitely do the job of keeping the window out of the way when there are long periods of uptime.

It's a bit awkward though, so I'll offer some UI suggestions in case you want to refine it.

There's always a debate about whether this kind of UI feature should do, from a user perspective, "close to tray" or "minimize to tray", I'm in the latter camp.

Close generally means quit. I think it's counterintuitive to close and have it do something else. Minimizing, is a "get out of the way" control, and the tray is that. So, if the feature followed this logic, the wording for the checkbox in settings would simply be "minimize to system tray".

It's also nice to be able to minimize to the tray *from* the tray, where left clicking on the icon while the window is open will then minimize it.

Lastly, it's debatable whether an app that lives in the tray should appear in the window list when the window is open. I lean in the direction of no, but this detail is less important..

mig5:
The GUI setting says "Minimize to system tray when window closed?", to me this makes sense as that's exactly what happens with most apps that offer such a setting - if you then click the X button in the window's toolbar, it will then 'minimize' (whatever you want to call it.. really the window is simply hidden) to the system tray.

I can't work out if what you are asking for is that to occur when you click the '_' (minimize) button in the window's toolbar as well/instead...

I'm adding the menu action now - ironically I had that in, then I took it out before making the PR.

ghostlands:
The menu action is nicer. There's some weird clicky behaviour where when you right click on the icon in the tray when minimized, instead of the menu popping up, the whole window pops up. Seems like the whole window should only open on left click.

>  with most apps that offer such a setting

Some do this, true. QPS does this, for instance, so there's precedent certainly. Still doesn't make intuitive sense to me; close means quit in most windowed applications. This really isn't important enough to be a blocker though.

> to occur when you click the '_' (minimize) button in the window's toolbar [...]instead

This. Yeah, not all min-to-tray apps follow the same logic. Qps, for instance, uses a mixture of what we've been discussing. It only reduces to tray when the close button is ticked, but once there can be maximized and "trayed" with single left clicks. And then when maximised it shows the window in the window list.

So there's three details typically:

- window list when maximised, or no?
- to tray on minimise (-) or close (x)?
- steps to show window once in tray: single left click, or menu action?

To me, the most intuitive is that the window acts like any normal window, _unless_ the "min to tray" option is checked, in which case, it:

- does not appear in the window list
- the window can be "shown" and "trayed" with single left clicks
- is "trayed" with the minimize (-) window toggle (not the close (x) window toggle).

mig5:
Cool, I couldn't get the minimize to work in Qt (the alleged QEvent that is meant to fire (QEvent.WindowStateChange), does not for me (only QEvent.ActivationChange does), so I'm gonna give up on this for now. Hopefully someone else can take it further.

PR #655, opened 2018-02-28
Move cmdline parsing to separate function

delirious-lettuce:
This is just a little experiment to try and break up some of these long functions into smaller functions (starting with command-line parsing). Although I think there could be a few more functions pulled out of `main`, I wanted to see what the reaction would be to this smaller PR first.

* Replaced the `*` import to be more explicit
    ```python
    from .onion import *

    # VS

    from .onion import (
        BundledTorNotSupported,
        BundledTorTimeout,
        Onion,
        TorErrorAuthError,
        TorErrorAutomatic,
        TorErrorInvalidSetting,
        TorErrorMissingPassword,
        TorErrorProtocolError,
        TorErrorSocketFile,
        TorErrorSocketPort,
        TorErrorUnreadableCookieFile,
        TorTooOld
    )
    ```
    [Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools.
](http://pep8.org/#imports)
* Remove `dest` parameters since they were only returning the defaults anyways

    ```python
    >>> from argparse import ArgumentParser
    >>> parser = ArgumentParser()
    >>> parser.add_argument('--local-only')  # dest='local_only' unnecessary
    _StoreAction(option_strings=['--local_only'], dest='local_only', nargs=None, const=None, default=None, type=None, choices=None, help=None, metavar=None)
                                                  ^^^^^^^^^^^^^^^^^
    ```

    [For optional argument actions, the value of `dest` is normally inferred from the option strings. `ArgumentParser` generates the value of `dest` by taking the first long option string and stripping away the initial -- string. If no long option strings were supplied, `dest` will be derived from the first short option string by stripping the initial - character. Any internal - characters will be converted to _ characters to make sure the string is a valid attribute name.](https://docs.python.org/3/library/argparse.html#dest)

* Use `type=int` to ensure that `--shutdown-timeout` is given as an integer. Currently, a string could be given which would raise a `ValueError` on this line: 
    ```python
    shutdown_timeout = int(args.shutdown_timeout)
    ```
    Not a big problem but using `type=int` will raise this clear error message as soon as a non-integer is passed in.

    ```python
    >>> from argparse import ArgumentParser
    >>> parser = ArgumentParser()
    >>> parser.add_argument('--shutdown-timeout', default=0, type=int)
    _StoreAction(option_strings=['--shutdown-timeout'], dest='shutdown_timeout', nargs=None, const=None, default=0, type=<class 'int'>, choices=None, help=None, metavar=None)

    >>> parser.parse_args(['--shutdown-timeout', 'NON_INTEGER'])
    usage: pydevconsole.py [-h] [--shutdown-timeout SHUTDOWN_TIMEOUT]
    pydevconsole.py: error: argument --shutdown-timeout: invalid int value: 'NON_INTEGER'
    ```

* Use `type=FileType('r')` for both `--config` and `filename` to ensure only readable files can be passed in. Letting `argparse` take care of this allows the "Validation" code to be removed.

    ```python
    >>> from argparse import ArgumentParser, FileType
    >>> parser = ArgumentParser()
    >>> parser.add_argument('filename', type=FileType('r'))
    _StoreAction(option_strings=[], dest='filename', nargs=None, const=None, default=None, type=FileType('r'), choices=None, help=None, metavar=None)

    >>> parser.parse_args(['README.md'])
    Namespace(filename=<_io.TextIOWrapper name='README.md' mode='r' encoding='UTF-8'>)

    >>> parser.parse_args(['no_file'])
    usage: pydevconsole.py [-h] filename
    pydevconsole.py: error: argument filename: can't open 'no_file': [Errno 2] No such file or directory: 'no_file'

    >>> parser.parse_args(['test'])
    usage: pydevconsole.py [-h] filename
    pydevconsole.py: error: argument filename: can't open 'test': [Errno 21] Is a directory: 'test'

    ```
* Alphabetize and format exceptions for readability. I know it adds a few lines but it just seemed a bit better than having to scroll far to the right to see all the exceptions.

    ```python
    except (TorTooOld, TorErrorInvalidSetting, TorErrorAutomatic, TorErrorSocketPort, TorErrorSocketFile, TorErrorMissingPassword, TorErrorUnreadableCookieFile, TorErrorAuthError, TorErrorProtocolError, BundledTorNotSupported, BundledTorTimeout) as e:

    # VS

    except (BundledTorNotSupported,
            BundledTorTimeout,
            TorErrorAuthError,
            TorErrorAutomatic,
            TorErrorInvalidSetting,
            TorErrorMissingPassword,
            TorErrorProtocolError,
            TorErrorSocketFile,
            TorErrorSocketPort,
            TorErrorUnreadableCookieFile,
            TorTooOld) as e:
    ```

* This section of code was almost entirely redundant since most of the arguments were already the types they were being cast into here.

    ```python
    local_only = bool(args.local_only)  # already a boolean
    debug = bool(args.debug)  # already a boolean
    stay_open = bool(args.stay_open)  # already a boolean
    shutdown_timeout = int(args.shutdown_timeout)  # error if non-integer, fixed by `type=int` above
    stealth = bool(args.stealth)   # already a boolean
    config = args.config
    ```

Anyways, let me know what you guys think!

### DIFF
--- a/onionshare/__init__.py
+++ b/onionshare/__init__.py
@@ -18,11 +18,11 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 
-import argparse
 import os
 import sys
 import threading
 import time
+from argparse import ArgumentParser, FileType, HelpFormatter
 
 from . import strings, common, web
 from .onion import (
@@ -43,10 +43,37 @@ from .onionshare import OnionShare
 from .settings import Settings
 
 
+def _parse_cmdline_args():
+    parser = ArgumentParser(
+        formatter_class=lambda prog: HelpFormatter(prog, max_help_position=28))
+    parser.add_argument(
+        '--local-only', action='store_true',
+        help=strings._("help_local_only"))
+    parser.add_argument(
+        '--stay-open', action='store_true',
+        help=strings._("help_stay_open"))
+    parser.add_argument(
+        '--shutdown-timeout', metavar='<int>', default=0, type=int,
+        help=strings._("help_shutdown_timeout"))
+    parser.add_argument(
+        '--stealth', action='store_true',
+        help=strings._("help_stealth"))
+    parser.add_argument(
+        '--debug', action='store_true',
+        help=strings._("help_debug"))
+    parser.add_argument(
+        '--config', metavar='config', default=False, type=FileType('r'),
+        help=strings._('help_config'))
+    parser.add_argument(
+        'filename', metavar='filename', nargs='+', type=FileType('r'),
+        help=strings._('help_filename'))
+    return parser.parse_args()
+
+
 def main(cwd=None):
     """
-    The main() function implements all of the logic that the command-line version of
-    onionshare uses.
+    The main() function implements all of the logic that the
+    command-line version of onionshare uses.
     """
     strings.load_strings(common)
     print(strings._('version_string').format(common.get_version()))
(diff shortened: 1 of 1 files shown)


delirious-lettuce:
Just realized this won't take directories from cmdline, back to the drawing board!

PR #664, opened 2018-03-07
#662 Allow local-only mode in GUI

### LINKED ISSUE #662: With --local-only flag, don't try connecting to Tor
I've been using the `--local-only` flag more often while developing, but the Onion object still tries connecting to Tor. Instead, when `--local-only` is set, Onion should basically just stub out connecting to Tor, as if it was successful.

mig5:
Let me know if this fixes #662 for you @micahflee ?

I'm basically pulling the local_only flag into various elements of the GUI so that various things get ignored (don't connect to Tor, don't 'reboot' Onion when settings change in SettingsDialog, don't alert when Tor is disconnected/authenticated, etc)

### DIFF
--- a/onionshare_gui/__init__.py
+++ b/onionshare_gui/__init__.py
@@ -108,7 +108,7 @@ def main():
     app = OnionShare(onion, local_only, stay_open, shutdown_timeout)
 
     # Launch the gui
-    gui = OnionShareGui(onion, qtapp, app, filenames, config)
+    gui = OnionShareGui(onion, qtapp, app, filenames, config, local_only)
 
     # Clean up when app quits
     def shutdown():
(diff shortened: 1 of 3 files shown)


Baccount:
I get these messages when sharing a file with the local_only flag.

127.0.0.1 - - [06/Mar/2018 23:33:45] "GET /qla2q2swijs7m4xfojk2qrv2ce/shutdown HTTP/1.1" 200 -
 * Running on http://127.0.0.1:17609/ (Press CTRL+C to quit)

mig5:
That’s normal, and occurs for non-local mode too, it’s just the shutdown sequence sent to the Flask webserver .

(Local only mode still starts the web server, it just doesn’t connect a Tor onion service to it)

micahflee:
Nice, this totally fixes #662

PR #667, opened 2018-03-08
Remove the zip progress bar on stop/cancel, if it exists

### LINKED ISSUE #665: If I stop a share when the file's are being compressed the "compressing progress bar" is not cleared
If I stop a share when the file's are being compressed, the "compressing progress bar" is not cleared.

1. Add a very large amount of large files.
2. start a share and when the progress bar is at 0% stop the share/cancel it.
3. the progress bar is not cleared 
bug

Video showing the bug.
https://www.icloud.com/iclouddrive/0kW7nYDj0u8ZBEW_lTJB2dseQ#bug
@mig5

mig5:
Fixes #665

### DIFF
--- a/onionshare_gui/onionshare_gui.py
+++ b/onionshare_gui/onionshare_gui.py
@@ -511,6 +511,10 @@ class OnionShareGui(QtWidgets.QMainWindow):
         # Remove ephemeral service, but don't disconnect from Tor
         self.onion.cleanup(stop_tor=False)
         self.filesize_warning.hide()
+        # Remove zip progress bar
+        if self._zip_progress_bar is not None:
+            self.status_bar.removeWidget(self._zip_progress_bar)
+            self._zip_progress_bar = None
         self.downloads_in_progress = 0
         self.downloads_completed = 0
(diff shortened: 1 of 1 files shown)


mig5:
It looks like it's because `start_server_step2()`'s function for updating the zip progress bar etc (`finish_starting_server()`) is actually in its own thread. 

If you recall, it's hard to short-circuit a normal Thread, but it's possible to do so with QThread. (We did so in order to introduce 'canceling' a share mid-startup).

Perhaps this function should also use our overridden OnionThread, and that way we can call its quit() function in `stop_server()`. What do you think? I'm trying something like that as I type.

mig5:
@micahflee mad props if you can figure out a way to fix this. I've had a few stabs at it, but I can't find a way to short-circuit that thread that is responsible for updating the zip progress bar, from the parent, on cancellation.

See also #706 which I think is a thread-related crash relating to canceling a share mid-startup. Makes me wonder if that 'cancel' feature causes more harm than good, but I would find it hard to remove it as it's a nice feature..

micahflee:
Closing in favor of #758

Issue #647, opened 2018-02-26
meek_lite not supported in the obfs4proxy binary shipped with Tor Browser


Baccount:
I get stuck at connecting to directory server when using amazon meek lite.
`Connecting to the Tor network: 5% - Connecting to directory server`
tested using the latest develop build.

mig5:
It works for me on Linux, but I noticed this on Windows too yesterday.

I'm not sure it's OnionShare..

mig5:
Yeah I reproduce it on macOS too :/

mig5:
Ah.

I ran tor from CLI with a custom torrc using the same meek lite settings and I saw this Tor message

```
Client managed proxy encountered a method error. (meek_lite no such transport is supported)
```

So indeed, I guess the obfs4proxy binary that ships with Tor Browser doesn't support meek :/ what to do?

Baccount:
meek-client and meek-client-torbrowser are both included in Tor browser. TorBrowser.app/Contents/MacOS/Tor/PluggableTransports

mig5:
I will try and do some tests but my concern is whether those binaries support meek_lite (not just main meek)

mig5:
Yeah similar error with meek-client (meek_lite no such method')

micahflee:
Confirmed, I'm not able to connect to meek_lite Amazon in macOS.

mig5:
meek_lite was introduced to obfs4proxy in v.0.0.6, back in 2016: https://github.com/Yawning/obfs4/blob/obfs4proxy-0.0.6/ChangeLog

On Fedora I have v0.0.7 so that's why it's working.

The version bundled with Tor Browser on Mac (and probably same on Windows) is for some reason super old! v0.0.5, so it predates the meek_lite feature.

mig5:
Created https://trac.torproject.org/projects/tor/ticket/25356#ticket

micahflee:
Marking this blocked until Tor Browser starts shipping more up-to-date obfs4proxy binaries.

mig5:
The updated binary should arrive in the 'alpha' Tor Browser channel in about 2 weeks. We can always choose to download both binaries and extract the obfs4proxy from alpha, tor binary from stable. Or we can just wait til it lands in stable, but that might take a lot longer.

Baccount:
@mig5 The latest TorBorwser nightly should include obfs4proxy in v0.0.7 
[...]

PR #670, opened 2018-03-13
Update Tor to 0.3.2.10

Baccount:
The latest version of TorBrowser adds meek-lite v0.0.7. After this is merged  #647 would no longer be blocked on macOS and windows.

### DIFF
--- a/install/get-tor-windows.py
+++ b/install/get-tor-windows.py
@@ -28,9 +28,9 @@ import inspect, os, sys, hashlib, shutil, subprocess
 import urllib.request
 
 def main():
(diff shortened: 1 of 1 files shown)


micahflee:
Thanks! Since I like shipping the latest version of Tor with each release, I'm going to wait until we're closer to another release before merging this, just in case there's an even newer version of Tor available at that time.

(And of course, #647 still needs a separate PR to support it, because the meek options are hidden in Windows and Mac right now. But stopping hiding them, and also of course testing to confirm that it actually works under meek, should be a simple PR.)

Baccount:
@micahflee Tor 0.3.3.5-rc was just released, so the final version of 0.3.3 should be out within a few weeks.

micahflee:
I've duplicated this PR in #710, so I'm closing this one

PR #676, opened 2018-03-22
Spelling: marked nouns, ellipsis, +,  set up, addresses, could not, expire

comradekingu:
Possibly one-time links "expire" rather than "stop".
"Torified" I am not quite sure about.


share/locale/en.json
```diff
@@ -106,7 +106,7 @@
     "gui_settings_tor_bridges_meek_lite_azure_radio_option_no_obfs4proxy": "Use built-in meek_lite (Azure) pluggable transports (requires obfs4proxy)",
     "gui_settings_tor_bridges_custom_radio_option": "Use custom bridges",
     "gui_settings_tor_bridges_custom_label": "You can get bridges from <a href=\"https://bridges.torproject.org/options\">https://bridges.torproject.org</a>",
-    "gui_settings_tor_bridges_invalid": "None of the bridges you supplied seem to be valid.\nPlease try again with valid bridges.",
+    "gui_settings_tor_bridges_invalid": "None of the bridges you supplied seem work.\nPlease try again by double-checking them or adding other ones.",
```
Baccount: None of the bridges you supplied seem **to** work.

share/locale/en.json
```diff
@@ -119,35 +119,35 @@
     "settings_error_socket_file": "Can't connect to Tor controller using socket file {}.",
     "settings_error_auth": "Connected to {}:{}, but can't authenticate. Maybe this isn't a Tor controller?",
     "settings_error_missing_password": "Connected to Tor controller, but it requires a password to authenticate.",
-    "settings_error_unreadable_cookie_file": "Connected to Tor controller, but can't authenticate because your password may be wrong, and your user doesn't have permission to read the cookie file.",
-    "settings_error_bundled_tor_not_supported": "Bundled Tor is not supported when not using developer mode in Windows or macOS.",
-    "settings_error_bundled_tor_timeout": "Connecting to Tor is taking too long. Maybe your computer is offline, or your clock isn't accurate.",
+    "settings_error_unreadable_cookie_file": "Connected to Tor controller, but can't authenticate because your password may be wrong, and your user lacks permission to read the cookie file.",
+    "settings_error_bundled_tor_not_supported": "Use of the Tor version bundled with OnionShare is not supported when not using developer mode on Windows or macOS.",
```
Baccount: `Use of the Tor version bundled with OnionShare`

`Bundled Tor is not supported when not using developer mode` sounds better.

comradekingu: @Baccount It begs the question of "Bundled Tor" is.

share/locale/en.json
comradekingu: Good catch

share/locale/en.json
Baccount: Bundled Tor is the Pre compelled version of tor that is inside the .app

comradekingu: This may not be not all that clear for the user, whereas if it says it in cleartext, no confusion will arise.
That is the intent anyhow.

share/locale/en.json
```diff
@@ -26,10 +26,10 @@
     "systray_download_completed_message": "The user finished downloading your files",
     "systray_download_canceled_title": "OnionShare Download Canceled",
     "systray_download_canceled_message": "The user canceled the download",
-    "help_local_only": "Do not attempt to use tor: for development only",
+    "help_local_only": "Do not attempt to use Tor: For development only",
     "help_stay_open": "Keep onion service running after download has finished",
     "help_shutdown_timeout": "Shut down the onion service after N seconds",
-    "help_transparent_torification": "My system is transparently torified",
+    "help_transparent_torification": "My system is transparently Torified",
```
micahflee: Actually, the whole string `help_transparent_torification` is a relic from the past and isn't in use in the app anymore. This string can be removed completely. I just opened an issue to remove all such strings #681, but in the meantime you can just delete this one in this PR.

share/locale/en.json
micahflee: I'm fine with `Use of the Tor version bundled with OnionShare`, because that's the same language used in ` "gui_settings_connection_type_bundled_option": "Use the Tor version that is bundled with OnionShare"`. Also, this string is only ever displayed to people who run `./dev_scripts/onionshare-gui`, and not to people who run OnionShare from the installer, and I think developers are likely to understand it either way.

But, I realized there's a typo. Bundled Tor isn't supported when *using* developer mode, not when *not using* developer mode. So the string should be this instead: `Use of the Tor version bundled with OnionShare is not supported when using developer mode on Windows or macOS.`

PR #710, opened 2018-06-18
Version 1.3.1, include 3rd party licenses, update tor

### LINKED ISSUE #709: Comply with the obfs4proxy license.
obfs4proxy and associated code is distributed under a two clause BSD style license, which imposes requirements on binary re-distribution.

>  * Redistributions in binary form must reproduce the above copyright notice,
>   this list of conditions and the following disclaimer in the documentation
>   and/or other materials provided with the distribution.

The Windows and OSX binary packages are not in compliance with the requirements.  Please remove the infringing binary packages, and either cease re-distributing the obfs4proxy binary, or do so according to the terms of the license.

micahflee:
This fixes #709.

The Mac version now includes `/Applications/OnionShare.app/Contents/Resources/licenses/` and the Windows version includes `C:\Program Files (x86)\OnionShare\licenses\`, with the software licenses for OnionShare, Tor, and obfs4. You can see the licenses [here](https://github.com/micahflee/onionshare/tree/license/install/licenses).

It also reimplements #670, so that PR can be closed after this is merged into develop as well.

I also ran into a new issue related to the `get-tor-windows.py` script. In Windows it turns out downloading from https://archive.torproject.org/ using `urllib.request` throws an SSL verification error. My guess is it's because `urllib` uses the Windows certificate store, and some how it's not including the full Let's Encrypt chain. So I fixed it by adding a build dependency of `requests`, and using `requests` to make the download, which works fine without a certificate error.

Because this is a point release, this PR will get merged directly into master. So it also includes the version bump to 1.3.1 and an updated changelog.


mig5:
@micahflee I am having an issue with Persistent URLs on Mac OS on this branch.

If I go and set Persistent Mode on in settings, then start a share, only the slug is saving to the json file, not the private key.

Not sure yet if this is an issue with the new Tor binary or not.. though hmm, it can't be, because I'm using the onionshare dev scripts, meaning I'm using Tor Browser as my connection.. may have just stumbled across an unrelated bug? Do you reproduce?

mig5:
@micahflee interesting: I can only reproduce the persistent URL issue when using the Tor Browser connection method *and* using the dev_scripts/ script.

Persistent URLs work fine for me on OS X and on Windows when using the bundled Tor, and when using the Tor Browser method when opening the app 'properly' from a compiled binary e.g not via dev_scripts/

Seems like an edge case, so I think this is fine to go through... curious if you reproduce the above though (dev_scripts/ and bundled tor mode, with persistent URLs enabled, no private_key setting saved to json after starting share).

Note also the conflict in install/pyinstaller.spec in this PR that will need fixing up to merge.

LICENSE
```diff
@@ -1,6 +1,6 @@
 OnionShare
 
-Copyright © 2016
+Copyright © 2018
```
mig5: @micahflee Can I recommend we add a single line under the Copyright here saying 'Note: further licenses relating to Onionshare's use of third party software can be found under install/licenses/'

This makes it clear at the top level that there are other licenses deeper in the source code.

micahflee: Good idea!

PR #716, opened 2018-07-04
suggesting standard checks for *BSD systems

egypcio:
Hi,

for some 'if' checks I noticed that we might use the ".endswith('BSD')" function in order to verify if a system's platform is a BSD system or not.

other particular checks still enforce only 'BSD' as platform, and I preferred to kept them as is; once particular case is https://github.com/torbsd/onionshare/blob/egypcio/onionshare/common.py#L57 

I hope this brings something good for OnionShare and later on we can port it to even more BSD systems without the need of extra checks. the OpenBSD package is already available, and the FreeBSD one is almost ready to land; https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=225539

if any other particular needs are required to this merge, please tell me. would be happy to help!

thank you for all your work and concern! KR,

### DIFF
--- a/onionshare/common.py
+++ b/onionshare/common.py
@@ -57,7 +57,7 @@ def get_platform():
     Returns the platform OnionShare is running on.
     """
     plat = platform.system()
-    if plat.endswith('BSD'):
+    if plat.endswith('BSD') or plat == 'DragonFly':
         plat = 'BSD'
     return plat
 
--- a/onionshare/onion.py
+++ b/onionshare/onion.py
@@ -334,7 +334,7 @@ class Onion(object):
             # guessing the socket file name next
             if not found_tor:
                 try:
-                    if self.system == 'Linux' or self.system == 'BSD':
+                    if self.system == 'Linux' or platform.system().endswith('BSD'):
                         socket_file_path = '/run/user/{}/Tor/control.socket'.format(os.geteuid())
                     elif self.system == 'Darwin':
                         socket_file_path = '/run/user/{}/Tor/control.socket'.format(os.geteuid())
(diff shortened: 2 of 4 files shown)


jbeich:
Bitrig (discontinued) and DragonFly don't end system name with `BSD`. DragonFly uses [an overlay](https://github.com/DragonFlyBSD/DeltaPorts) for FreeBSD Ports, so downstream packages are almost identical but maybe out of date due to lack of manpower to sync.

egypcio:
Hi @jbeich ,
thanks for writing your words here. very appreciated.

sadly Bitrig is discontinued, so this one should not be a problem here. is [TrueOS](https://trueos.org) also "affected", or does it return FreeBSD as platform?

I thought that DragonFly was also covered by those patches, but its return of `platform.system()` does not end with 'BSD' and `uname` works the same way; I did see it as 'DragonFlyBSD'.

do you have any suggestions that we might try to implement for getting DragonFly included here? I opted for this approach because of the embracing of FreeBSD, NetBSD and OpenBSD (+GhostBSD, ElectroBSD, MidnightBSD and RusBSD). I can try to work on [get_platform()](https://github.com/micahflee/onionshare/blob/master/onionshare/common.py#L55) if that makes more people happy.

btw, are you trying to use OnionShare on DragonFly? did you get the [patch](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=225539) on FreeeBSD's Bugzilla and merges any stuff into DPorts to get its package available?

micahflee:
Sorry, I'm just now looking at this. I just changed this PR to merge into `develop` instead of `master`, and a lot has changed in the `develop` branch since you first started working on this, so there's a big merge conflict that needs resolving.

PR #724, opened 2018-07-22
[receiver-mode-gui] remove obsolete strings

### LINKED ISSUE #681: Remove obsolete strings
I've noticed that some of the strings including in the locale files, such as `help_transparent_torification`, are obsolete and no longer referenced in the source code. All of the obsolete strings should be removed from the locales of all languages.

mig5:
Fixes #681 

I used `python install/check_lacked_trans.py | grep disused` for this.

I also based it on receiver-mode-gui since that kind of has the most recent addition of strings due to all the refactoring. But we should also do another sweep of this on develop or master branch prior to the next release.


mig5:
Fixed the check_lacked_trans.py to use re.findall() instead of re.search() (so it can find more than the first match on a single line), and some other improvements. Re-ran and removed what should *truly* be disused strings now...

micahflee:
Nice. Yeah, I agree we should do this again right before the release, using your nicely fixed up `check_lacked_trans.py`.

PR #722, opened 2018-07-21
Expand 'public mode' (optional slugs) to be possible for sharing too

mig5:
Took a pass at #713 - this expands the 'public mode' of Receive to be possible for Share too, in which case there's no slug and no 'DoS' risk of invalid slugs.

For this reason, I based this work off of the receiver-mode-gui branch rather than develop branch, since Receive Mode already had this feature.

To represent this more clearly in the SettingsDialog, I added a 'General Settings' and moved the Public Mode into it, out of receive mode. I also noted that the 'persistent URL' mode should not be in 'Share mode' settings since it presumably can be a persistent URL for receive mode too, so moved that as well.

Appreciate plenty of testing on this sensitive idea.

Issue #713, opened 2018-06-24
Slugs should be optional in order not to allow DoSing of 'public' OnionShares


b-meson:
Right now the keep alive option is great but kind of defeats the purpose of a slug if you want to keep a .onion running for a long time. A malicious relay could hit the slug and DOS the server. Really a public server doesn’t need a slug. This should be easier to implement in send mode.

You may also want to do a similar feature for receive mode.

Baccount:
@b-meson When ticket #677 is finished users will be able to use V3 Onions, V3 onions are Not vulnerable to Onion Service discovery attacks. 

>  For example, did you know that in the legacy onion system, the network could [learn about your onions?](https://lists.torproject.org/pipermail/tor-talk/2017-March/042986.html) However, with this next-generation design, your onion address is completely private and only known to you and whoever you choose to disclose it to.


More information on V3 onions 
https://blog.torproject.org/tors-fall-harvest-next-generation-onion-services
https://trac.torproject.org/projects/tor/wiki/doc/NextGenOnions

mig5:
I agree with this - it is not so much about discovery attacks as it is about a user's conscious choice to perhaps publicise their OnionShare URL for public access. Any malicious actor could immediately bring down the share by deliberately hitting 20 false slugs. Regardless of v3 we should definitely make the slug optional.

mig5:
A PR at #722 - note it's based on the receiver-mode-gui branch, because that branch already had this feature for Receive Mode. Just expanding it to support Share mode.

micahflee:
This is implemented in the `receiver-mode-gui` branch.

Issue #594, opened 2018-02-08
Should slugs be made optional for the next generation Onion URL's?


Baccount:
@micahflee Do you think the the slugs should be made optional for next generation onion URL's?  Here is a excerpt from [Next-Generation Hidden Services Design document](https://gitweb.torproject.org/torspec.git/tree/proposals/224-rend-spec-ng.txt#n1115)
```
2.5.1. First layer of encryption [HS-DESC-FIRST-LAYER]
   The first layer of HS descriptor encryption is designed to protect
   descriptor confidentiality against entities who don't know the blinded
   public key of the hidden service.
```

@mig5 commented on this in https://github.com/micahflee/onionshare/issues/461#issuecomment-358854192
` I am aware of the lack of need for slug (I would argue it could still be useful though in case some future vulnerability is discovered in v3 services that allowed discovery of the address - defense in depth)`

mig5:
This will be fixed by #722 which expands 'public mode' for the Share feature (not just Receive feature), per #713 

I think it would be ok to have the slug be able to be turned on or off even for v3 (no need to *force* the slug to be off when using v3 onions - let's leave it up to the user).

micahflee:
Slugs will indeed be made optional with v3 onion services, but here's an argument for keeping them by default.

I shared some files and ended up with this OnionShare address: `http://fzrmu6tifi6byvwhwdop7isprka6nffgw4sdpx2564wtnwkc7q5l2qad.onion/backlog-brisket`

Then I opened wireshark to sniff traffic, and tried loading that address in Chrome. Here's a screenshot of the traffic dump (github won't let me upload the .pcapng file directly to this issue, but this is just as good):

![screenshot_2018-09-17_09-42-16](https://user-images.githubusercontent.com/156128/45636930-0ea22680-ba5e-11e8-85db-8bb3e604c928.png)

Chrome makes a DNS request (plaintext, for anyone on the local network or anywhere upstream to spy on) for `fzrmu6tifi6byvwhwdop7isprka6nffgw4sdpx2564wtnwkc7q5l2qad.onion`, but since it fails to resolve a domain, it just doesn't work, and results in an error:

![screenshot_2018-09-17_09-44-10](https://user-images.githubusercontent.com/156128/45637029-4ad58700-ba5e-11e8-83bb-8843adc0f9a1.png)

However the slug, `backlog-brisket`, is nowhere in the traffic dump, since the HTTP request never actually gets sent.

[...]

PR #731, opened 2018-08-21
Introduce v3 onion support

mig5:
This PR introduces v3 onion support. I'm deliberately basing this on the receiver-mode-gui branch because it has so many other breaking changes that it's not worth branching off of develop branch anymore.

This PR:

1) favours v3 onions if the Tor version is new enough

2) Requires the (testing) user to `pip install pysha3` if they don't have it already (I encountered an issue on Debian 9 whereby, if the version of Python is too low, the logic that uses hashlib doesn't quite work (`hashlib.sha3_256`) , but pysha3 does. However, there's no deb or rpm for it, so it introduces the need to pip install at least one package on those platforms (for the first time) if building/testing from source. Or, more generally, introduces that dependency. Not sure how that will fly for Debian/Ubuntu .debs....

3) Deliberately hides the 'persistent URL' and 'stealth' features in the settings dialog (since we know those don't work properly with v3 onions) 

4) even if somehow persistent or stealth mode is enabled (think: previously-saved .json file, or command line flags), they are silently ignored (resulting in a non-persistent onion with no stealth)

Otherwise it contains the logic from my previous prop224 branch. Note the onionkey.py, which *pre-generates* private keys for both v3 and v2 now. This is necessary to interact with Stem with v3 keys, but even more significantly, it paves the way for saving that v3 private key to json once persistence is possible, *and* it paves the way for a 'dead man's switch' feature (reverse shutdown-timer: a *start-up* timer that allows the user to communicate a *future* onionshare URL yet to be published), which I will send a PR for later down the track.

I'm assuming it's not quite ready yet.

the v3 onion works for both share and receiver mode.

Welcome testing.

Once the persistence feature is released in Tor stable (expected in December), we can remove the logic that ignores-or-hides persistence mode. Likewise for stealth (Client Auth)

@Baccount


mig5:
Modified this PR so that rather than hiding the entire persistence/stealth settings if the Tor version is high enough, the user can optionally 'Use legacy v2 onion' mode via a checkbox in the Settings Dialog.

When this setting is checked on, persistence and stealth options are shown.

When persistence or stealth options are enabled, the legacy mode cannot be switched off (because those settings currently only work with v2 mode).

When persistence or stealth is disabled, the user can uncheck 'use legacy v2 mode' again, and go back to using v3 onions by default (on the assumption their Tor version is new enough for v3 in the first place).

mig5:
P.S the 'use legacy v2 onion' mode checkbox doesn't really belong in 'Sharing' section of the Settings dialog, since it affects receiver mode too. So really #722 would need to be merged first, then there'll probably be conflicts here needing adjustment.

Baccount:
When I open the settings panel, the app crashes every time.
```
  File "/Users/jackimoore/Downloads/onionshare-prop224_receiver_mode/onionshare_gui/onionshare_gui.py", line 318, in open_settings
    d = SettingsDialog(self.common, self.onion, self.qtapp, self.config, self.local_only)
  File "/Users/jackimoore/Downloads/onionshare-prop224_receiver_mode/onionshare_gui/settings_dialog.py", line 71, in __init__
    self.use_legacy_v2_onions_checkbox.setText(strings._("gui_use_legacy_v2_onions_checkbox", True))
  File "/Users/jackimoore/Downloads/onionshare-prop224_receiver_mode/onionshare/strings.py", line 61, in translated
    return strings[k]
KeyError: 'gui_use_legacy_v2_onions_checkbox'
Abort trap: 6
```
Here is the traceback.
@mig5

mig5:
D'oh, I was working across macihnes and I forgot to commit the share/locale/en.json.

Fixed..

asn-d6:
Hello, will test in the next 5-6 days max.

Baccount:
@mig5 When navigating to the V3 onion using TorBrowser 7.5.6, the Onion loads and looks like this.
screen shot 2018-08-22 at 1 59 39 pm

Baccount:
When I have onionshare set too connect to TorBrowser and I close TorBrowser and Onionshare, then open onionshare, it crashes every time. Here is the traceback.
```
[Aug 22 2018 14:04:15] SettingsDialog.__init__
  File "onionshare_gui/onionshare_gui.py", line 318, in open_settings
    d = SettingsDialog(self.common, self.onion, self.qtapp, self.config, self.local_only)
  File "onionshare_gui/settings_dialog.py", line 73, in __init__
    if Version(self.onion.tor_version) < Version('0.3.2.9'):
AttributeError: 'Onion' object has no attribute 'tor_version'
Abort trap: 6
```

Baccount:
@mig5 https://github.com/micahflee/onionshare/pull/731#issuecomment-415179747 crash still occurs even with #701

mig5:
@Baccount that looks like a valid bug when using Tor Browser as connection method and having no connection. I think that predates this bug e.g you could probably trigger that via the 'Update Checker' button too in latest version of OnionShare? Haven't tried, but I will look at whether we can workaround it later today. Thanks for finding

mig5:
the Tor version parsing depends on an already-connected Tor. I don't think I can work around that to parse the Tor version any other way. But obviously we need the SettingsDialog to open if there is no Tor connection (to allow users to change the connection method).

Maybe we should just show the 'use legacy v2 onions' checkbox regardless of the Tor version (e.g even if the Tor version can *only* support legacy v2). Because either way, the backend/CLI will ignore any attempt to use persistence/stealth regardless of what the json settings say, as a safety measure.

I'll push such a 'fix' for now, and state explicitly in the 'use legacy v2 onion' label that this is true even if unchecked, if the Tor version is too old.

mig5:
Pushed a fix that doesn't try and do any clever show/hide of features that are v2-only, based on Tor version.

Instead the labels show what is v2-only or not (for now).

If persistent mode or stealth are enabled, then the legacy v2 checkbox is forced on and read-only, since they are v2-only features.

Baccount:
Receive mode says "Don't prevent users from guessing onionshare address" does not seem appropriate for V3 onions, because the V3 design prevents the HSDIR from collecting Onion addresses , the crypto used is still unbroken ` (replaced SHA1/DH/RSA1024 with SHA3/ed25519/curve25519)` lastly, brute forcing a 56 characters long onion may be near impossible.
@mig5

mig5:
I guess I agree, but the feature really means 'don't use a slug', and I think that should still be optional for the end user.. however, 'Don't use a slug' isn't useful language for non-technical users which is why @micahflee used the language currently present.

I think I'll let Micah decide if slugs should be forcibly-disabled if v3 is in use, but again, there's no way for me to 'know' if v3 can be used right now because checking the Tor version can crash the app as you've already discovered.

I also think it's already 'fixed' by making the slug optional in share mode too, a PR for which I made in #722, maybe even there the language could be changed to 'Don't add a random path to the URL' or something that means more than 'slug' to less tech-savvy users...

Anyway for now let's keep this PR purely about whether v3 works and whether forcing v2 with/without persistence/stealth mode also still works :) we have tickets about the slug stuff in #594 and #713

mig5:
Thanks @asn-d6 - we encounter some difficulties in PyQt5 trying to auto-resize the parent widgets according to the size of the child widgets in some cases.

I initially experimented with elided text which expanded as the parent window was resized, but then encountered PyQt5 stylesheet issues with the text which I couldn't resolve. Followed @micahflee's out-of-band advice to just force a wrap on the string, but you're right, I didn't think about when the window is stretched..

Curious, did you stretch the window manually or did you somehow encounter it auto-expanding like that?

We can probably do a bit better at this anyway before merging, indeed..

asn-d6:
> Curious, did you stretch the window manually or did you somehow encounter it auto-expanding like that?

I did not stretch the window manually. I use a tiling WM (i3) and it just started up auto-stretched. :face_with_head_bandage:

mig5:
Thanks @asn-d6 ! I totally forgot about tiling WMs!

I've pushed a fix that only wraps it if the window size is below a certain size where QT's WordWrap can't sanely wrap it properly.

asn-d6:
Works great, @mig5. Thanks!

micahflee:
Also about pysha3, it's packaged in Debian testing (buster):
https://packages.debian.org/buster/python3-sha3

And it's also in Ubuntu 18.04 (bionic), which is the new LTS so everyone should upgrade:
https://packages.ubuntu.com/bionic/python3-sha3

But also, I'm actually ok with not supporting Debian stable for OnionShare 2.0, just because software in stable is always very old, and it's easy enough to use testing if you're interested in using newer software. Also, we can avoid having any pip dependencies this way.

mig5:
For Fedora I can't find an equivalent python3-sha3 package, at least not in Fedora 27, maybe later? Finding it harder to search than in debian repo pages.

micahflee:
Sorry, I just merged develop into receiver-mode-gui. That's why you have conflicts now. (Once #722 is merged, I think it's about time to merge receiver-mode-gui into develop, and then we can just start working off develop again...)

Yeah I also can't find pysha3 in Fedora 27 just by searching their package websites.

I tested this branch in Ubuntu 18.04, and it worked fine without having the python3-sha3 package instead, I guess because the version of python3 was new enough? I don't have a F27 VM handily available right now -- does it also have a new enough version of python3 so that we don't need the pysha3 dependency at all?

mig5:
I'll install this branch quickly on a Fedora DispVM and test that.

And I'll fix these new conflicts here after #722 is merged (at which point I expect more conflicts anyway).

mig5:
I confirm Fedora 27 doesn't need pysha3 installed (Python version is high enough).

The catch-22 is that Fedora 27's Tor binary is too old for v3 anyway, lol. I had to install Tor Browser and use that as my Tor connection method in order to even test v3 style onions on Fedora 27.

micahflee:
Haha, well ok then. I'm fine with basically requiring Debian testing, Ubuntu 18.04, or Fedora 28 for OnionShare 2.0 then. Do you want to remove the references to pysha3? We'll have to remember to update the [Linux Distribution Support](https://github.com/micahflee/onionshare/wiki/Linux-Distribution-Support) wiki page (and probably the whole wiki) for the 2.0 release.

Also, I've merged #722 now, so you can finish up this PR. Getting close!

mig5:
Thanks @micahflee - argh, this is so close. It turns out that a QCheckbox can't be rich text (contain a hyperlink). Instead they still need to be QLabels in their own right.

I initially placed them underneath the checkbox but didn't like the look. So I've added them  as widgets inside a QHBoxLayout (and then add a widget with that layout), but now the checkboxes are indented compared to the other checkboxes. Do you want to copy my branch and fix if it's easy for you?

mig5:
@micahflee NM, I fixed it. I think this is now ready (albeit the tests fail, but I think that's due to the public mode 404 route stuff)

micahflee:
Can you update `stdeb.cfg` to use these? We now require python 3.6+ because of the sha3 stuff. (And now that Ubuntu bionic is the new LTS the older versions are obsolete, I can't make a PPA release to older versions.)

```
Suite: bionic
X-Python3-Version: >= 3.6
```

micahflee:
Also, I found a bug. I just started OnionShare in share mode, public mode (also I started with `--local-only`). My OnionShare URL is http://127.0.0.1:17650/, but the download link in the corner links to http://127.0.0.1:17650/kitty-guts/download -- it includes the slug even though it shouldn't. And clicking that brings me to a 404 error.

Receive mode handles this by passing different paths into the template (for the action attr of the upload and close forms):

```python
if self.common.settings.get('public_mode'):
    upload_action = '/upload'
    close_action = '/close'
else:
    upload_action = '/{}/upload'.format(self.slug)
    close_action = '/{}/close'.format(self.slug)
```

We'll have to do something similar for the download link in the send template.

micahflee:
Also I like the https://github.com/micahflee/onionshare/wiki/Legacy-Addresses wiki page!

micahflee:
And finally a few more things about the settings dialog.

We're mixing and matching the term "settings" and "options" -- maybe we should just stick to "settings" since it's the "Settings" dialog. Can you change the headers to "General", "Sharing", and "Receiving"?

The auto-stop timer is actually a general setting, not just a sharing setting. Can you move it into the "General" section?

Can you rename "OnionShare is open to the public" to "Public mode (what's this?)" with a link to the wiki (I'll go through and thoroughly update the wiki before the 2.0 release too!) -- it just seems odd to say "OnionShare is" when really we're referring to the specific share. And even saying "don't prevent people from guessing the OnionShare address" is confusing without explaining slugs and rate-limiting.

Can you re-order the General settings to this?

- "Public mode (what's this?)"
- "Use auto-stop timer (what's this?)" -- we might as well link to the wiki for every setting, no?
- "Use legacy addresses (what's this?)"
- "Use a persistent address (legacy) (what's this?)"
- "Create stealth onion services (legacy) (what's this?)"

Finally, I think it's reasonable now to hide the legacy settings. When "Use legacy addreses" is unchecked, the two legacy settings can be totally hidden. When it's checked, they can be shown.

mig5:
> Also, I found a bug. I just started OnionShare in share mode, public mode (also I started with --local-only). My OnionShare URL is http://127.0.0.1:17650/, but the download link in the corner links to http://127.0.0.1:17650/kitty-guts/download -- it includes the slug even though it shouldn't. And clicking that brings me to a 404 error.

I don't reproduce this :( it works as expected for me both local-only and real mode. The Download Files button is just a route to /download.

The template already has this:

```
                {% if slug %}
                <li><a class="button" href='/{{ slug }}/download'>Download Files</a></li>
                {% else %}
                <li><a class="button" href='/download'>Download Files</a></li>
                {% endif %}
```

mig5:
@micahflee thanks for the thorough review. I think I've addressed all the issues. Note as above, I couldn't reproduce the slug appearing in the Download page when using public mode..

Can you find a way to align all those 'what's this' labels to the right? addStretch() doesn't seem to be doing anything here for some reason.

onionshare/onionkey.py
```diff
@@ -0,0 +1,129 @@
...
+    Returns both the serialized private key (compatible with Stem)
+    as well as the public key (.onion hostname)
+    """
+    # Generate v2 Onion Service private key
+    private_key = rsa.generate_private_key(public_exponent=65537,
+                                           key_size=1024,
+                                           backend=default_backend())
+    hs_public_key = private_key.public_key()
+
+    # Pre-generate the public key (.onion hostname)
```
maqp: Unlike with v3 services, the address of v2 services is not the public key, just a truncated hash of it. AFAIK the domain name is called "service ID" by the Tor devs.

PR #745, opened 2018-09-16
Update all pip dependency versions

### LINKED ISSUE #740: Update pip dependencies
For the 2.0 release, we should update all the pip dependencies for Windows and Mac. There was recently a CVE in flask: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-1000656

micahflee:
Fixes #740. I updated all of the pip dependencies, not just flask, and I did some very basic tests in Mac and Windows and everything seems to work fine.

Also, interesting enough, github has been warning me about this too. Pretty cool.

screen shot 2018-09-15 at 8 34 46 pm


mig5:
This is looking good on MacOS.  When running with the --debug flag, a warning is now printed by Flask to the console in red 'WARNING: Do not use the development server in a production environment. Use a production WSGI server instead'. I think this is harmless for our use case, but I wonder if there's a way to silence that warning. Not a blocker.

mig5:
I got a funky message in my Windows env but I think my Windows VM might be a bit messed up. I didn't see the message after a second try. And things seem to work OK too.

screen shot 2018-09-16 at 2 46 13 pm

micahflee:
Regarding the flask warning, it happens even without `--debug`, including when just using the CLI. I'm going to see how easy it is to suppress the warning...

PR #750, opened 2018-09-16
Bump windows versions of pip packages to match those of OS X

mig5:
Another pass at #740 ~~~but I haven't tested in Windows yet~~~. This time I diffed the files!

I wasn't sure if requirements-windows needs PyQt5-sip explicitly, as is in requirements.txt? It's the only package that requirements.txt has that requirements-windows.txt doesn't (as opposed to the other way around, where windows has several extra packages).


micahflee:
Thanks! I'll test this soon when I'm in front of my computer with my Windows VM.

I actually think a bunch of these dependencies we don't need explicitly. I pip installed them all to the latest version, then ran pip freeze to generate requirements-windows.txt, so several of the dependencies should get handled automatically.

Issue #748, opened 2018-09-16
Final receive mode changes


micahflee:
The #695 PR is becoming a monster, and is more current than develop. I think it's not _quite_ done, but we should merge anyway if it's in a good state. This issue is to make sure we continue to keep track of the final bits that need finishing.

Here's a to do list copied and pasted from that PR:

* If an upload fails to finish, it should be marked as failed in the upload history
* Prevent 0 byte, filename-less files (this happened when I made a public receive mode for some reason)
* Often, the open folder icon doesn't appear next to uploaded file. It should always appear

Finished:

* ~~In Upload History window, really long filenames should be word-wrapped #704~~
* ~~Related, if I start the server in receive mode, start uploading large files, then stop the server before they've uploaded, the upload history window still shows the in-progress upload. If I then start the server again, it crashes #756~~
* ~~Make sure `Settings.build_default_downloads_dir()` works in Windows~~
* ~~Update `install/onionshare.nsi` to make the Windows installer/uninstaller use all the new filenames~~
* ~~More thoroughly test the toggle button for Uploads/Downloads, making sure it actually scrolls to the bottom at the appropriate times, etc. #757~~
* ~~Improve the language of strings, based on feedback from @glenn-sorrentino~~
* ~~Change receive mode web interface, based on feedback from @glenn-sorrentino~~
* ~~In settings dialog, persistent address is a common setting, not just for share mode~~
* ~~Update "uploads complete" and "uploads in progress" numbers, and change the mouseover text #755~~
* ~~You should be able to clear upload history #766~~
* ~~In ReceiveMode, make `Uploads.open_folder()` actually open folders with specific file selected in Windows and macOS #763~~
* ~~In `ReceiveMode.timeout_finished_should_stop_server()`, wait for current uploads to finish before stopping the server, but still prevent new uploads from starting #752~~

mig5:
"I'm finished sending" button can be pressed by a user whilst another user is still uploading a large file.

The button should be removed somehow or its route made to do nothing if an upload is still in progress.

The problem is, doing so leaks the fact that an upload *is* in progress, which might be bad for security.

Should we just prevent the option for the sender to close the share, placing that responsibility on the user who is running OnionShare?

micahflee:
> Should we just prevent the option for the sender to close the share, placing that responsibility on the user who is running OnionShare?

[...]

PR #752, opened 2018-09-17
Hold a share open if its timer has expired but a file is still uploading.

mig5:
Took a pass at "In ReceiveMode.timeout_finished_should_stop_server(), wait for current uploads to finish before stopping the server, but still prevent new uploads from starting" from #748 

The only awkward thing is that once a large upload *has* completed after the share was meant to timeout, the share immediately stops, and as a result, the user gets a connection error when they were otherwise meant to be redirected to the share page. Maybe we need a sleep() in the function that calls timeout_finished_should_stop_server() to give it time to load before closing..

Also, I block attempts to upload or close the share by other users/tabs in the browser when the share has meant to timeout but is waiting on that other upload. I took the easy way out and just force a 404 route on /upload and /close in that circumstance rather than adding any more HTML templates for this corner case.

As mentioned in #748 I think we have a bigger issue with the 'I'm Finished Uploading' button in that other users can cut off a large upload (running timer or not), but hiding it during a large upload also gives away that a large upload is taking place. Personally I wonder if we should just remove that /close route/button altogether...


onionshare/web.py
```diff
@@ -343,96 +344,104 @@ def upload_logic(slug_candidate=''):
...
-                self.add_request(Web.REQUEST_ERROR_DOWNLOADS_DIR_NOT_WRITABLE, request.path)
-                print(strings._('error_downloads_dir_not_writable').format(self.common.settings.get('downloads_dir')))
-                valid = False
-            if not valid:
-                flash('Error uploading, please inform the OnionShare user', 'error')
-                if self.common.settings.get('public_mode'):
-                    return redirect('/')
-                else:
-                    return redirect('/{}'.format(slug_candidate))
+            while self.can_upload:
```
micahflee: I don't quite understand what this while loop is doing. I searched the code for `self.can_upload` and it seems to that it never gets set to False, and it seems like all code paths return from the function before ending the first loop anyway. Am I missing something?

mig5:
Oh, is the while loop not necessary? I guess it's not in fact, since there's logic in the /upload stuff (which I added later) to prevent another upload whilst the timer has expired..

onionshare/web.py
mig5: No, I'm probably just stupid.

self.can_upload gets set to False in the GUI in  timeout_finished_should_stop_server(), when the timer hsa expired, as a way of saying 'the timer's run out, but there's an upload running, no-one else Can Upload (false)'

mig5:
I've encountered another bug where if a smaller upload gets through first, the timer will stop and end the service immediately (because upload_count != 0).

Maybe don't wait on me to fix this (e.g merge receiver-mode-gui into develop, and I can modify this PR to target develop branch afterward)...

mig5:
Nearly there now. Fixed those various issues (we throw a 403 instead of a 404 now for new uploads after timer expires).

But I've found another strange bug, a 404 on the Flask shutdown route:

```
Sep 18 2018 08:34:14 AM] Web.check_slug_candidate: slug_candidate=k3u2escilc5g6dyqyvpjg6nclu, slug_compare=k3u2escilc5g6dyqyvpjg6nclu
127.0.0.1 - - [18/Sep/2018 08:34:14] "GET /k3u2escilc5g6dyqyvpjg6nclu/shutdown HTTP/1.1" 404 -
```

mig5:
As discussed separately - the shutdown slug bug is pre-existing and being fixed as we speak.

So I've fixed the conflict in this branch, I think it is therefore essentially ready to merge in. I also removed the 'Receive mode can be stopped by the sender' feature.

mig5:
Changed branches now that receiver-mode-gui is merged (!)

maqp:
BTW is there a chance a second download can be initiated while the first download is still in progress? This would allow attacker to keep the service online indefinitely after the timer has run out, by re-downloading the file using multiple browsers.

mig5:
> BTW is there a chance a second download can be initiated while the first download is still in progress and the timer has reached zero?

This PR is designed to prevent exactly that..

micahflee:
Actually, I found another issue... when the upload finally does finish, the upload history says it's stuck at 99%, and the server still remains open forever.

mig5:
Trying to program 'time' really is harder than it looks! :crying_cat_face:

mig5:
I've fixed the 403 on index page, though keep in mind the user sending that last upload is going to either get a 403 or a connection error since the share will stop right after the upload finishes... could do with better UX at some stage, but this is a pretty obscure corner case.

Will look at the 99% thing now, I guess it's something to do with the Web Requests or some signal/slot somewhere..

mig5:
Nope can't figure out how best to fix best. 

The issue is in `ReceiveModeRequest.init()` where we need to check if self.web.can_upload is True (so that we don't append the new upload attempt to the count or the in progress queue). If it's false, self.upload_request also stays false.  And this is why the `ReceiveModeRequest.close()` function doesn't fire and therefore 'complete' the upload.

We can take out `self.web.can_upload` from `ReceiveModeRequest.init()` so that `self.upload_request` becomes True and therefore `close()` fires properly, and the upload 'completes' and server shuts down.  

But a *subsequent* upload attempt will manage to insert itself into the queue and appear in the upload list but without its file name (because the upload action still fired a 403 later).

So then if we move the `self.web_can_upload` out of that 'top' condition in `ReceiveModeRequest.init()` and down to line 759, it won't add to the upload queue, but now a new problem: `close()` throws an exception because now there's no `self.upload_id` for this new request, and the page throws a 500 internal server error.

I can't find the sweet spot.

Meanwhile, even after all this, the original user's gonna get a timeout even after their upload completes successfully, because the share stops as soon as it hits 100%.

Might need someone smarter than me (you) to take over from here...

micahflee:
I haven't taken a fresh look at the code, but what about setting a cookie? If we set a session cookie in the user's browser with a random value like a UUID, then we can tell individual http users apart. When the shutdown timer expires, we can make a list of sessions that are currently uploading, and then let them finish their uploads and display a thank you page to them, but give all other users 403 requests?

Of course that's a lot more to implement. I'll take a look soon and see if I can find a simpler way, unless you want to take a stab at the cookie thing.

mig5:
@micahflee I think I've nailed this finally! Didn't end up needing to use sessions or cookies.

I've merged latest develop into this branch and fixed conflicts. In my manual tests:

1) an upload is allowed to complete after the expiry runs out
2) no new uploads or visits are allowed (403)
3) the last upload completes and a thankyou page is shown, and then the service stops
4) you can't access the service at all after this

mig5:
OK, I've said it several times, but it might be working now. I got it to work on CLI mode as intended too.

micahflee:
Hmm, I'm testing it myself and it's stuck at 99% again.

mig5:
Damn, sorry about this. I think I'm struggling too much with the Request middleware there.

mig5:
Damn - I really did not reproduce this, but I only tested with smaller (5MB) files, maybe that is the difference

mig5:
It looks like I lost some changes perhaps on a merge of develop, where the mode stuff was refactored in GUI. The `timeout_finished_should_stop_server` function in receive_mode is still a TODO.

I think I had it working but failed to push a commit, and have since rebuilt my dev VM :( I'll see if I can get these changes back.

```
    def timeout_finished_should_stop_server(self):
        """
        The shutdown timer expired, should we stop the server? Returns a bool
        """
        # TODO: wait until the final upload is done before stoppign the server?
        return True
```

mig5:
@micahflee I added the missing bit - please let me know if it's better via GUI now.

Issue #743, opened 2018-09-15
Receive mode should organize files for you


micahflee:
Instead of files all going to one place like `~/OnionShare`, it should organize it better. Like maybe:

```
~/OnionShare/[YYYY-MM-DD]/[HH.MM.SS]/[filename]
```

So if someone uploads 3 files in one POST request right now, someone uploads one in a POST request in 5 minutes, and someone and someone uploads 1 file tomorrow, it would like this:

```
~/OnionShare/2018-09-14/17.43.10/file1
~/OnionShare/2018-09-14/17.43.10/file2
~/OnionShare/2018-09-14/17.43.10/file3
~/OnionShare/2018-09-14/17.48.32/another_file
~/OnionShare/2018-09-15/09.00.02/whatever.pdf
```

This also means we'd no longer have to worry about renaming files because there won't be filename collisions.

micahflee:
I guess unless multiple uploads happen the same second. In which case, we can keep the filename collision code.

PR #763, opened 2018-09-19
Open folder containing uploads with the file selected, in macOS and Windows

mig5:
Works in macOS. I'm still in the process of loading up Windows again to try it..


mig5:
OK, it doesn't work on Windows, it just opens Explorer at This PC -> Documents instead of the OnionShare directory :/ sigh...

mig5:
OK, this should now work on Windows too.

I also identified a bug with `build_default_downloads_dir()`. On Windows, `os.path.expanduser('~/OnionShare')` was actually retaining the last forward slash. This seemed to be fine in terms of saving the file, but it breaks explorer when trying to open the folder.

So this change also explicitly makes it a backslash for windows, which makes explorer happy, and doesn't cause any problems as far as I can tell. But be aware that your saved onionshare.json on your Windows VM probably has the 'forward slash' version saved already. You should probably delete your onionshare.json to let it save it with the corrected version, before testing opening the folder of an uploaded file.

mig5:
This will probably need refactoring by the time #743 happens. (I had a crack at that feature yesterday, and even had files uploading into a timestamped folder as intended, but I could not successfully communicate the timestamped folder back to the UI in order to change the 'link' in each upload's folder icon. As usual, what looked like a simple job turned out to be much bigger!)

mig5:
I've merged develop into this and fixed conflicts. Let me know if it works for you on macOS and Windows?

PR #771, opened 2018-09-20
Update BUILD.md

19hundreds:
python3-crypto was missing. Without it you get
`ModuleNotFoundError: No module named 'nacl'` while building deb package
I'm not sure if the package is required in Fedora as well

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -11,7 +11,7 @@ cd onionshare
 
 Install the needed dependencies:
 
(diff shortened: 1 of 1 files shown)


mig5:
Hi @19hundreds,

Please use the `develop` branch, which already contains this change.

`master` branch is very far behind right now as it's our 'pristine' release branch, and we work off `develop` as our default branch. You probably cloned and as a result are on develop branch, but were following the BUILD.md from `master` in some other way.

Cheers

PR #772, opened 2018-09-20
Make fedora and debian dependencies get added to packages built

micahflee:
I tried building an RPM off develop to install in my template, but realized that it didn't install all the dependencies.

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -1,6 +1,6 @@
 [DEFAULT]
 Package3: onionshare
-Depends3: python3-flask, python3-stem, python3-pyqt5, python-nautilus, tor, obfs4proxy
-Build-Depends: python3-pytest, python3-flask, python3-stem, python3-pyqt5
+Depends3: python3-flask, python3-stem, python3-pyqt5, python3-cryptography, python3-crypto, python3-nacl, python3-socks, python-nautilus, tor, obfs4proxy
+Build-Depends: python3-pytest, python3-flask, python3-stem, python3-pyqt5, python3-cryptography, python3-crypto, python3-nacl, python3-socks, python-nautilus, tor, obfs4proxy
 Suite: bionic
 X-Python3-Version: >= 3.6
--- a/BUILD.md
+++ b/BUILD.md
@@ -11,11 +11,11 @@ cd onionshare
 
 Install the needed dependencies:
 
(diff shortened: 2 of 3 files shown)


stdeb.cfg
```diff
@@ -1,6 +1,6 @@
 [DEFAULT]
 Package3: onionshare
-Depends3: python3-flask, python3-stem, python3-pyqt5, python-nautilus, tor, obfs4proxy
-Build-Depends: python3-pytest, python3-flask, python3-stem, python3-pyqt5
+Depends3: python3-flask, python3-stem, python3-pyqt5, python3-cryptography, python3-crypto, python3-nacl, python3-socks, python-nautilus, tor, obfs4proxy
+Build-Depends: python3-pytest, python3-flask, python3-stem, python3-pyqt5, python3-cryptography, python3-crypto, python3-nacl, python3-socks, python-nautilus, tor, obfs4proxy
 Suite: bionic
 X-Python3-Version: >= 3.6
```
mig5: Does `X-Python3-Version: >= 3.6` mean it is literally *impossible* to build a .deb on a platform that has an older Python3 version?

mig5: Related to this and the above comment - is there any implication for Tails here, if they decide to package the latest OnionShare? I thought 3.x was also based on Debian Stretch (Python 3.5).

I feel like adding python3-sha3 might be important for such projects in case they decide to package downstream.. though I appreciate it looks not possible for Fedora..

micahflee:
Hmm these are good questions about Whonix and Tails support. It would be pretty cool to have OnionShare 2.0 in both of those, but they're both based on Debian stretch. The also both have the stretch-backports repo enabled.

The problem is that stretch just doesn't have a `python3-sha3` package at all, but buster does. So I think, because we really want to ship with v3 onion support, the only option for getting OnionShare 2.0 in stretch would be:

- Get `python3-sha3` backported into stretch-backports
- Get `onionshare` 2.0 backported into stretch-backports, which would basically just mean adding the `python3-sha3` dependency, and lowering the python3 version to 3.5.3.

If we left `python3-sha3` as a dependency (in `stdeb.cfg`, not just in `BUILD.md`) then it still wouldn't work in stretch -- a required package wouldn't be available.

So I think the best way forward is to make OnionShare 2.0 require python 3.6+ and not require `python3-sha3`, and then try to get debian developers to do the work necessary to backport it to stretch. I just opened an issue #775 about that that we can start figuring out after the 2.0 release.

Does that sound reasonable?

mig5:
Ugh, good call, I forgot that stretch doesn't have python3-sha3. This is the right approach.

u451f:
Note that Debian Buster will be released in March 2019. Tails works on porting Tails to Buster at the same time. Hence I'm not sure if it's worth the effort backporting to Stretch (depending on when you would want to release Onionshare 2.0.

PR #774, opened 2018-09-21
Don't zip if only sharing one file, and big refactor of web module

micahflee:
In addition to resolving #663, this PR also includes a big refactor of the web module.

Sort of like how we refactored the GUI code to split the share and receive mode logic out into separate modules, I've now done this with the web module too. Now, all of the share mode-specific code is in a separate ShareModeWeb class, and the receive mode-specific code is in a ReceiveModeWeb class. This will make it much cleaner in the future to implement future modes (like static web server, etc.).

In share mode, if you're only sharing a single file, it no longer zips it up and serves the zip file. Instead, it gzips that single file. If the browser supports gzip encoding, it serves the gzipped version with the header `Content-Encoding: gzip`, and otherwise it servers the original uncompressed file. Since Tor Browser supports gzip by default, in nearly all cases the file will still be compressed.

You can test this with `curl` and the `--compressed` flag. If you `curl -v http://[address]/[slug]/download > output`, it won't use compression. But if you `curl -v --compressed http://[address]/[slug]/download > output` it will add `Accept-Encoding: deflate, gzip` to the request, and the response will be gzip-compressed. However, the ultimate file downloaded should be exactly the same either way because the http client decompresses it after downloading -- just the download size will be smaller if you use compression.

Like with zipping, the GUI shows a compression progress bar when gzipping. And both the final zip, or final gzip, files should get cleaned up automatically when the share is complete.


mig5:
Nice! The web stuff is IMO the most complex bit of code to read so this is great.

I found one, maybe two issues.

The first looks like a bug. When sharing a single file (which works great! And I confirmed in TBB's dev toolbar that I see the gzip content encoding), it seems the progress bar (and that which is printed to stdout on command line when in --debug mode), never seems to reach 100%.

onionshare-computing-eta

it works fine when it's a .zip containing multiple files. So there's probably some bug in updating the progress when it's just a single file.

The second issue may not be an issue, but I wondered if you needed to add `onionshare.web` to the list of modules in setup.py, similar to this change? https://github.com/micahflee/onionshare/pull/772/commits/12a5b68d16084499d79f3338ce4dfb341773c1fa

Otherwise, looking good!

micahflee:
> Or is it reporting an ironic 'reduced' (compressed) length of bytes due to the gzipping perhaps?

Hah, I just confirmed that this is indeed the issue.

mig5:
Sorry Micah, still one tiny detail: in the stdout when running from ./dev_scripts as either the GUI or CLI program, the individual file is still reported as its compressed size/progress:

```
127.0.0.1 - - [22/Sep/2018 09:38:17] "GET /spectator-chant/download HTTP/1.1" 200 -
1.4 KiB, 40.58%          
Stopped because download finished
```

compared to a .zip file of several files:

```
127.0.0.1 - - [22/Sep/2018 09:35:39] "GET /donut-remote/download HTTP/1.1" 200 -
3.8 KiB, 100.00%   
```

I think it's just wherever Web does its print() to stdout, it needs to use the gzipped filesize.

mig5:
I tried changing this:

```
                            percent = (1.0 * downloaded_bytes / self.download_filesize) * 100
```

to use `self.gzip.filesize` conditionally (else `self.download_filesize`) but this seems to trigger an exception (and therefore the 'Canceled' catch that follows) :confused: maybe it doesn't like calculating the percentage of the smaller size?

micahflee:
The tricky thing is that we can't just always use either `self.download_filesize` or `self.gzip_filesize` -- it depends on the request. Some browser might not support gzip (e.g. `curl` without `--compressed`) in which case we actually use `self.download_filesize`. But when they do support gzip, we then use `self.gzip_filesize`.

micahflee:
Does that fix it?

mig5:
This does seem to fix it!

PR #751, opened 2018-09-17
Some potential savings and fixes to language

comradekingu:
I tried to find the wins I could.

PR #777, opened 2018-09-21
Remove unused wait_for_hs string, and change tests to test with a dif…

micahflee:
…ferent string. Also remove a few other unused strings.


micahflee:
Also, I'm making a few other changes related to strings in this PR, thanks to feedback from #751.

mig5:
Ah hah. I couldn't believe `wait_for_hs` stayed in after #724, but it was because of the test. Nice..

PR #786, opened 2018-09-23
Language rework

comradekingu:
"Completion" could be "delivery"
"Receive mode" could be "Inbox mode"


share/locale/en.json
```diff
@@ -1,184 +1,184 @@
...
-    "timeout_download_still_running": "Waiting for download to complete",
-    "large_filesize": "Warning: Sending large files could take hours",
+    "no_available_port": "Could not find an available port to start the onion service",
+    "other_page_loaded": "Inbox address loaded",
+    "close_on_timeout": "Stopped because timeframe ran out",
+    "closing_automatically": "Stopped because you have the complete share",
+    "timeout_download_still_running": "Awaiting completion of share transfer to you",
+    "large_filesize": "Warning: Sending a large share could take hours",
     "systray_menu_exit": "Quit",
-    "systray_download_started_title": "OnionShare Download Started",
```
comradekingu: In my head a download is something sent to me. This lacks any distinction to that effect.

share/locale/en.json
```diff
@@ -1,184 +1,184 @@
...
-    "gui_please_wait": "Starting… Click to cancel",
-    "gui_download_upload_progress_complete": "%p%, Time Elapsed: {0:s}",
-    "gui_download_upload_progress_starting": "{0:s}, %p% (Computing ETA)",
+    "gui_copied_url_title": "Copied",
+    "gui_copied_url": "OnionShare address copied to clipboard",
+    "gui_copied_hidservauth_title": "Copied",
+    "gui_copied_hidservauth": "HidServAuth line copied to clipboard",
+    "gui_please_wait": "Starting… Click to cancel.",
+    "gui_download_upload_progress_complete": "%p%, {0:s} elapsed.",
+    "gui_download_upload_progress_starting": "{0:s}, %p% (calculating end time.)",
```
comradekingu: Maybe "delivery time" is less dystopic?

share/locale/en.json
```diff
@@ -1,184 +1,184 @@
...
+    "gui_receive_start_server": "Receive",
+    "gui_receive_stop_server": "Stop",
+    "gui_receive_stop_server_shutdown_timeout": "Stop receiving ({}s remaining)",
+    "gui_receive_stop_server_shutdown_timeout_tooltip": "Allowed receive timeframe ends {}",
     "gui_copy_url": "Copy Address",
     "gui_copy_hidservauth": "Copy HidServAuth",
-    "gui_downloads": "Download History",
-    "gui_no_downloads": "No downloads yet.",
+    "gui_downloads": "History of sent shares",
+    "gui_no_downloads": "None yet.",
```
comradekingu: Copy that floppy to get started. :)

share/locale/en.json
```diff
@@ -1,184 +1,184 @@
...
-    "settings_error_unknown": "Can't connect to Tor controller because the settings don't make sense.",
-    "settings_error_automatic": "Can't connect to Tor controller. Is Tor Browser running in the background? If you don't have it you can get it from:\nhttps://www.torproject.org/.",
-    "settings_error_socket_port": "Can't connect to Tor controller on {}:{}.",
-    "settings_error_socket_file": "Can't connect to Tor controller using socket file {}.",
+    "gui_settings_shutdown_timeout_checkbox": "Stop sharing after a given timeframe",
+    "gui_settings_shutdown_timeout": "Allowed sharing timeframe:",
+    "settings_saved": "Settings saved in {}",
+    "settings_error_unknown": "Can't connect to Tor controller because it knows your settings aren't making sense.",
+    "settings_error_automatic": "Could not connect to the Tor controller. The Tor Browser (available from https://www.torproject.org/)\nneeds to be running in the background.",
+    "settings_error_socket_port": "Can't connect to the Tor controller at {}:{}.",
```
comradekingu: Is {}:{} a port?

share/locale/en.json
```diff
@@ -1,184 +1,184 @@
...
+    "gui_tor_connection_ask": "Open the settings to sort out connection to Tor?",
+    "gui_tor_connection_ask_open_settings": "Yes",
     "gui_tor_connection_ask_quit": "Quit",
-    "gui_tor_connection_error_settings": "Try adjusting how OnionShare connects to the Tor network in Settings.",
-    "gui_tor_connection_canceled": "OnionShare could not connect to Tor.\n\nMake sure you're connected to the Internet, then re-open OnionShare to set up the Tor connection.",
+    "gui_tor_connection_error_settings": "Try changing how OnionShare connects to the Tor network in the settings.",
+    "gui_tor_connection_canceled": "Could not connect to Tor.\n\nEnsure you are connected to the Internet, then re-open OnionShare and set up its connection to Tor.",
     "gui_tor_connection_lost": "Disconnected from Tor.",
-    "gui_server_started_after_timeout": "The server started after your chosen auto-timeout.\nPlease start a new share.",
-    "gui_server_timeout_expired": "The chosen timeout has already expired.\nPlease update the timeout and then you may start sharing.",
```
comradekingu: How does one update the timeout?

micahflee:
I don't understand the terminology "inbox mode". And several of the strings that you changed to say "inbox mode", "inbox address", "inbox folder" are strings used in both share and receive mode (and, in the future, other modes too), so changing them to say "inbox" there is confusing.

I do agree that the download and upload terminology is a little confusing. When you use OnionShare, your computer itself _is_ the web server. So when you're sharing files, people _download files from you_, and when you're receiving files, people _upload files to you_. If we do decide to totally change the existing terminology, we should discuss it first and come to an agreement on what terms we should use instead, and exactly what they describe.

But also a lot of these strings contain improvements. I'll take a stab at merging the simple ones that have improvements, and we can hold off on completely changing terminology for a new release in the future, maybe version 2.1.

micahflee:
Let's stick with the current terms for now -- I think they work better when looking at them within the UI, compared to just looking at the strings file -- but I'd be into opening a new issue here with some mockups of the GUI that contain new language, so we can discuss if that would be a better for or not.

PR #788, opened 2018-09-25
Analyse the right file size to determine if the download has finished…

### LINKED ISSUE #787: "Stop sharing after first download" seems to not work in 2.0.dev
Using v3 onion services, make sure "Stop sharing after first download" is checked. Then share something, and download it. The server stays open. This must have been a regression bug probably after the refactor of web?

mig5:
… in the UI (in order to decide whether to stop server yet)

Fixes #787 I think - this was a regression related to gzip compression: the conditional in the UI that decides whether to fire `stop_server()` was based on the wrong file size.


micahflee:
I confirmed that this works great.

PR #796, opened 2018-09-30
Ignore attribute error when optimistically trying to cancel compression

### LINKED ISSUE #790: When I clicked on "Stop Share" onionshare crashed
I built from the latest develop branch and onionshare crashed when I clicked the stop share button.

```
[Sep 24 2018 23:15:34] Web.check_shutdown_slug_candidate: slug_candidate=ymf5eposfwver2gpegmic5drki
127.0.0.1 - - [24/Sep/2018 23:15:34] "GET /ymf5eposfwver2gpegmic5drki/shutdown HTTP/1.1" 200 -
[Sep 24 2018 23:15:34] OnionShare.cleanup
[Sep 24 2018 23:15:34] Onion.cleanup
[Sep 24 2018 23:15:34] Onion.cleanup: trying to remove onion dfi5lhiakae2ztq2t4qd7xh2khrnddsfwldkeczafznfoitua7wbj5yd
[Sep 24 2018 23:15:34] CompressThread.cancel
  File "onionshare_gui/share_mode/threads.py", line 59, in cancel
AttributeError: 'Web' object has no attribute 'zip_writer'
Abort trap: 6
```
@mig5

mig5:
Fixes #790 

It seems that the `cancel()` function in `CompressThread` can fire before `ShareModeWeb` has executed its `set_file_info()` function. As a result, there is no `ZipWriter` object yet, and an AttributeError occurs.

To replicate #790:

1) start a share of a file, and let it complete its setup. 
2) stop the share.
3) start a new share, wait a second or two, but cancel it

..you get the attribute error.

I don't know if this is the correct way to fix it, but it seems harmless enough to optimistically try to set the value and ignore the AttributeError here, since we are dealing with the asymmetry of threads and timing here.

### DIFF
--- a/onionshare_gui/share_mode/threads.py
+++ b/onionshare_gui/share_mode/threads.py
@@ -56,5 +56,8 @@ class CompressThread(QtCore.QThread):
 
         # Let the Web and ZipWriter objects know that we're canceling compression early
         self.mode.web.cancel_compression = True
-        if self.mode.web.zip_writer:
+        try:
             self.mode.web.zip_writer.cancel_compression = True
+        except AttributeError:
+            # we never made it as far as creating a ZipWriter object
+            pass
--- a/tests_gui_tor/onionshare_790_cancel_on_second_share_test.py
+++ b/tests_gui_tor/onionshare_790_cancel_on_second_share_test.py
@@ -1,0 +1,197 @@
+#!/usr/bin/env python3
+import os
+import sys
+import unittest
+import pytest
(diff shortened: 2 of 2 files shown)


mig5:
Also added a GUI unit test for this bug! So you could also run `py.test-3 onionshare_790_cancel_test.py -vvv` on `develop` branch and it should fail without the patch, but succeed with it.

OnionShare, welcome to test-driven development!

PR #798, opened 2018-09-30
Toggle history button, and history indicator

### LINKED ISSUE #780: Make some new UI changes
Here's feedback from @glenn-sorrentino: https://github.com/micahflee/onionshare/pull/770#issuecomment-423412473

>I explored using a button/toggle to open up the download history. Would love your thoughts on it.
>
>Next to the status indicators is a section for opening a panel showing the download history.
>
>toggle close
>
>The empty state looks a bit more like the empty state for uploading
>
>toggle open - no downloads
>
>When a new download happens with the panel closed, an indicator is displayed over the icon.
>
>toggle closed - indicator
>
>Opening it clears the count on the toggle and shows all past and current downloads
>
>toggle open - download history

I really like the new empty state look, and we should definitely implement that.

I also really like the look of the "clear all" link, and the download count.

If we do add toggling showing and hiding, I like the status indicator with a number in it. But I'm not entirely convinced we should allow toggling... besides it just probably being tricky to implement correctly, I kind of think the upload/download progress is really the important information once you've already started a share and have sent someone the address, so I'm not sure that it's useful to allow people to hide it. But maybe I'm just anticipating weird window resizing bugs that I don't want to deal with... :)

micahflee:
When this is done, it will resolve #780, based on @glenn-sorrentino's work (it doesn't copy it exactly because that was difficult, but I think it gets all of the important stuff).

I think it's nearly there, but there are a few UX things to workout. I decided to tackle the window resizing problem head-on and I think it's nearly solved!

Here's what share mode looks like when you first open it:

screenshot from 2018-09-30 11-43-14

And when you click the toggle downloads button:

screenshot from 2018-09-30 11-43-44

Now let's hide the history and share some files.

screenshot from 2018-09-30 11-44-32

Now I download these files in Tor Browser. (Note, stop sharing after first download is disabled.) See the history indicator?

screenshot from 2018-09-30 11-45-47

When I click it, the indicator disappears and the count goes back down to zero.

screenshot from 2018-09-30 11-46-23

One change I made here is that, in order to always allow the the user to toggle the history on and off, the info widget is always showing now. But in share mode, it just hides the label from it when it isn't useful.

If the share mode history is visible but receive mode history is hidden, when you flip between the two modes the window now automatically resizes to a reasonable size. In order to make this possible, OnionShareGui now has a `adjust_size` slot, that can get triggered by the modes emitting their `adjust_size` signals. This will recursively run `.adjustSize()` on every widget, while running `self.qtapp.processEvents()` at specific times, which seems to give Qt enough information to actually resize properly.

This took some figuring out, and it's still not perfect. Basically, without the `processEvents()` calls, the final window didn't get resized to be smaller, even though it's sizeHint said it was supposed to be smaller. But there are still a few situations where things get their size adjusted to be way too small. I actually think carefully placing a `processEvents()` call in the right places might fix this, but it would be helpful to be able to consistently reproduce the sizing issues, which I haven't spent time on yet.

Finally, all of the tests pass, including Tor tests. I did a little bit of refactoring of tests, too. I made it so `tests_gui_tor.CommonTests` inherits from `tests_gui_local.CommonTests`, so now we can avoid a swath of duplicated code. (However, there is still a lot of duplicated code in the tests, and I have some ideas for house to reduce that duplication, but I'm going to do that in a separate PR.) I only actually test the new history indicator in local tests right now, not in Tor tests.

@mig5 this is ready to review. I expect to have to fix a few problems before merging, but I think it's ready to have a second set of eyes on it.

Also, here's what it looks like with several downloads:

screenshot from 2018-09-30 12-01-03

Here's a receive mode screenshot with the history indicator:

screenshot from 2018-09-30 12-03-16

And here it is again with the history showing:

screenshot from 2018-09-30 12-03-38


onionshare_gui/receive_mode/uploads.py
```diff
@@ -299,22 +280,116 @@ def reset(self):
         """
         Reset the uploads back to zero
         """
-        self.common.log('Uploads', 'reset')
         for upload in self.uploads.values():
-            upload.close()
```
mig5: @micahflee the removal of `upload.close()` here I think introduced the regression of seeing 'old' uploads per my review. This was necessary in the old code to actually pop an upload out of being displayed permanently on 'clear history'

mig5: Confirmed in my tests that re-instating `upload.close()` here fixes the widgets from re-appearing/overlapping each other.

Upload widget is different to Download widget because it's not just a progress bar, it's a collection of child widgets. That's why it's not enough to just close its child progress bar.

onionshare_gui/onionshare_gui.py
```diff
@@ -194,8 +199,8 @@ def update_mode_switcher(self):
             self.share_mode_button.setStyleSheet(self.common.css['mode_switcher_selected_style'])
             self.receive_mode_button.setStyleSheet(self.common.css['mode_switcher_unselected_style'])
 
-            self.share_mode.show()
             self.receive_mode.hide()
+            self.share_mode.show()
```
mig5: I don't know why, but undoing this change whereby `self.share_mode.show()` occurs first (before `self.receive_mode.hide()` is called), is enough to fix the weird download history widget from being resized badly (and the indicator icons moving to the left), per my earlier review. 

It does seem to cause a slight vertical stretch in share mode when this change is reverted. but I'm not sure that's a big deal, share mode is always 'bigger' anyway due to the FileList and associated widgets.

If you were to revert this change, and re-add the `upload.close()` as per the other comment, and fix the conflicts, I think this could be merge-able (and then #800 could be merged and we could move on to fixing the rest of the tests)

mig5:
I've pushed, just to my own fork's branch, my proposed changes that fix upload widget, seem to fix mode-switching UI resize issues (may not be perfect, I see slight size change of the app window in general after about 10 'switches', but the widgets are not messed up anymore), and fix the test conflict.

Re: The widget issue, I tried adding the adjust_size() just once after the share/receive conditionals, but I didn't get as good results. adjusting size in between showing/hiding modes seems to help a lot.

https://github.com/mig5/onionshare/tree/780_improved_ui

Commits in question: 
https://github.com/mig5/onionshare/commit/c86c6f9db97384388460cd5255171c8d7b3979ce
https://github.com/mig5/onionshare/commit/00f40adaf8aa5411e3294163cccce4c94b4a7866

micahflee:
Eek, I just noticed your branch, while I was sitting there refactoring a bunch of stuff in mine. Checking it out now...

mig5:
hahah I see you've been busy... no worries, my 'fixes' were small, maybe you've already fixed the issues in these commits, reading through now..

mig5:
https://github.com/micahflee/onionshare/pull/798/commits/38e62d85288c9947f089b6ba80dc3146178d2a3a "Also, reset hides individual items because, for some reason, they still show up otherwise." - this sounds like the issue where `upload.close()` had been removed from the code in `reset()`..

micahflee:
Nice, I managed to cherry-pick one of your commits, despite all my refactoring :)

micahflee:
Alright, I still need to merge develop into here and re-deal with those conflicts (and probably more, because I refactored some of the tests a bunch too). And I also still need to go and address some other earlier stuff you were suggesting in your review.

But basically, as you're probably seeing in the commits, I create a new History widget, as well as related widgets like the toggle button, and made both share and receive mode use it. They basically had nearly the exact same code, so this is way more DRY, and easier to work with I think.

So the adjust size thing appears to be working now. But unfortunately, it's still not done. Try maximizing the window and switching modes -- chaos ensures. Same if you snap the window to half of your screen (which you can do easily enough in GNOME -- I'm developing on my Ubuntu laptop at the moment).

Maximized:

screenshot from 2018-10-07 21-54-13

Snapped to half the screen:

screenshot from 2018-10-07 21-54-27

This resizing issue is so annoying... I feel like there's gotta be a simpler way for Qt to just make everything the perfect size :).

micahflee:
I'm Still not ready for a second review, but check it out now! I just increased the minimum window with of the full window, and then removed all of the adjust size logic. So in this branch OnionShare isn't trying to ever resize the window for you, and because of this we don't end up with weird resizing-too-small problems.

What do you think? I think this maybe just solves the problem...

mig5:
This is great - a lot to observe but I can tell it's more efficient code. I cry a bit inside each time I see the test refactoring (because I wish I hadn't wasted so much time doing it the hard way earlier!)

I've indeed spotted two other cases already where the widgets resize funny - 

1. on 'clear' in Receive mode, the logo shifts to the left

onionshare-mon-receive-logo

2. when a timer runs out, the file list and history widget shift.

onionshare-mon-timer

mig5:
Oh nice - I didn't see your last commit, let me test again..

mig5:
@micahflee totally seems to fix it!

I don't mind the wider default view frankly.

Tiny observation: now in receive mode, when toggling *on* the history view, the history view is slightly two wide for the default minimum width (can't see the right-hand borderline of the history box unless you slightly increase width). Maybe just make the minimum width a tiny bit wider.

onionshare-mon-history-receive-overlap

Another tiny observation (I know it's not ready for review yet) - a debug message is printed to console with an object rather than an attribute. Not really a bug, just the only case I think we do this, maybe you intended on printing the upload's filename attribute rather than the object itself.

```
[Oct 08 2018 04:16:17 PM] History.add: id: 0, item: <onionshare_gui.mode.history.UploadHistoryItem object at 0x78b2e03944c8>
```

onionshare_gui/mode/share_mode/__init__.py
```diff
@@ -199,25 +206,25 @@ def stop_server_custom(self):
...
+            self.common.log('ShareMode', 'cancel_server: quitting compress thread')
             self.compress_thread.quit()
 
     def handle_tor_broke_custom(self):
         """
         Connection to Tor broke.
         """
         self.primary_action.hide()
-        self.info_widget.hide()
+        self.info_label.hide()
```
mig5: I think this should be removed or commented out same as Receive mode (info label e.g '7 files, 120 kb' above FileList) is OK to stay)

mig5:
Found another minor resize issue. When toggling *off* history and using v3 onions, the onion address widget gets squashed.

screenshot_2018-10-10_14-14-10

I think this could be rectified by setting a minimumHeight on the app that is slightly larger than the current one. (second screenshot)

screenshot_2018-10-10_14-15-54

onionshare_gui/mode/share_mode/__init__.py
micahflee: Good idea

mig5:
Omg Qt....

So that's looking better for when history is toggled... unfortunately now the URL label is squashed in Share Mode

screenshot_2018-10-10_15-43-55

micahflee:
Weird... on my computer, with the window at it's minimum width and height, this is what share mode looks like both with history hidden and shown:

screenshot_2018-10-09_21-50-13
screenshot_2018-10-09_21-50-21

And this is what receive mode looks like:

screenshot_2018-10-09_21-50-36
screenshot_2018-10-09_21-50-45

micahflee:
I've change the UI since I opened this PR. Now the in progress/completed counters are part of the history, so they only are shown if you toggle history on. Given that, do you still think we need this tooltip? 

> Is it possible to add a tooltip when hovering over the toggle button when it has a count? Perhaps something like 'There's been a new download/upload'. From a UX perspective I wondered if users might get confused between the steadily-incrementing 'green tick' count, and the red count, if they have toggled back and forth the history mode, and therefore seen a disrepancy between the numbers. But again I think this would be unusual user behaviour and not a big deal.

mig5:
Strange, it is consistently squashed for me but only in Share mode. What's your OS here? I'm using a debian-10 AppVM (from a TemplateVM I upgraded from Debian 9 to Debian 10). Could be a Qt version difference?

I agree we no longer need that tooltip, I'll resolve that review note.

Branch is looking really good now other than this odd squash issue.

micahflee:
I'm running Fedora 28 right now. I'll test in Debian 10 too. So here's my to do list for this PR:

* Test in Debian 10 and make sure the widget squishing issues are gone
* Make downloads history show timestamps as well as progress bars

mig5:
(I also reproduce when sharing just one file in OS X, so it looks like it's just an issue with sharing a small number of files as opposed to Qt version)

micahflee:
Strange, I'm not reproducing on my computer in Fedora 28 or Debian 10. Here's my Debian 10 screenshots, without resizing the window at all:

screenshot_2018-10-09_22-31-54
screenshot_2018-10-09_22-32-05
screenshot_2018-10-09_22-32-17
screenshot_2018-10-09_22-32-28

I'll try to reproduce in Windows and Mac next, but I wonder why it works on my Debian but not on yours... anyway, I've gotta go for the night. Feel free to solve this if you have time, otherwise I'll try to get to it tomorrow.

mig5:
Lol, without doing anything I now don't reproduce it on Debian :o but I do on Mac, whether I share 1 or 3 files.. I'll try and debug.

mig5:
I forgot that there's extra padding that occurs in Qt on Mac - something we observed earlier this year. This is why I was still seeing overlap/squash issues on Mac. It was resolved in your last adjustment to setMinimumHeight on Linux.

I've bumped it up from 620 to 650, which resolves it on Mac, and still looks OK on Linux AFAIC. I'm testing now on WIndows.

If you think you're done here I'll do a final review and then merge maybe ~this time tomorrow.

mig5:
All good on Windows!

tests_gui_local/onionshare_receive_mode_upload_test.py
```diff
@@ -94,89 +94,97 @@ def test_settings_button_is_visible(self):
...
-    @pytest.mark.run(order=5)
-    def test_info_widget_is_not_visible(self):
-        CommonTests.test_info_widget_is_not_visible(self, 'receive')
-
     @pytest.mark.run(order=6)
     def test_click_mode(self):
-        CommonTests.test_click_mode(self, 'receive')
+        CommonTests.test_click_mode(self, self.gui.receive_mode)
+
+    @pytest.mark.run(order=6)
```
mig5: Not a major issue in this case, but two tests have `order=6` here.

tests_gui_local/onionshare_receive_mode_upload_test.py
```diff
@@ -94,89 +94,97 @@ def test_settings_button_is_visible(self):
...
+
+    @pytest.mark.run(order=8)
     def test_history_is_visible(self):
-        CommonTests.test_history_is_visible(self, 'receive')
+        CommonTests.test_history_is_visible(self, self.gui.receive_mode)
 
     @pytest.mark.run(order=8)
     def test_server_working_on_start_button_pressed(self):
-        CommonTests.test_server_working_on_start_button_pressed(self, 'receive')
+        CommonTests.test_server_working_on_start_button_pressed(self, self.gui.receive_mode)
```
mig5: Two `order=8` here too

tests_gui_tor/onionshare_receive_mode_upload_test.py
```diff
@@ -94,15 +94,19 @@ def test_settings_button_is_visible(self):
...
 
-    @pytest.mark.run(order=5)
-    def test_info_widget_is_not_visible(self):
-        CommonTests.test_info_widget_is_not_visible(self, 'receive')
-
     @pytest.mark.run(order=6)
     def test_click_mode(self):
         CommonTests.test_click_mode(self, 'receive')
 
+    @pytest.mark.run(order=6)
```
mig5: 2x `order=6`

onionshare_gui/mode/receive_mode/__init__.py
```diff
@@ -154,67 +162,36 @@ def handle_request_upload_file_renamed(self, event):
...
+        self.history.update_completed()
+        self.history.update_in_progress()
 
     def on_reload_settings(self):
         """
         We should be ok to re-enable the 'Start Receive Mode' button now.
         """
         self.primary_action.show()
-        self.info_widget.show()
+        #self.info.show_more()
```
mig5: May as well get rid of this commented out old code

micahflee:
Awesome, your plan sounds good to me. I'll try to test in Windows and Mac later today, and if I confirm that it looks good on my end, then I think it's ready to merge.

Issue #795, opened 2018-09-28
Fix test warnings


micahflee:
Note urgent, but when running GUI tests, I see a bunch of warnings about deprecated method names and stuff, like:

```
======================================================== warnings summary =========================================================
/home/user/.local/lib/python3.6/site-packages/pytest_ordering/__init__.py:54: RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
Please use node.get_closest_marker(name) or node.iter_markers(name).
Docs: https://docs.pytest.org/en/latest/mark.html#updating-code
  order = mark.kwargs.get('order')
/home/user/.local/lib/python3.6/site-packages/pytest_ordering/__init__.py:54: RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
Please use node.get_closest_marker(name) or node.iter_markers(name).
Docs: https://docs.pytest.org/en/latest/mark.html#updating-code
  order = mark.kwargs.get('order')
/home/user/.local/lib/python3.6/site-packages/pytest_ordering/__init__.py:54: RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
Please use node.get_closest_marker(name) or node.iter_markers(name).
Docs: https://docs.pytest.org/en/latest/mark.html#updating-code
  order = mark.kwargs.get('order')
/home/user/.local/lib/python3.6/site-packages/pytest_ordering/__init__.py:54: RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
Please use node.get_closest_marker(name) or node.iter_markers(name).
Docs: https://docs.pytest.org/en/latest/mark.html#updating-code
  order = mark.kwargs.get('order')
/home/user/.local/lib/python3.6/site-packages/pytest_ordering/__init__.py:54: RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
Please use node.get_closest_marker(name) or node.iter_markers(name).
Docs: https://docs.pytest.org/en/latest/mark.html#updating-code
  order = mark.kwargs.get('order')
/home/user/.local/lib/python3.6/site-packages/pytest_ordering/__init__.py:54: RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
[...]

PR #805, opened 2018-10-11
Refactor GUI Unit Tests

mig5:
@micahflee this is not for merging, but for checking I'm on the right track with refactoring the tests?

I've only worked on the local tests for the moment.

All of the local tests have been refactored to reuse the common tests.

Still requires running via the `run_unit_tests.sh` script. (Actually they all fire at once with just `pytest`, but since they use the same file name for saving settings, I think this leads to setting collision that breaks tests. I guess each test could have its own uniquely-named json settings file... I tried this but no settings file was being written :/ maybe more is needed to let them run in parallel...). 

I am a bit overwhelmed with that object inheritance stuff and might need help getting it over the line.

But at least they are all using the 'common' test assertions now, no real duplication. And they still pass.

Maybe you want to fork this branch and fix some things up.. I'm sure it's not perfect but maybe it's a start..


mig5:
Actually what causes them to run seemingly 'in parallel' is the pytest ordering... they are all order=1, order=2 so they run at the same time..

I hit some collision with the persistent slug test relating to its use of a global slug for comparing.. the only way I find I can avoid that is if I order all the tests sequentially (through all tests in all files) - not ideal..

mig5:
Cleaned up the tests to use proper inheritance:

Receive tests inherit from GuiReceiveTest object
Share tests inherit from GuiShareTest object

Both GuiReceiveTest and GuiShareTest inherit from GuiBaseTest (where the 'common' tests reside)

I still get a segfault when running `pytest` in the `tests_gui_local/` dir (when they all run simultaneously). Also can't move them into the `tests` dir because the `conftest.py` is different for the CLI tests vs GUI tests.

mig5:
More refactoring! The tests can now be run with `xvfb-run pytest tests_gui_local/`.

Still can't be merged into the `tests` dir as the existing tests there don't like the change to `conftest.py`.

mig5:
aaaand of course it works on my machine but segfaults on Travis *sigh*

mig5:
I have now moved all the GUI tests into a 'tests_gui' dir (since I've been unable to resolve the issue whereby its conftest.py breaks the original CLI tests. Maybe you know better than me what to do there; see if you can move all these into `tests`/ and then run `pytest`...).

I have refactored all the Tor tests to use TorGuiBaseTest and related child objects (TorGuiReceiveTest and TorGuiShareTest) which themselves inherit from GuiBaseTest, GuiShareTest and GuiReceiveTest for most of their functions, overriding only what they need to.

I have introduced the `--runtor` flag in pytest to run the tor tests - without it, `pytest tests_gui/` skips them and only runs the local tests as intended.

And on that: `pytest tests_gui/` works on my Debian 10 machine. But it segfaults in Travis. So, for now, I have reinstated the bash script for the sake of Travis. But would be good to fix properly. I instinctively feel as though we should no longer be using Trusty in Travis since we're not usually developing off of it.. I don't know what other options there are. I wonder if we can even switch to CircleCI or something or even our Jenkins rather than use Travis? If it integrates with Github decently? I'll even build Docker images for running those tests on our Jenkins if we need to (pr maybe we can just run them natively and clean up after) - Travis has been a source of frustration for some time... nonetheless, maybe there's an 'easy' fix for this segfault issue that I've not found.

mig5:
OK, I modified the pytest-ordering stuff so that they run more sequentially now and this passes in Travis when running `pytest tests_gui/`.

Given that per your request we are 'grouping' common tests under parent functions (technically considered 'bad' in the Python world where they prefer each test to be its own function with one assertion each), we could maybe do away with having 2 to 3 'test functions' in each test file and just have the one test function. Then we would not even need pytest-iordering at all.

https://travis-ci.org/micahflee/onionshare/jobs/440857136

mig5:
Ha! Think I fixed the test_onionshare_strings.py issue with the changed conftest.py (not sure I fixed it the 'right' way but the strings needed the settings object loaded properly), and have now moved all the tests into the `tests/` dir!

Travis still passing.

I indirectly encountered #807 here and apparently fixed it to get these to pass, but not sure it was the right way or not..


Meanwhile - wow, 92% coverage of the History and ReceiveMode GUI!

```
----------- coverage: platform linux, python 3.6.6-final-0 -----------
Name                                               Stmts   Miss  Cover
----------------------------------------------------------------------
onionshare/__init__.py                               138    129     7%
onionshare/common.py                                 157     21    87%
onionshare/onion.py                                  311    257    17%
onionshare/onionkey.py                                55     39    29%
onionshare/onionshare.py                              47      3    94%
onionshare/settings.py                                71      5    93%
onionshare/strings.py                                 23      0   100%
onionshare/web/__init__.py                             2      0   100%
onionshare/web/receive_mode.py                       174     28    84%
onionshare/web/share_mode.py                         209     43    79%
onionshare/web/web.py                                139     27    81%
onionshare_gui/__init__.py                            70     45    36%
onionshare_gui/mode/__init__.py                      150     39    74%
onionshare_gui/mode/history.py                       303     23    92%
onionshare_gui/mode/receive_mode/__init__.py          75      6    92%
onionshare_gui/mode/share_mode/__init__.py           200     28    86%
onionshare_gui/mode/share_mode/file_selection.py     219     42    81%
onionshare_gui/mode/share_mode/threads.py             27     15    44%
onionshare_gui/onionshare_gui.py                     307     86    72%
onionshare_gui/server_status.py                      214     28    87%
onionshare_gui/settings_dialog.py                    720    681     5%
onionshare_gui/threads.py                             35     17    51%
onionshare_gui/tor_connection_dialog.py               84     46    45%
onionshare_gui/update_checker.py                     121     91    25%
onionshare_gui/widgets.py                             30     23    23%
----------------------------------------------------------------------
TOTAL                                               3881   1722    56%
```

Feeling like this PR is actually in a pretty good place now. Let me know what you think of all the refactoring and if it was where you wanted it to go.

mig5:
I made it so each test object has just the one `test_gui()` method. I don't consider this all that bad, given that the methods being called are already 'grouped' tests/assertions already (almost always `self.run_all_common_setup_tests()` followed by either the `self.run_all_share_mode_tests() or `self.run_all_receive_mode_tests()`, or sometimes a subset of those wrapper methods).

This solves some bugs I encountered with pytest-ordering when there were several methods using `pytest.mark.run(after='some other method')`. 

It in fact means we don't need to depend on pytest-ordering at all. Therefore, it probably 'fixes' #795 too.

mig5:
Tests will fail now because I uncovered a bug (and wrote a test for it). #810 would need merging in for them to pass.

mig5:
There are a huge number of tests here now - it's hard to get the Tor ones to run all together without at least one failing (due to Tor timeouts etc) but the overall coverage is now 72% or higher. ReceiveMode 
has only 1 function not covered.

The local-only tests now cover 70% alone.

There's still more to do such as:

1) Fix the teardownClass - the files are not being cleared up for some reason except for /tmp/test.txt (but not /tmp/OnionShare), and I've ignored clearing up the json settings files for each test

2) Change all the /tmp stuff and adding of /etc/hosts so that it might work on Windows

micahflee:
Wow, this is really cool. I've only skimmed the code so far, and I tried running the tests. I love how it looks in GNOME shell lol.

screenshot from 2018-10-16 22-20-00

This is looking very good. I've been busy, but I'll take a deeper look soon and offer a more solid review, or maybe even try making some commits.

Oh also, it's a bit absurd, but for some reason Ubuntu 14.04 (4.5 years old) is the newest linux that Travis offers, at least as far as I can tell (here's the [linux build options](https://docs.travis-ci.com/user/reference/overview/#sudo-enabled)). I'm not at all attached to sticking with travis. Circle-CI or Jenkins would be great, in my opinion, especially if they could be running like, Debian stretch or buster, or a newer Fedora.

mig5:
Hehe - be careful with running the tests in your actual desktop session like that, because they can pop up into focus and you can accidentally click a button if you were on your way to clicking something else on your computer.

I often run them through xvfb-run just to get them out of the way. But yeah sometimes it's fun to watch :)

mig5:
For some reason both Travis and CircleCI hate one of my tests now, despite liking it earlier.. meh. They all pass on my machine.. will look into it another time.

mig5:
I have the tests passing at CircleCI now.

https://circleci.com/gh/mig5/onionshare/33

So we could switch to this with an 'official' CircleCI rather than this one which can only use my fork.

micahflee:
Nice! Is this ready for final review?

mig5:
Yes, please go ahead when you have time!

Issue #461, opened 2017-08-02
Add support for prop 224


Baccount:
The slugs can be removed because the The first layer of HS descriptor encryption

2.5.1. First layer of encryption [HS-DESC-FIRST-LAYER]
 The first layer of HS descriptor encryption is designed to protect
   descriptor confidentiality against entities who don't know the blinded
   public key of the hidden service.

Here is the design document for Tor's next generation onion service.
https://gitweb.torproject.org/torspec.git/tree/proposals/224-rend-spec-ng.txt

Baccount:
The next generation onion service is expected on December 15th, will OnionShare support before that date? https://trac.torproject.org/projects/tor/wiki/org/teams/NetworkTeam/CoreTorReleases

mig5:
I think OnionShare depends on Stem library for creating the hidden service, so it would depend on whether Stem supports prop224 (which I am struggling to find out).

Baccount:
Opening a ticket on https://trac.torproject.org/projects/tor/report and asking the dev's may be the easiest option.

Baccount:
If I open TorBrowser and have onionshare connect to it, it gets stuck at please wait and never creates the hidden service.

Baccount:
I opened a Ticket asking if stem supports prop 224.
https://trac.torproject.org/projects/tor/ticket/24205#ticket

Baccount:
dgoulet Replied with
“I'm sure it will at some point but for now it can't for ephemeral services until 0.3.3 because prop 224 onion services don't have control port support (#20699).
I would think that Stem can possibly support the creation of persistent service for now. We'll see what atagar decides ;).”

https://trac.torproject.org/projects/tor/ticket/24205#comment:1
@mig5

micahflee:
Yes, I'd love to support this, but it's blocking on stem, which is blocking on prop224 not supporting the control port yet https://trac.torproject.org/projects/tor/ticket/20699.

I think it's less work overall to just wait until the dependencies support the features OnionShare needs before starting work on this. However, as soon as early alpha/dev versions are available to test, I'd like to try to start adding support for OnionShare so it can be ready when the features are stable.

Baccount:
The control port ticket "20699" has been set to merge_ready
@micahflee

micahflee:
[...]

PR #806, opened 2018-10-12
Change how v3 onion services work to support Debian Stretch

micahflee:
@maqp [pointed out](https://github.com/micahflee/onionshare/issues/461#issuecomment-427768795) that stem 1.7.0 was released so I took a look at how it worked. Their [changelog](https://stem.torproject.org/change_log.html#version-1-7) says:

>Documented v3 hidden service support (ticket 25124, spec)

They updated the documentation for how to use the `create_ephemeral_hidden_service` method. It turns out, had we been calling that method in a different way even with older versions of stem, like 0.5.3 that comes with stretch, we could have gotten v3 onion services working. This means we can generate v3 onion keys without need python 3.6 -- it works in python 3.5.3, which is what comes with stretch.

This branch supports v3 onion services, and also runs in stretch with packages only installed from the package manager, nothing from pip. It removes the python3-nacl and python3-cryptography dependencies (python3-crypto is still required though). 

Before merging, we should decide if we really want to yet. Here are the downsides:

- We don't generate keys ourselves anymore, which means it takes longer to start an onion service (I think)
- If we want a "dead man's switch" in the future, in this branch we'll have to briefly create an actual onion service just to generate a key, and then stop it (however, we can pass `await_publication=False`, so I think we can cancel it right away before it's even ready, but with access to the key immediately).

But here are the upsides:

- Runs on Debian Stretch, which means OnionShare 2.0 can be run in Tails and Whonix
- We don't need to worry about cryptography implementation bugs, because we're not generating keys ourselves anymore
- Less dependencies for OnionShare
- Onion service-related code is less complex, and we don't need a separate onionkey module anymore

There's still one more bug. I'm trying to add python 3.5.3 to the travis tests, but running into an annoying pytest-qt related bug. The local GUI tests run fine in a stretch VM for me though, but the travis tests aren't running in stretch  (they're running in ubuntu 14.04, and most of the dependencies are installed with pip and not the package manager).

Also, when you review this, can you make sure that I didn't introduce any regression bugs related to checking "use legacy", or running it with a tor too old for v3 onion services?


mig5:
Well, shit. I *did* know that I could generate onions with the older Stem with no problem, but *something* made us write our own key gen approach.

Now that I think about it, I think it related to persistence: the key that we get *back* from Stem is, or was, not the actual key we can reuse later. E.G I think we got the binary key back, or something like that. Would need to dig back into old tickets. Let's deal with that when we come to it.

In the meantime, there is indeed a regression when using Persistence and Legacy mode:

```
  File "/home/user/git/onionshare_stem17/onionshare_gui/threads.py", line 54, in run
    self.mode.app.start_onion_service()
  File "/home/user/git/onionshare_stem17/onionshare/onionshare.py", line 87, in start_onion_service
    self.onion_host = self.onion.start_onion_service(self.port)
  File "/home/user/git/onionshare_stem17/onionshare/onion.py", line 444, in start_onion_service
    if self.is_v2_key(key_content):
TypeError: is_v2_key() takes 1 positional argument but 2 were given
Aborted
```

Looks like an easy fix. Will continue to review soon

onionshare/onion.py
```diff
@@ -458,29 +457,33 @@ def start_onion_service(self, port):
...
+        self.common.log('Onion', 'start_onion_service', '{}'.format(debug_message))
+        await_publication = True
         try:
             if basic_auth != None:
-                res = self.c.create_ephemeral_hidden_service({ 80: port }, await_publication=True, basic_auth=basic_auth, key_type=key_type, key_content=key_content)
+                res = self.c.create_ephemeral_hidden_service({ 80: port }, await_publication=await_publication, basic_auth=basic_auth, key_type=key_type, key_content=key_content)
             else:
                 # if the stem interface is older than 1.5.0, basic_auth isn't a valid keyword arg
-                res = self.c.create_ephemeral_hidden_service({ 80: port }, await_publication=True, key_type=key_type, key_content=key_content)
+                res = self.c.create_ephemeral_hidden_service({ 80: port }, await_publication=await_publication, key_type=key_type, key_content=key_content)
```
mig5: on line 497, change

```
self.settings.set('private_key', key_content)
```

to

```
self.settings.set('private_key', res.private_key)
```

mig5:
FYI I found the old discussion - we knew already that we could generate a v3 onion with existing Stem and a new-enough Tor version that supports v3. The issue was that when using persistent mode, the value of `private_key` in the Stem response was in a 'base 64 blob' that could not be re-used as a 'pre-prepared key' to start a new onion with that key.

https://github.com/micahflee/onionshare/issues/461#issuecomment-360963786

>  for some reason the base64-encoded v3 private key gets rejected with "http status 400 ("Invalid HS descriptor. Rejected.") response from dirserver 'xx.xx.xx.xx:xxxxx'. Malformed hidden service descriptor?" (a REQUEST_REJECTED response). Digging into the source code of Tor's control.c, apparently an issue with decoding, I guess, despite the fact that it is just using the encoded private key that's returned from the control response itself on the first share.. could it be a bug in Tor and how it's encoding the private key in the response, preventing it from being reused?

https://github.com/micahflee/onionshare/issues/461#issuecomment-360967161

> It appears to use a format called "Base64 blob".

https://github.com/micahflee/onionshare/issues/461#issuecomment-360968319

> That's a good find, but what confuses me is that we're just storing the PrivateKey response sent from control.c itself in the response to ADD_ONION. You would expect it to send back the key in a format that Tor itself thinks it can work with (but maybe not!)

> I dare say this could be worked around if you implement the key generation 'by hand' in onion.py like in the other PR, but it's a bit weird how it doesn't already work..

... and that's what led to @maqp writing the onionkey stuff.


So, the issue I foresee if we go back to this method, is that once Persistence is more practical for v3 (that is not yet practical because of the rev counter bug that we are waiting on 0.3.5 release to fix), will we even be able to do Persistence, if the `res.private_key` Stem response is a useless value (maybe it can be derived back to the format we require, but couldn't figure it out at the time).

maqp:
[This Tails issue](https://labs.riseup.net/code/issues/14792) says:

>Right now tons of packages transitively depend on python3.5, and only pitivi (transitively) depends on python3.6. Let's not bother and come back to it later.

So far there has not been enough pressure to move into Python3.6. Upgrading to Python3.6 would be beneficial because [PEP 524](https://www.python.org/dev/peps/pep-0524/) improved security of `os.urandom()` by blocking until the system has enough entropy. This is especially important for live distributions like Tails that AFAIK do not retain collected entropy across sessions. Thus, if OnionShare wants to pre-generate private keys using Cryptography/PyNaCl, Python3.6 is beneficial. Also, f-strings are beautiful but 3.5 forces the use of `.format()` method across code run on Tails.

In [this issue](https://github.com/pyca/cryptography/issues/3509) atagar asked pyca to introduce support for Ed25519 to allow removal of PyNaCl as a dependency. So it looks like Tor/Stem will depend on Cryptography in the future.

Also, it would be beneficial for smaller projects (like TFC) if OnionShare could help push dependencies to Tails that would otherwise require manual installation for every session. Currently the installation takes several minutes.

For OnionShare, Cryptography is currently only needed for v2 Onion Service key-blob generation, and until Ed25519 is added to Cryptography, PyNaCl is needed for v3 Onion Service key-blob generation.
As soon as pyca adds bindings for X448 to Cryptography, I can switch to using the library for all (a)symmetric crypto in TFC. If Stem/Tor will need cryptography's Ed25519 like atagar wrote above, the dependency might be needed for Stem anyway, and OnionShare could use the single library for both v2 and v3 key blob generation. So I hope at least Cryptography remains in use, and that OnionShare can bump the requirement so that even if Ed25519 is released before X448, the support makes it into Tails.

micahflee:
The main purpose of the key generation code is to make persistence work in v3 onion services, but even if we do generate our own keys, we still don't allow v3 persistence because it's not stable in tor yet, so it doesn't (yet) achieve that purpose anyway. On the other hand, if we generate keys using stem, we can still support stretch. (I wouldn't be opposed to adding back the key generation stuff in version 2.1, or whenever, if it means we'll be able to support v3 persistence.)

@maqp I'd love to push Tails to switch to python 3.6, but I'm not sure making OnionShare 2.0 require it will have any affect on that. Tails is _still_ only including OnionShare 0.9.2, released almost 2 years ago, despite that fact that OnionShare 1.3.1 has been out for 4 months, and 1.3 has been out for 8 months. Until they make a new major release that switched from stretch to buster, Tails will just continue using python 3.5.3, and will only include (at most) OnionShare 1.3.1, without v3 support and without receive mode.

Tails has a policy of only including software by default that's already packaged in Debian, including Debian Backports. The only exceptions they make are core Tails packages they write themselves. I think, if OnionShare supports stretch, and we can get 2.0 into Debian Backports, we have a chance of getting it into Tails without having to wait like a year for them to switch to buster (at which point they'll be switching to python 3.6).

mig5:
Tested on Fedora 28 (0.3.4.8), Debian 10 (0.3.4.8) and Debian 9 (0.2.9.15, pre-v3) without issues (with my patch). Looking good if we want to introduce this (I'm a bit conflicted - Tor 0.3.5 is due in December, maybe even less than 2 months away, at which point we'd want to introduce persistence for v3, and I'm pretty sure we'll need to pre-gen the keys again for that, though we can see if I'm wrong about that when the time comes.. but if not, for 2 months, is it worth it? But I also agree it's nice having less crypto dependencies/custom stuff in the codebase where we can..)

I hit that same pytest bug some weeks ago but could not fix it (it went away of course once we depended on 3.6).

maqp:
>I wouldn't be opposed to adding back the key generation stuff in version 2.1, or whenever, if it means we'll be able to support v3 persistence.

Wonderful, thank you! I don't mind my stuff waiting for Buster. Indeed, OnionShare 0.9.2 should be upgraded to 2.0 at the earliest convenience. Hopefully Tails developers will upgrade from 2.0 to 2.1 more quickly than they upgraded from 0.9.2. So to summarize and confirm if I understood correctly

OnionShare 2.0 for Stretch via Debian Backports
* Python3.5
* Tor 0.3.4
* Stem 1.6 (or 1.7 if it's in OnionShare requirements)
    - v2 persistence

OnionShare 2.1 for Buster with

* Python3.6
* Tor 0.3.5
* Stem 1.7
* Cryptography 2.3.1
    - Optional: RSA keygen for v2 persistence
    - Ed25519 for v3 persistence
    - if available, a version with x448 (probably 2.4)

mig5:
So I experimented with merging develop into this branch and seeing if it made a difference using CircleCI instead of Travis for the 3.5.3 container - of course, it didn't make any difference at all.

I think the annoying issue with Python 3.5 and pytest/pytest-qt is a bug in Python 3.5.

Investigating led me to https://github.com/pytest-dev/pytest-qt/issues/218, which led me to the full thread starting from https://www.riverbankcomputing.com/pipermail/pyqt/2018-July/040492.html, which suggests the bug is in https://docs.python.org/3/c-api/slice.html#c.PySlice_GetIndicesEx

The original reporter in the Riverbank thread simply gave up and dropped 3.5 support for their tests.

I would obviously not want to do that (if we want to reintroduce support for Python 3.5, we probably want to make sure the tests work on it?) but I can't find a workaround.. what should we do?

mig5:
Just as I write that, I figure it out. We can use a Python 3.5.6 container on CircleCI and it works (whereas 3.5.3 fails)

I also figured out how to run the tests on multiple versions properly, in CircleCI.

Is it OK if I push some commits to this branch and we can observe the results of the tests?

micahflee:
Sorry @mig5, yes it's definitely ok for you to push commits to this branch to check on CircleCI.

micahflee:
Yay, tests are passing. @mig5, anything left before this can be merged?

PR #823, opened 2018-11-26
Enable macOS sandbox

### LINKED ISSUE #406: Turn on macOS sandbox
If possible, I'd love to enable the build-in macOS sandbox. Here are some resources about it:

-
 https://developer.apple.com/library/ios/documentation/Miscellaneous/Reference/EntitlementKeyReference/Chapters/EnablingAppSandbox.html
- https://github.com/electron/electron/blob/master/docs/tutorial/mac-app-store-submission-guide.md

Here's the pertinent part from Apple's [About App Sandbox docs](https://developer.apple.com/library/content/documentation/Security/Conceptual/AppSandboxDesignGuide/AboutAppSandbox/AboutAppSandbox.html):

> By limiting access to sensitive resources on a per-app basis, App Sandbox provides a last line of defense against the theft, corruption, or deletion of user data, or the hijacking of system hardware, if an attacker successfully exploits security holes in your app. For example, a sandboxed app must explicitly state its intent to use any of the following resources using entitlements:

And specifically, it mentions this as resources:

> User Files (Downloads, Pictures, Music, Movies, User Selected Files)

For the sandbox to work, we're going to need to allow OnionShare to access "User Selected Files" -- this means that making sure that PyQt5's drag and drop of files, as well as the file selection dialogs, work.

micahflee:
Will resolve #406 when it's complete. I've been making good progress, it's not quite done yet. So far I enabled the sandbox, and had to make these exceptions:

- Allow network servers (binding ports) and clients (connecting to sockets). Tor needs to do both, and OnionShare needs to bind a port for the http server, and connect to sockets for connecting to the Tor control port and the the Tor socks5 port
- Read/write user-selected files. These are files/folders selected in a file dialog, or dragged into the window. We need read access for share mode, and if the user clicks Browse and selects a folder for receive mode, we need write access to that.
- Read access to `/private/etc/apache2/mime.types` because apparently Flask needs this when starting a server
- Read access to `~/Library/Application Support/TorBrowser-Data/Tor/control_auth_cookie`, which makes it so OnionShare can still connect to Tor Browser's tor if we want
- Read/write access to `~/OnionShare`, which is necessary because the user never selects this folder.

I've also had to change a few things around.

I had to make it so in macOS, OnionShare connects to bundled Tor using a port instead of a unix socket file, because the only path on the filesystem I could write to was in a container and had a very long filename, and unix socket filenames can't be longer than 100 chars.

I also made it so the Tor data directory is now always in the OnionShare data directly (where `onionshare.json` is), e.g. in `~/.config/onionshare/tmpfilename` instead of in `/tmp/tmpfilename`. Now in macOS with a sandbox, this folder is `~/Library/Containers/com.micahflee.onionshare/Data/Library/Application Support/OnionShare`. (I wonder how much trouble it is to change the application bundle ID to `org.onionshare` in the future...)

I tested and share mode and receive mode both work. Saving and loading settings appears to work too, because now the settings file is in the sandboxed homedir.

What I still need to do is fix the "Add" button in share mode, which opens a file dialog and lets you select files and folders to add. When you click it, you end up in the sandbox's homedir, not the user's real homedir. This is because if you want the user to be able to see their actual homedir, you have to use a native file dialog, but this file dialog specifically is not a native one. In `onionshare_gui/widgets.py`, here's the `AddFileDialog`:

```python
class AddFileDialog(QtWidgets.QFileDialog):
    """
    Overridden version of QFileDialog which allows us to select folders as well
    as, or instead of, files. For adding files/folders to share.
    """
    def __init__(self, common, *args, **kwargs):
        QtWidgets.QFileDialog.__init__(self, *args, **kwargs)

        self.common = common
        self.common.log('AddFileDialog', '__init__')

        self.setOption(self.DontUseNativeDialog, True)
        self.setOption(self.ReadOnly, True)
        self.setOption(self.ShowDirsOnly, False)
        self.setFileMode(self.ExistingFiles)
        tree_view = self.findChild(QtWidgets.QTreeView)
        tree_view.setSelectionMode(QtWidgets.QAbstractItemView.ExtendedSelection)
        list_view = self.findChild(QtWidgets.QListView, "listView")
        list_view.setSelectionMode(QtWidgets.QAbstractItemView.ExtendedSelection)

    def accept(self):
        self.common.log('AddFileDialog', 'accept')
        QtWidgets.QDialog.accept(self)
```

Specifically, this one runs `self.setOption(self.DontUseNativeDialog, True)` because this was the hack we had to do to make it so a file dialog could select both files and folders, which let us get rid of two separate add buttons for files and folders.

So I think I'll either need to figure out how to select both files and folders with a native dialog, or I'll have to make two add buttons, one for selecting files and one for selecting folders. I really don't want to do the latter, so I hope the former is possible.

Finally, I think it needs to be tested thoroughly. Enabling the sandbox is part of the `codesign` command that digitally signs the app bundle, which means no one will be able to make their own sandbox app bundle to test with without a signing key. So I think I'll make a special test, code-signed "release" just to test this PR before merging.

When running the codesigned, sandboxed OnionShare, you can open the Console app and filter for "onionshare" or "sandbox" to see OnionShare sandbox errors.


micahflee:
I discussed the UX situation with the "Add" button with @glenn-sorrentino and we basically decided this. Since it doesn't seem possible to have a native file dialog window that lets you select both files and folders, the Mac version of OnionShare I'll make two separate buttons, "Add Files" and "Add Folders". But the Windows and Linux version will keep the single "Add" button.

So I'll implement that in this PR, and then I think it'll be ready for review.

micahflee:
Note: After #839 is merged into develop, I need to merge develop back into this branch and refactor `Web.debug_mode` to use the new `common.build_data_dir` method, to avoid duplicating code.

micahflee:
This PR is currently blocked by #846

micahflee:
I merged the PR that fixes #846 into this branch, so I removed the blocked flag because now it works.

I also bumped the version to `2.0.dev1` and wrote a changelog, because I have to make a dev release before anyone can test this, and I wanted to sign a git tag for the dev release (the tag is `v2.0.dev1`).

mig5:
When I start the packaged 2.0.dev1 OnionShare, I get an Alert dialog with the message:

```
2

Try changing how OnionShare connects to
the Tor network in the settings.
```

I recognise '2' as being the return code from Stem probably.

But when I go to the Settings Dialog and just click 'Save' (still using the bundled Tor mode), it connects just fine and I can use OnionShare. File/Folder add dialogs are working fine.

After closing OnionShare and opening it again, I don't get the error :confused: 

In Console I see some sandbox error messages such as 'unable to obtain configuration from file:///Library/Preferences/com.apple.ViewBridge.plist due to Error Domain..... couldn't be opened because you don't have permission to view it'. Unsure if important. Doesn't seem to impact operation otherwise, not sure if it explains the '2' error code above.

mig5:
The '2' error above may have been due to an existing onionshare.json file... may not be important.

So far so good other than the sandbox error above, if you care about it..

mig5:
I see some other sandbox messages `deny(1) file-read-xattr /Users/miguel/xxxxx` to files in my home dir. Again doesn't seem to have any impact in terms of OnionShare use.

Receive Mode also seems to be working fine. I will do some more testing tomorrow hopefully, but so far seems good. If you don't care about this and the above plist sandbox message, you could probably merge.

micahflee:
The alert was totally because of your `onionshare.json` file. I've gotten it before, but I don't think users who upgrade will.

I think it's fine if the sandbox is denying things so long as it doesn't break any functionality. It's kind of cool actually, it means the sandbox is working and strict.

Baccount:
@micahflee "attempt auto configuration with Tor Browser" option does not work in the Sandbox enabled branch. Tested with TorBrowser 8.0.4
`2018-12-19 09:32:29.404 onionshare-gui[22140:1292419] unable to obtain configuration from file:///Library/Preferences/com.apple.ViewBridge.plist due to Error Domain=NSCocoaErrorDomain Code=257 "The file “com.apple.ViewBridge.plist” couldn’t be opened because you don’t have permission to view it." UserInfo={NSFilePath=/Library/Preferences/com.apple.ViewBridge.plist, NSUnderlyingError=0x7ffc4a89a900 {Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted"}}
`

micahflee:
@Baccount odd, I can't reproduce this issue. I have Tor Browser 8.0.4 open, and here's a screenshot of 2.0.dev1 successfully connecting to it.

screen shot 2018-12-19 at 11 51 12 am

In `install/macos_sandbox/parent.plist` I specifically give the sandbox access to the Tor Browser's `control_auth_cookie` file:

```xml
<!-- For OnionShare to be able to connect to Tor Browser's tor control port,
     it needs to read it's control_auth_cookie file -->
<key>com.apple.security.temporary-exception.files.home-relative-path.read-only</key>
<array>
	<string>/Library/Application Support/TorBrowser-Data/Tor/control_auth_cookie</string>
</array>
```

@mig5 can you connect using Tor Browser's tor?

mig5:
Connecting using Tor Browser's Tor is working fine for me on Mac with v2.0.dev1

Baccount:
@micahflee I deleted the TorBrowser-Data folder and opened Torbrowser, Onionshare connects with no errors.

micahflee:
Interesting, I wonder what was wrong. In any case, @mig5 if this looks good mark it as approved and I can merge into develop!

PR #831, opened 2018-12-06
[WIP] Fix v3 onion services

### LINKED ISSUE #812: "Stop sharing after first download" doesn't work with v3 onions
Just while using OnionShare (built from the develop branch) to send a file to myself, I'm running into a weird issue. When I try sharing a 899kb jpg to myself, the server stops and the client thinks it's complete, but it only downloaded 250kb. I was able to reproduce this.

micahflee:
This PR will resolve #812.

In the settings dialog, if you're not connected to Tor, it hides the onion settings and shows you a message:

screenshot from 2018-12-05 20-49-34

If you're connected to an old version of Tor that doesn't support v3 onions, it hides the "Use legacy addresses" checkbox:

screenshot from 2018-12-05 20-50-27

And if you're connected to a version of Tor that does support v3 onions, it shows the "Use legacy addresses" checkbox and it's unchecked by default:

screenshot from 2018-12-05 20-54-37

The only thing that needs to change is this line in `onionshare/onion.py` to change the minimum version of Tor that supports v3 onion services.

```python
self.supports_v3_onions = self.tor_version >= Version('0.4.0.0')
```

At the moment, `0.4.0.0-alpha-dev` is the only packaged version that includes the bugfix, but I believe once it's released `0.3.5` will include the bugfix too (and I think it might get backported to the `0.3.4.x` branch too). So let's wait for that before merging this PR -- or, if that's going to be a long time, we could choose to merge this PR as-is, because I'm confident `0.4.0.0` will work.

In the meantime, to test this you can add the [nightly-master debian repository](https://www.torproject.org/docs/debian.html.en) to a Debian/Ubuntu system to install Tor `0.4.0.0-alpha-dev`, where v3 onion services work great.


micahflee:
Actually, the other thing I need to do before this PR is ready is fix the `tests/local_onionshare_settings_dialog_test.py` test.

mig5:
@micahflee just to clarify this:

> And if you're connected to a version of Tor that does support v3 onions, it shows the "Use legacy addresses" checkbox and it's checked by default:

Right now it's *unchecked* by default (as per the screenshot), was that the actual intention e.g a typo above, or do you really want legacy addresses checked by default (in which case, that's a missing bit of code)?

micahflee:
Oh yes, that was a typo, it should be unchecked by default. I definitely want to use v3 onions if they're available.

micahflee:
I'm not sure why, but the following test is failing in CI, apparently while trying to listen on a socket for the web service:

```
tests/local_onionshare_settings_dialog_legacy_tor_test.py::SettingsGuiTest::test_gui_legacy_tor Fatal Python error: Segmentation fault
```

But the new tests all pass locally for me:

```
$ python3 -m pytest tests/local_onionshare_settings_dialog_*
========================================================================= test session starts ==========================================================================
platform linux -- Python 3.6.7, pytest-3.6.4, py-1.6.0, pluggy-0.6.0
rootdir: /home/user/code/onionshare, inifile:
collected 3 items                                                                                                                                                      

tests/local_onionshare_settings_dialog_legacy_tor_test.py .                                                                                                      [ 33%]
tests/local_onionshare_settings_dialog_no_tor_test.py .                                                                                                          [ 66%]
tests/local_onionshare_settings_dialog_v3_tor_test.py .                                                                                                          [100%]

======================================================================= 3 passed in 0.57 seconds =======================================================================
```

tests/local_onionshare_settings_dialog_legacy_tor_test.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
+import json
+import unittest
+import time
+from PyQt5 import QtCore, QtTest
```
mig5: Minor nitpick: the import of `json`, `time`, and the `Qt*` modules are not needed in these new tests - only `unittest`

tests/SettingsGuiBaseTest.py
```diff
@@ -8,10 +10,27 @@
 from onionshare_gui import Application, OnionShare
 from onionshare_gui.settings_dialog import SettingsDialog
 
+
+class OnionStub(object):
+    def __init__(self, is_authenticated, supports_v3_onions=False):
+        self._is_authenticated = is_authenticated
+        self.supports_v3_onions = True
```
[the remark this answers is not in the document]
micahflee: Oh good catch. Yeah I definitely didn't do that on purpose.

tests/local_onionshare_settings_dialog_legacy_tor_test.py
micahflee: Good catch, I removed the unneeded imports

micahflee:
Yay, I fixed the tests! I don't really understand why, but I confirmed that this line was causing the segfault in circleci:

```python
self.gui.qtapp.processEvents()
```

It doesn't appear to be necessary for this tests to actually work though, so I just removed it, and now they pass without segfaulting.

Wanna take a final review?

micahflee:
Oh yeah, and of course there's still the matter of the minimum Tor version number that supports v3 onions, too.

mig5:
LGTM

Yes, again, note that Tor 0.3.5 is/was scheduled for release on December 15  (two days ago) - I would recommend to wait for that before merging this - if 0.3.5 contains the fix, that can be the base version for v3. We could also soon introduce persistent mode for v3 too from then..

https://trac.torproject.org/projects/tor/wiki/org/teams/NetworkTeam/CoreTorReleases#Listofreleases

micahflee:
I know for sure that Tor 0.4.0.0 works with v3 onions, and I'd rather finish #842 and make a beta release of OnionShare 2.0 instead of waiting on Tor to make a release first (even though it could be any day). So I'll merge this into `develop` but keep #812 open until there's a Tor Browser release that includes a version of Tor that fixes the v3 bug, which will most likely be 0.3.5.

PR #844, opened 2018-12-15
updates to es, fr, pt_BR translations that create conflict in weblate

emmapeel2:
this conflicts were generated because i asked for mergin a file and it got more translations while the push request was being reviewed. this will hopefully not happen often if we automate the locales.


emmapeel2:
ok, this has been superseeded now

PR #853, opened 2018-12-20
Make settings dialog checkboxes line up in macOS

### LINKED ISSUE #847: Settings dialog checkboxes don't line up in Mac
The v2 onion service options are offset in Mac, but don't appear to be offset in Linux:

screen shot 2018-12-16 at 6 36 03 pm

micahflee:
Resolves #847

### DIFF
--- a/onionshare/common.py
+++ b/onionshare/common.py
@@ -395,6 +395,13 @@ class Common(object):
             'settings_connect_to_tor': """
                 QLabel {
                     font-style: italic;
+                }""",
+
+            # For some reason, this prevents extra padding around the v2 onion
+            # settings when viewing in macOS
+            'settings_onion_settings': """
+                QWidget {
+                    border: 0;
                 }"""
         }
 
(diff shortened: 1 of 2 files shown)


mig5:
The gap between the checkbox and 'Use', on Persistent/ClientAuth options, is a *tiny* bit wider than that for Public Mode/auto-stop.

The height between those two settings, and between persistent and auto-stop, is also slightly shorter than that between Public mode and auto-stop.

I had to put my face right up to the screen to make sure, as it's barely noticeable, but it might annoy Glenn :) 

It's so hard to tell though that I don't mind LGMT'ing it...

micahflee:
Huh, that's really subtle. And also I have no idea how to fix it, since the checkbox and its label are both part of the same `QCheckBox` object, they don't have separate CSS. (For that matter, I have no idea why setting the `QWidget` border to 0 solves the spacing problem either...)

So I think I'm going to call this done :)

PR #864, opened 2019-01-04
Fix building debian packages

### LINKED ISSUE #803: Fails to build on Debian 9 due to upstream error in dh-python
Sys info

```
$ uname -a
Linux pythagoras 4.9.0-8-amd64 #1 SMP Debian 4.9.110-3+deb9u4 (2018-08-21) x86_64 GNU/Linux
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 9.5 (stretch)
Release:	9.5
Codename:	stretch
```

```
dpkg-buildpackage: info: source changed by Micah Lee <micah@micahflee.com>
 dpkg-source --before-build onionshare-2.0~dev
 fakeroot debian/rules clean
dh clean --with python3 --buildsystem=pybuild
   dh_testdir -O--buildsystem=pybuild
   dh_auto_clean -O--buildsystem=pybuild
E: Please add apropriate interpreter package to Build-Depends, see pybuild(1) for details at /usr/share/perl5/Debian/Debhelper/Buildsystem/pybuild.pm line 199.
debian/rules:7: recipe for target 'clean' failed
make: *** [clean] Error 25
```

This appears to be a bug with `dh-python`: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=828883

It says it was closed, but this seems to be the exact same issue. (and reporting it here for others)

```
$ apt-cache policy dh-python
dh-python:
  Installed: 2.20170125
  Candidate: 2.20170125
  Version table:
 *** 2.20170125 500
        500 http://ftp.us.debian.org/debian stretch/main amd64 Packages
        100 /var/lib/dpkg/status
```

micahflee:
Will resolve #803.

I added `python3` to `Build-Depends`, which makes it so running this doesn't crash early on:

```sh
python3 setup.py --command-packages=stdeb.command bdist_deb
```

However, there's another problem. Now when I run that, it automatically runs the tests, including GUI tests. And because it's not running them with `xvfb-run`, in a desktop environment it pops up tons of OnionShare windows. And also, I'm not sure why, but I believe because of the environment the tests are being run in, many of them fail.

I like the idea of tests passing before building a `.deb`, but the GUI tests take a long time to run, and the popping up windows thing is quite annoying. I wonder if it might make more sense to only run the non-GUI tests? I don't know how to tell `pybuild` what tests to run -- it appears to just execute this as part of the build:

```
I: pybuild base:217: cd '/home/user/code/onionshare/deb_dist/onionshare-2.0~dev2/.pybuild/cpython3_3.7_onionshare/build'; python3.7 -m pytest tests
```

Right now running the tests runs the local GUI tests, but you need to pass `--runtor` to run the Tor GUI tests. We could change the tests so that you need to pass `--rungui` to run any of the GUI tests, and otherwise it doesn't run them. @mig5 what do you think about that solution?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
   test-3.5: &test-template
     docker:
       - image: circleci/python:3.5.6
-      
+
     working_directory: ~/repo
 
     steps:
@@ -33,7 +33,7 @@ jobs:
       # run tests!
       - run:
           name: run flake tests
-          command: | 
+          command: |
             # stop the build if there are Python syntax errors or undefined names
             flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
             # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
@@ -42,7 +42,7 @@ jobs:
       - run:
           name: run tests
           command: |
-            xvfb-run pytest --cov=onionshare --cov=onionshare_gui --cov-report=term-missing -vvv tests/
+            xvfb-run pytest --cov=onionshare --cov=onionshare_gui --cov-report=term-missing -vvv tests/ --rungui
 
   test-3.6:
     <<: *test-template
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
       - run:
           name: run tests
           command: |
-            xvfb-run pytest --cov=onionshare --cov=onionshare_gui --cov-report=term-missing -vvv tests/ --rungui
+            xvfb-run pytest --rungui --cov=onionshare --cov=onionshare_gui --cov-report=term-missing -vvv tests/
 
   test-3.6:
     <<: *test-template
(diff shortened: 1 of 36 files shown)


mig5:
@micahflee I think that makes sense to add `@rungui` markers to all the local GUI unit tests so that they're skipped for package build purposes (and then add `--rungui` to the CircleCi yml)

And then those in the know can also pass `--rungui --runtor` to run the Tor GUI tests.

micahflee:
Ok, I think this is done! @mig5 wanna review it?

.circleci/config.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - run:
           name: run tests
           command: |
-            xvfb-run pytest --cov=onionshare --cov=onionshare_gui --cov-report=term-missing -vvv tests/
+            xvfb-run pytest --cov=onionshare --cov=onionshare_gui --cov-report=term-missing -vvv tests/ --rungui
```
mig5: minor nitpick: it's more readable to have the flag `--rungui` on the left side of `tests/` with the other flags. But it doesn't seem to impact the actual command.

micahflee:
Ok @mig5, I fixed those. Wanna give it a final review?

PR #872, opened 2019-01-20
Stop referring to "download" and "upload"

### LINKED ISSUE #868: Change "upload" and "download" language and icons
Right now, the upload and download language and icons are confusing.

In share mode, the show history icon is a down arrow, and when you click it it shows a big down arrow and says, "No Downloads Yet". When at least one person has downloaded the files, it says "Download History" at the top and "Download Started [date]" next to each time someone downloaded it. For receive mode, the show history icon is an up arrow, before anyone uploads anything it says, "No Uploads Yet", and when there are uploads it says "Upload History" and "Uploaded [date]" by each upload.

In share mode, you're sending files from your computer to somewhere else, so the arrow should be up. And in receive mode, you're getting new files onto your computer, so the arrow should be down. The words "download" and "upload" are basically reversed in OnionShare -- it's using the terms for the Tor Browser user, rather than for the OnionShare user.

Instead, share mode:

- Should use up arrows, because files are leaving your computer
- Should say "No Files Sent Yet" instead of "No Downloads Yet"
- Should say "History" instead of "Download History"
- Should say "Started sending [date]" instead of "Started Downloading [date]"

And receive mode:

- Should use down arrows, because files are coming into your computer
- Should say "No Files Received Yet" instead of "No Uploads Yet"
- Should say "History" instead of "Upload History"
- Should say "Received [date]" instead of "Uploaded [date]"

After making these changes, translators will have to update the translations for these strings before making the 2.0 release.

micahflee:
Resolves #868.

In addition to changing language and icons, it also:

- Finishes removing the REQUEST_CLOSE_SERVER web event type (it used to allow users to close the receive mode server, but we removed that feature -- this finishes removing it)
- Unifies the language in the History widget for both share and receive mode even more, and when a transfer is complete it now also displays the start and end times of the transfer


micahflee:
Thank you! Just made those few minor changes.

PR #873, opened 2019-01-20
Change version of Tor that supports v3 onion services to 0.3.5.7

### LINKED ISSUE #812: "Stop sharing after first download" doesn't work with v3 onions

micahflee:
When finished, this will resolve #812.

I tested that Tor 0.3.5.7 supports v3 onion services and doesn't have the bug where when the setting "Stop sharing after files have been sent" (formerly "Stop sharing after first download") is enabled, share mode doesn't send the full file.

But before #812 can be totally finished, I still want to wait for a Tor Browser release that includes Tor 0.3.5.7, and make the Windows and Mac versions extract this version of Tor to bundle with OnionShare.


micahflee:
Awesome! And sorry I just pushed a commit to this branch before I realized you had already removed it, so rewrote history to un-push that commit. It really should be in a separate PR anyway.

PR #874, opened 2019-01-20
Stopping the server during a receive mode transfer immediately stops the transfer

### LINKED ISSUE #866: Receive Mode Repeatable crash
1. Start receive mode "I tested this bug with Tor 0.3.5.7"
2. Using TorBrowser go to the hosted V3 onion.
3. Upload a file, 10Mb+
4. While the file is being downloaded, Click **Stop Receive Mode**
5. Then quickly click **Start Receive Mode** and Onionshare crashes every time.
```
[Jan 07 2019 20:58:37] Onion.cleanup: trying to remove onion 5yym6w3kg4tof56umfy6a632nvtoc5dkkpkhj4zd2slqzp5wuiwml7yd
=> 2.4 MiB         video 1.mov[Jan 07 2019 20:58:37] ReceiveMode.update_primary_action
  File "onionshare_gui/onionshare_gui.py", line 382, in timer_callback
    mode.handle_request_progress(event)
  File "onionshare_gui/mode/receive_mode/__init__.py", line 159, in handle_request_progress
    'progress': event["data"]["progress"]
  File "onionshare_gui/mode/history.py", line 451, in update
    self.item_list.update(id, data)
  File "onionshare_gui/mode/history.py", line 347, in update
    self.items[id].update(data)
KeyError: 0
=> 2.4 MiB         video 1.movAbort trap: 6
```
Video showing the bug. https://streamable.com/gu6xk
@mig5

micahflee:
Resolves #866.

This was trickier than I was expecting to fix. Here's what the PR does:

- Refactor the Web.ShareMode client_cancel variable to be Web.stop_q, a thread-safe queue that communicates to both share and receive mode when the user stops the server. In share mode this still stops sending the file. In receive mode, if there's a transfer in progress, it cancels it in the middle, and doesn't end up saving that file
- In receive mode, make the receive mode dir right before saving a file (so if it doesn't complete, don't make an empty dir)
- Minor UX tweak: resizing the window stretches the History widget first


micahflee:
@Baccount would you mind testing this to let me know if it fixes the bug you found? (It should.)

And @mig5 do you mind reviewing?

Baccount:
@micahflee The crash still occurs on my end using your latest branch. Here is the traceback
```
[Jan 20 2019 19:12:47] Onion.start_onion_service: key_type=NEW, key_content=RSA1024
  File "onionshare_gui/onionshare_gui.py", line 382, in timer_callback
  File "onionshare_gui/mode/receive_mode/__init__.py", line 159, in handle_request_progress
  File "onionshare_gui/mode/history.py", line 482, in update
  File "onionshare_gui/mode/history.py", line 378, in update
KeyError: 0
Abort trap: 6
```

micahflee:
@Baccount odd, I just tried several times to reproduce the issue in your bug video but I can't get it to work -- every time I stopped the server, it doesn't crash, and instead it just shows that the transfer was canceled in the history widget.

But still, based on the stacktrace you provided, it looks like the GUI is receiving a `REQUEST_PROGRESS` message and trying to update the UI with progress, when those UI elements don't exist anymore, and hence the crash. So I just pushed a commit that _should_ prevent the `REQUEST_PROGRESS` message from getting sent if the request was already closed (although I don't see why it would be getting sent anyone). But in any case, can you test again with this latest commit?

@mig5 oh yeah, I'm totally bad at adding new GUI unit tests. Hmm, making this test does seem tricky though. I suppose it could send a large file over receive mode, and then click cancel after like 2ms, and hope that the file didn't transfer completely before cancel was clicked. But still, it could be a race condition. Like, what if on the CI server it _did_ transfer in 2ms? I think I'll not worry about writing a test for this one just because of how tricky it is.

Baccount:
@micahflee The latest branch still crashes while following the same steps, but it now gives me a different traceback.
```
 * Running on http://127.0.0.1:17647/ (Press CTRL+C to quit)
  File "onionshare_gui/onionshare_gui.py", line 382, in timer_callback
    mode.handle_request_progress(event)
  File "onionshare_gui/mode/receive_mode/__init__.py", line 159, in handle_request_progress
    'progress': event["data"]["progress"]
  File "onionshare_gui/mode/history.py", line 482, in update
    self.item_list.update(id, data)
  File "onionshare_gui/mode/history.py", line 378, in update
    self.items[id].update(data)
KeyError: 0
Abort trap: 6

```

mig5:
I don't reproduce it on MacOS 10.14.2

PR #888, opened 2019-01-29
Minor settings fixes

micahflee:
This fixes a few issues I found while taking screenshots for weblate.

I noticed that when you have both client authentication and persistence enabled, and you open settings, it crashes because it was trying to access a local variable that should be a class variable.

I also noticed that when using both client auth and persistence, the settings dialog looked pretty bad:

screen shot 2019-01-28 at 8 07 53 pm

It turns out, the button looks bad because of css that's used to attempt to prevent moving the onion settings checkboxes a few pixels in macOS. To solve it, I made a new, separate onion settings group.

Now it looks like this:

screen shot 2019-01-28 at 8 28 25 pm
screen shot 2019-01-28 at 8 28 50 pm


micahflee:
@mig5 mind reviewing this one?

PR #898, opened 2019-02-11
Support persistent v3 onions

mig5:
Please do a lot of testing!

v3 persistent onions ought to work after start/stop/start like 10-20 times. (Prior to 0.3.5.7 they would stop starting back up, at least by 10th time if not earlier)

### DIFF
--- a/onionshare/onion.py
+++ b/onionshare/onion.py
@@ -451,20 +451,9 @@ class Onion(object):
             key_content = self.settings.get('private_key')
             if self.is_v2_key(key_content):
                 key_type = "RSA1024"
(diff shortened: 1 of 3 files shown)


mig5:
(why haven't the CircleCI tests fired?)

micahflee:
I don't know why CircleCI wasn't firing, so I pushed your branch to my repo to see if it would work here, and now it looks like it's firing for you. So... cool, tests pass.

mig5:
Done. Reordering also uncovered an unnecessary extra conditional for legacy mode, so I cleaned that up too.

PR #900, opened 2019-02-12
Receive mode partial download

micahflee:
This PR refactors `ReceiveModeWeb` to make is to files get written to the OnionShare data dir (by default `~/OnionShare`) immediately while the Tor Browser user is sending them, instead of after the transfer is complete. While the file is in progress, it has a `.part` file extension, which gets removed when the transfer is complete.

In order to make this work, much of the receive mode web logic moved from the `upload_logic` flask route into the custom `ReceiveModeRequest` class, as well as the `ReceiveModeFile` class (which used to be called `ReceiveModeTemporaryFile`).

I think this is quite a bit nicer, and lets users watch uploads on the filesystem as they're happening. And if an upload cancels early, the user gets to at least access what had uploaded so far.

### DIFF
--- a/onionshare/web/receive_mode.py
+++ b/onionshare/web/receive_mode.py
@@ -82,7 +82,14 @@ class ReceiveModeWeb(object):
 
             if request.upload_error:
                 self.common.log('ReceiveModeWeb', 'define_routes', '/upload, there was an upload error')
+
+                self.web.add_request(self.web.REQUEST_ERROR_DATA_DIR_CANNOT_CREATE, request.path, {
+                    "receive_mode_dir": request.receive_mode_dir
+                })
+                print(strings._('error_cannot_create_data_dir').format(request.receive_mode_dir))
+
                 flash('Error uploading, please inform the OnionShare user', 'error')
+
                 if self.common.settings.get('public_mode'):
                     return redirect('/')
                 else:
--- a/onionshare/web/receive_mode.py
+++ b/onionshare/web/receive_mode.py
@@ -58,104 +58,37 @@ class ReceiveModeWeb(object):
 
         def upload_logic(slug_candidate=''):
             """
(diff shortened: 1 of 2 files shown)


micahflee:
Please review @mig5.

Also, I checked, and unfortunately this doesn't affect the NoScript bug.

onionshare/web/receive_mode.py
```diff
@@ -214,19 +153,30 @@ def __call__(self, environ, start_response):
...
-        self.f = tempfile.TemporaryFile('wb+')
+        self.filename = os.path.join(self.onionshare_request.receive_mode_dir, filename)
+        self.filename_in_progress = '{}.part'.format(self.filename)
+
+        # Open the file
+        self.upload_error = False
+        try:
+            self.f = open(self.filename_in_progress, 'wb+')
+        except:
+            # This will only happen if someone is messaging with the data dir while
```
mig5: Nitpick: 'messing with', not 'messaging with'

onionshare/web/receive_mode.py
```diff
@@ -214,19 +153,30 @@ def __call__(self, environ, start_response):
...
+            # OnionShare is running, but if it does make sure to throw an error
+            self.upload_error = True
+            self.f = tempfile.TemporaryFile('wb+')
```
mig5: Do you mean to deliberately still use a TemporaryFile as a sort of fallback here, even though you're considering this an upload error? Should it not simply return early?

micahflee: I actually do do that deliberately. The reason is because `self.f` needs to be a file-like object. Originally I was setting to to `None`, but that causes another exception to get thrown later on, when it tries to call `.write` on it (which it will try to call at least once, before it can cancel gracefully).

Issue #903, opened 2019-02-16
In receive mode, when Tor Browser user cancels an upload, the History widget should show it as canceled


micahflee:
Right now, the history widget just says that item was "Transferred", but doesn't show the open file manager button.

mig5:
I couldn't figure this one out. It seems we have all this logic for handling a 'canceled' event, but it is somehow instead being handled as a 'finished' event.

mig5:
If I 'cancel' an upload from the browser (e.g close the tab mid-upload), this bit of code in the `ReceiveModeRequest.close()` function always evaluates to false:

```
                if not self.web.stop_q.empty():
                    # Inform the GUI that the upload has canceled
                    self.web.add_request(self.web.REQUEST_UPLOAD_CANCELED, self.path, {
                        'id': upload_id
                    })
                else:
                    # Inform the GUI that the upload has finished
                    self.web.add_request(self.web.REQUEST_UPLOAD_FINISHED, self.path, {
                        'id': upload_id
                    })
```

The 'canceled' logic never fires. Makes sense, as the app doesn't add anything to your `stop_q` Queue when the browser does something like that.

Issue #902, opened 2019-02-16
Clicking "clear all" in the history widget shouldn't clear currently transferring items


micahflee:
If you start a receive mode server, start an upload, and then click clear all, it clears the currently transferring item (though it's still in the middle of transferring). I think this happens with share mode transfers too.

PR #901, opened 2019-02-13
[WIP] Use Ajax to send files in receive mode to workaround browser bug with large files

mig5:

PR #904, opened 2019-02-16
Implement ajax uploader 

### LINKED ISSUE #899: Receive mode seems to stall at 99% with large files
While working on the 2.0 release, I ran into a bug in receive mode that is reproducible. It only seems to happen with large files (>100mb), but smaller files seem to work fine. When uploading a large file, the progress gets to 99%, and it appears like it's still transferring in both OnionShare and Tor Browser, but it never finishes.

micahflee:
Resolves #899.

This PR does several things:

- If JavaScript is disabled, shows instructions about how to bypass the large file upload bug by either setting security slider to Standard or disabling the Tor Browser XSS setting
- Adds jQuery as a dependency, including adding a new license file
- Add new routes, `/[slug]/upload-ajax` and `/upload-ajax` for public mode, which respond with JSON (but the original non-ajax `/[slug]/upload` should work exactly the same)
- If JS is enabled, has a _much_ nicer UX that let's you upload multiple files at once and cancel files while they're being uploaded

The UX also handles an interesting UX problem. The progress bar tracks the progress of sending data from Tor Browser to the first Tor node, but when that completes you still have to wait for the data to finish traversing the entire onion circuit, before OnionShare responds to the request and the response traverses back.

So when you start an upload, it says "Sending data to initial Tor node" (and also, at this stage, there's a cancel button that lets you cancel the upload):

screenshot from 2019-02-15 19-06-04

After the progress bar gets to 100%, it removes the cancel button and changes the status message to "Waiting for data to finish traversing Tor network ..." (with an OnionShare logo-looking ajax loader animated gif):

screenshot from 2019-02-15 19-10-50

And finally when the upload finishes:

screenshot from 2019-02-15 19-11-26

While working on this, I also discovered #902 and #903, which I think can be fixed in separate PRs.

What's left before this PR is ready:

- There's a bug where if you have multiple progress bars at once, I think they're all getting updated instead of the specific one that's supposed to be getting updated
- When you stop the receive mode server during an upload, the error doesn't display the message clearly


micahflee:
Ok @mig5, I think this is ready for review!

If javascript is disabled, the noscript warning should display, and (assuming the user disabled NoScript's XSS setting) uploads should work fine without javsacript.

And if javascript is enabled, there should be a nice UX for uploading, with progress bars, multiple uploads at a time, etc.

micahflee:
> I left a note, not sure if you wanted to get rid of the `console.log` messages, but it's up to you.

Oh yeah, good idea. I'll remove them.

> A small observance: in #901 your screenshot of the red warning div, was centered on the page. For me it seems left-aligned in this branch.

I changed the css around a bit.I think it looks better centered when there's just one flash message, which is what happens when you submit the form without javascript. But when there are several, like if you have multiple uploads, or upload something and cancel, it looks nicer left-aligned. So I figured I'd just keep it left aligned.

> I also tested filling up my disk and got the expected error (though it seemed to send the data all the way through the Tor network before it then understood there'd be a problem, so it took a while. But same is true of a non-JS upload. The OnionShare UI, on the other hand, prematurely marks the file 'transferred' almost immediately in the History view, but it too waits for the full POST request to complete before it throws its Alert() dialog. This corner case is not a blocker for release though.

Interesting. Yeah, I'm not too worried about this corner case -- even as it stands, it's not that big of a problem, as the errors all get displayed properly, it just wastes some time. I think that when a browser starts making a large HTTP POST request, it has to finish sending the entire request before getting a response -- there's just no way in HTTP to respond to a request before it's finished being sent. So I'm not sure there's really a way around that anyway (that part of it at least).

PR #911, opened 2019-02-19
Resolve #878 UX: Enhance URL copying

### LINKED ISSUE #878: On monitors with poor contrast, the "Copy Address" link is not very visible
screenshot_2019-01-22_09-24-03

This could be made into a button, and also the actual onion address could be set to being selectable.

mgalicer:
resolves #878 
- Style Copy Address as high contrast button
- Make .onion URL text selectable

co-authored-by: @aguestuser
As per @heartsucker's request

button-ux


micahflee:
Thanks for implementing this!

Here's what the "copy address" button looks like before:

Screenshot from 2019-03-10 12-26-26

@holantonela @glenn-sorrentino How do you think this new style of button compares to the old?

glenn-sorrentino:
@micahflee @holantonela I much prefer the text-only version. I'm not opposed to the button style, but we should have a primary, secondary, and tertiary button. The purple one is primary, for sure.

aguestuser:
@glenn-sorrentino : the motivation for implementing this was that (as per @heartsucker's observation in #878), the text-only version was not very visible on low-contrast monitors, and thus presented a usability barrier.

see:

not-visible

the request in that issue was to make a button with higher contrast to eliminate said usability barrier. we chose purple and white because it was the highest contrast combination in the existing color pallete.

if you think that a "secondary" button combination would be better (sounds quite resonable! :)), could you point us to an example of where such a combination exists in the existing codebase? more broadly: is there a general purpose color palette (or style guide around css) that we (or other contributors) could consult?

if so... would be happy to work according to it! if not... would be happy to help pull it together! :)

PS: if the whole button idea is bad... should we just carve out a PR that makes the URL in the text box selectable?
---

cc: @holantonela @mgalicer @micahflee

glenn-sorrentino:
Alternatively, we can just make the blue higher contrast...

micahflee:
Oh yeah I kind of like that idea, making the shade of blue higher contrast.

aguestuser:
So... I'm not at *all* a designer and don't have any real skin in the game here (was just dropping by to try to be helpful to a project I really admire by way of introducing my friend @mgalicer to a cool project as she shopped for OSS things to contribute to). 

But I must confess to being a bit confused as to apparent preference for plain blue text being expressed by maintainers. (Read: I'm not saying I think you're wrong or something. I'm just genuinely confused and would like to try to learn!) 

By way of helping me learn: could y'all explain what design principles or project priorites make you prefer it?

To explain my confusion a bit:

"Copy link" is a clickable action. Almost everywhere else in the "clickable action" affordance is indicated by a purple button...

See downloading from the onion in send mode:

onionshare-download

Or uploading a file to the onion in receive mode:

send-files

As an outsider to the project, plain blue text (whether it's darker or not) does not (to my subjective eye) indicate "click me and an action will happen." Perhaps I *would* have such an intuition if the blue text acted like a hyperlink (for example if it had a pointer cursor when I moused over it, etc...).

But even then, it would seem odd that the project chooses "hyperlink for action" in this one case, but buttons (be they primary purple or green/red for start/stop) almost everywhere else.

Maybe there's something that I'm missing? @glenn-sorrentino : it seems as if you are a designer? Could you help me better understand the thinking here?

Thanks for your patience in looking at this and taking the time to converse. Really do admire the project and am very grateful for all of your hard work on it. :)

holantonela:
Hola! What if we convert that critical link in a button? 

copy url

Since the app width is fixed, we may have estimated space to have a button there.

mgalicer:
@holantonela that seems like a nice solution.

aguestuser:
+1! @holantonela : that looks great! :)

preferences on purple or blue for the button?

mig5:
Hi,

I like that button being alongside the URL ! However there could be a catch (sorry to throw a spanner in the works).

When Stealth Mode is enabled, there is a second button 'Copy HidServAuth' alongside the Copy Address button. See the screenshot below.

Screenshot_2019-03-13_07-30-07

Not sure if you would be able to put both inside the URL label like your above mock-up? It could work though..

The other thing is, when you click the History toggle button in the top right of OnionShare, a second column appears on the right. This shrinks the address bar.

Screenshot_2019-03-13_07-32-18

The next-gen onion addresses are very long, so it wraps. Some thought would need to go in to whether the button(s) look OK in a wrapped view like this.

Personally I actually didn't mind the original idea with the purple buttons beneath the onion URL, but just my 2c :)

micahflee:
@holantonela I like the blue button you designed, and it does look good within the URL field.

But like @mig5 pointed out, the window isn't actually fixed-width (you can resize it as much as you want or maximize the window, and opening the history shrinks that field) and if the user chooses to make an authenticated onion service (which right now is only supported for v2 onion services) then there will be two buttons.

@mgalicer could you make the button still look like a button, but make it blue instead of purple? Then I'll test and make sure it still looks good with normal onion services, and also with authenticated ones.

SaptakS:
@mgalicer @aguestuser are you still able to work on this? If not, i would love to take it over to the finishing line as per suggestions by @holantonela @mig5 and @micahflee

mgalicer:
@SaptakS please, go ahead! 😸

SaptakS:
As discussed between me and @glenn-sorrentino I went ahead and modifed the code to make the buttons look similar to the buttons in the main page of onionshare tabs:
Screenshot from 2020-09-07 20-45-26

cc: @micahflee @mig5 @holantonela

SaptakS:
Resolved conflicts after refactoring of project structure. @micahflee should be ready for review and merge if all looks good.

micahflee (approved):
Looks great!

Issue #916, opened 2019-02-22
Messaging + Page Ordering


ninavizz:
Howdy! I worked on SecureDrop's NoScript workaround, and just saw screenshots of Onionshare's similar workaround. Very different users, so totally dig the general differences—however I did observe one very general ux best practice for the entire product (that I see many FOSS projects do wrong) that I wanted to recommend a change to. I don't code, so I'm just sharing modified screenshots here—on the chance that someone who does code, might take an interest in this. 

User messaging—task progress, success, warnings, or failures—should always be shown at the top of the user's screen, or shown in a visually connected/contained fashion when shown below content. 

**Like this:** 
![image](https://user-images.githubusercontent.com/8262612/53276656-e83f8180-36b4-11e9-88e8-d45a7a2a319f.png)

**...so current activity, error or success messaging, etc., all go here:**
![image](https://user-images.githubusercontent.com/8262612/53276756-63a13300-36b5-11e9-9d93-08b7b1946fa8.png)

**...or alternately, placed here**
![image](https://user-images.githubusercontent.com/8262612/53276686-04dbb980-36b5-11e9-84a3-676d62a90881.png)

**...and all follow the center-alignment of the rest of the page's content, to more naturally guide users to follow the page's hierarchy.**

When contextual messaging is shown below a page object where the user made an error, the messaging needs to pertain to just that one object (which, yes, in Onionshare, is only one thing, anyway). User mental models have been trained to expect the messaging to be paired directly to the functional widget, not the total contents of a page—so that'd be below the "Browse" button w/ adjacent selected file named, but above the page's primary action button to "Send Files."  

The pattern also has the specific object's color changing to match the text of the message—or some other visual device connecting the two objects. Example: you enter the digits for phone number without dashes, but the form needs dashes... so the message stating your error shows in red, below the text field, with the text field's border also changed to red (or have a red outer-glow).  Even on a single-item form, a message sitting below the top of the page still needs to paired just below the widget a user performed an action against, or they'll miss it. 

I appreciate these details may seem fussy, but the patterns have been tested exhaustively—and users who are not deeply-curious, self-starting hacker types, need visual cues and page hierarchies to be clear, or they overlook things that may seem obvious to the rest of us. Today's users just aren't trained to look for messaging below a page form's primary action buttons. That used to be a standard, but it's been moved away from as user research has come to influence page design more and more.

mig5:
Thank you so much for this advice @ninavizz ! It totally makes sense. This is great.

[...]

PR #917, opened 2019-02-23
Make progress, noscript warning, and flashes appear above the form, but beneath the logo

mig5:
A proof of concept per #916


mig5:
@micahflee I've updated with some of these recommendations. 

After making the flash messages concatenate as a comma delimited string, I felt it looked pretty good as-is, so didn't delve into adding a Javascript [x] option (but this is just an excuse for the fact that I'm not confident with javascript :) )

If you think that's still worthwhile, please do fork the branch and add it in...

PR #924, opened 2019-02-25
Update style.css

danielduplex:
Centered "Send Files" button and increased its font size to match other buttons.

### DIFF
--- a/share/static/css/style.css
+++ b/share/static/css/style.css
@@ -52,8 +52,13 @@ header .right ul li {
   border: 0;
   border-radius: 5px;
   text-decoration: none;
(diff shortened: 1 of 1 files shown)


micahflee:
Thanks for submitting a pull request @danielduplex, though I think this one (and also #925) might clash with #917, which is also changing the look of the web server side of OnionShare.

So I'm going to close both of these PRs, but feel free to submit another PR after #917 is done with some more HTML/CSS tweaks. And next time, it will make it easier to review if you put all of your commits in a single PR, and maybe also show a screenshot of the changes that you're making.

PR #933, opened 2019-03-04
[WIP] Add a Startup Timer feature (scheduled start / dead man's switch)

mig5:
I've got startup timer (dead man's switch) working, including displaying the onion address (and slug if applicable) in advance of actually starting up the service.

Technically it very briefly registers the onion via Stem (with `await_publicaion=False`) but never starts the web thread for that. Then when the scheduled time triggers, it starts the onion again (a bit like if it were a persistent onion, with a pre-set private key), but this time with the web thread, actually activating the service.

It works with or without public mode, and works with persistent mode too.

This was quite challenging and involved some refactoring, such as that I need to obtain the slug and the port associated with the onion service in advance, so that I can remember that stuff for when starting the web thread later.

There is a little test for local mode which schedules a share 5 minutes into the future, and is passing. I haven't written a Tor one yet.

There is a lot here and some of the changes might appear strange at first glance. Might be worth a real-time chat some time to walk you through it.


mig5:
Ugh, bugs I didn't anticipate, hooray for tests :)

PR #934, opened 2019-03-04
Add a Startup Timer feature (scheduled start / dead man's switch)

### LINKED ISSUE #571: [Feature request] Add start sharing timer

mig5:
Right, think I have fixed all the bugs, tests are passing locally!

Fixes #571 (just linking the old issue)


micahflee:
As far as writing out dates by spelling out the month, we actually already do this in the History widget. I think it would be fine to use the same date formatting here.

Screenshot from 2019-03-10 15-15-33

mig5:
Hi @micahflee , I think I have fixed all the above. Can you please test again when ready? I also added a test for the startup/shutdown timer clash issue, which now throws an Alert dialog as expected if the times either match or the shutdown is earlier than the startup.

I also realised that I had forgotten about stealth mode ('persisting' the HidservAuth across a scheduled share, and displaying the HidServAuth in advance, like the onion address / slug). That's now fixed too.

mig5:
Fixes #571 (just linking the old issue)

maqp:
@micahflee @mig5 

I'm able to edit the `start share at` time field by typing into it, but it keeps resetting to the initial value. 

It took me a moment to understand that I need to click a field, use `up`/`down` arrows to change values and then maybe use `tab`/`shift+tab` to to move between the fields.

---

The field highlight is not working properly, [at least under the Finnish locale](https://imgur.com/a/hnGsEqM).

---

Common settings should probably be visible during startup timer and sharing:

* (Use of the public mode is already visible from the (missing) slug in the address that's always visible.)

* (Use of v2/v3 address is already visible from the Onion URL.)

* Hovering mouse over the ⓘ 

    * already properly shows information about whether sharing tops or continues after first download.

    * only shows information if the address is persistent. So I'm wondering it should also say `Every subsequent share uses a random address. (To use a persistent address turn on "Use persistent address" in the settings)` when persistent addresses are not enabled.

However, the line with the ⓘ isn't visible until the sharing starts.

### Timers

I think it would be beneficial for the user if the UI was a bit more unified:

**Currently**

Start timer

* shows the date when sharing is supposed to start. However, [the year isn't showing](https://imgur.com/a/dG1pMmE) so if the user sets the share to start e.g. one year from now, it's not possible to view the start time in any way:  This gets problematic if the user accidentally sets the wrong year and the share doesn't start when it should.
* no effect when hovering mouse over the button

Stop timer

* shows countdown in seconds [which is hard to comprehend](https://imgur.com/a/LRSXadu) when share duration is long.
* hovering properly shows when timer stops

**So my proposal is**

Start timer
* Would show count-down timer `sharing starts in 1y 1m 1d 22h 55m 25s`
* Hovering shows the start and stop date 

Stop timer
* Would show count-down timer `sharing stops in 1d 23h 45m 56s`
* Hovering shows the stop date

And finally, slightly shorten the tool tip show duration.

maqp:
To add:

[majority of countries use 24h format](https://i.stack.imgur.com/43dwH.jpg) so it would be beneficial to at least have a setting for it. For date, I think ISO 8601's `YYYY-MM-DD HH:MM:SS` is intuitive enough even though my country uses the `DD-MM-YYYY` format. At the same time it wouldn't be too confusing in countries that tend to write months before days.

>I'm not sure what the use-case for [start+stop timed shares] is

Securely hosted, timed form retrieval and submission would be useful at least. You might want to receive anonymous comments to some article for limited period. Or maybe there's a story you're working on and you want public's experiences for it over some duration of time.

All in all, thank you so much for working on this!

mig5:
Yikes... I'm afraid I didn't follow much of this but thanks for the feedback.

We can't really handle a myriad of different ways to display the time based on the user's cultural background.. whether or not a country typically uses AM or PM, surely this is nonetheless understood

'Common settings should probably be visible during startup timer and sharing' I didn't understand any of this part I'm afraid. Maybe because there's special characters in the comment that aren't readable?

We can probably fix whatever hovers, but we are trying to avoid tooltips in general.

Nothing should 'keep resetting to the initial value', but it's normal/currently set so that after a share stops to have the timer reset to a time that isn't already in the past, for better UX

See what Micah thinks about the rest, he might've understood better than me

maqp:
>We can't really handle a myriad of different ways to display the time based on the user's cultural background.. 

I understand. Surely either will be understood.

>"The field highlight is not working properly, at least under the Finnish locale." presumably a Qt bug

That's entirely possible. Also, I might have touched Qt settings on my computer. I'll try with a clean OS at some point and see if the problem persists.

>'Common settings should probably be visible during startup timer and sharing' I didn't understand any of this part I'm afraid. Maybe because there's special characters in the comment that aren't readable?

You've got me puzzled too now. Can you post a screenshot about how it shows to you. I'm willing to create mockups too if those help.

>We can probably fix whatever hovers, but we are trying to avoid tooltips in general.

I agree its better if there are less tool and prefer there were none. But since those were being used I tried to utilize them in the best way I could.

>Nothing should 'keep resetting to the initial value', but it's normal/currently set so that after a share stops to have the timer reset to a time that isn't already in the past, for better UX

So what happens is, if I click e.g. after the hour section that says `10`, press backspace and `1` so it says `11` when I press enter to confirm the value, it will switch back to `10`. On the other hand, if I click on the hour, press up arrow to move the hour from 10 to 11 and press enter, it will remember the 11.

>See what Micah thinks about the rest, he might've understood better than me

I apologize I'm not explaining this clearly. I'll try to create the mockups soon.

mig5:
> You've got me puzzled too now. Can you post a screenshot about how it shows to you. 

Screenshot_2019-03-21_13-32-25


> So what happens is, if I click e.g. after the hour section that says `10`, press backspace and `1` so it says `11` when I press enter to confirm the value, it will switch back to `10`. On the other hand, if I click on the hour, press up arrow to move the hour from 10 to 11 and press enter, it will remember the 11.

OK, I see. I have never pressed enter on those time values - I've simply entered the number/time that I want and then clicked start..

micahflee:
Ok I'm finally reviewing this again. You totally fixed everything I wanted @mig5, but I have a few more issues, and I also want to incorporate some of @maqp's feedback into the review too. (Thanks for all of your feedback @maqp.) In order to try to make it more clear exactly what's done, and what else I think would be good to include, I'm gonna use check boxes:

- [x] This task is done
- [ ] This is a task I think you should do

---

- [x]  Bundled tor works (in both CLI and GUI)

- [x] The CLI output is friendlier

```
$ ./dev_scripts/onionshare --startup-timer 60 screenshots/*
OnionShare 2.0 | https://onionshare.org/
Give this address to your recipient, and tell them it won't be accessible until: Mar 22, 04:19:33PM
http://rx6yolakbfqqc4sgw3zdpdsvfhvzhhoesi45basgtlrfto4n6t7go3ad.onion/bakery-sulfite

Waiting for the scheduled time before starting...
Setting up onion service on port 17628.
Compressing files.
 * Running on http://127.0.0.1:17628/ (Press CTRL+C to quit)

Server started

Press Ctrl+C to stop the server
```

- [x]  "What's this?" button in settings links to the right place (though we still need to actually write that wiki page)

- [x] If the startup timer is later than the shutdown timer, it displays an error in the GUI
- [ ] However, this is not the case in the CLI. Can you make it print the same warning and quit in the CLI, in that case?

```
$ ./dev_scripts/onionshare --receive --startup-timer 120 --shutdown-timeout 60
OnionShare 2.0 | https://onionshare.org/
Files sent to you appear in this folder: /home/user/OnionShare

Warning: Receive mode lets people upload files to your computer. Some files can potentially take control of your computer if you open them. Only open things from people you trust, or if you know what you are doing.

Give this address to your sender, and tell them it won't be accessible until: Mar 22, 04:27:33PM
http://il26ubugeadubwlfis5fohnmzd7efjwptbk4eaftufs2n2iz7awokyqd.onion/fiber-pregnant

Waiting for the scheduled time before starting...
```

- [ ] Also, can you remove the newlines from these strings? I just think it looks better...

```diff
diff --git a/share/locale/en.json b/share/locale/en.json
index 4610f84..c932f51 100644
--- a/share/locale/en.json
+++ b/share/locale/en.json
@@ -128,10 +128,10 @@
     "gui_tor_connection_error_settings": "Try changing how OnionShare connects to the Tor network in the settings.",
     "gui_tor_connection_canceled": "Could not connect to Tor.\n\nEnsure you are connected to the Internet, then re-open OnionShare and set up its connection to Tor.",
     "gui_tor_connection_lost": "Disconnected from Tor.",
-    "gui_server_started_after_timeout": "The auto-stop timer ran out before the server started.\nPlease make a new share.",
-    "gui_server_timeout_expired": "The auto-stop timer already ran out.\nPlease update it to start sharing.",
-    "gui_server_startup_timer_expired": "The scheduled time has already passed.\nPlease update it to start sharing.",
-    "gui_timeout_cant_be_earlier_than_startup": "The auto-stop time can't be the same or earlier than the start-up time.\nPlease update it to start sharing.",
+    "gui_server_started_after_timeout": "The auto-stop timer ran out before the server started. Please make a new share.",
+    "gui_server_timeout_expired": "The auto-stop timer already ran out. Please update it to start sharing.",
+    "gui_server_startup_timer_expired": "The scheduled time has already passed. Please update it to start sharing.",
+    "gui_timeout_cant_be_earlier_than_startup": "The auto-stop time can't be the same or earlier than the start-up time. Please update it to start sharing.",
     "share_via_onionshare": "OnionShare it",
     "gui_connect_to_tor_for_onion_settings": "Connect to Tor to see onion service settings",
     "gui_use_legacy_v2_onions_checkbox": "Use legacy addresses",
```

Regarding the datetime selector issue that @maqp pointed out, I'm not able to reproduce the problem, with my OS locale set to `en_US`. I can type numbers to change the date, and it just works and moves onto the next field automatically.

And even when I change the OnionShare language to Swedish (although Finnish is at 100% translation, it's not an option yet in the settings yet), the `QDateTimeEdit` still displays the datetime in English:

Screenshot from 2019-03-22 16-31-12

So I agree this looks like it's probably a Qt issue. I think, for the moment, we should go ahead and keep it as-is, and possibly consider improving the localization of `QDateTimeEdit` widgets as a separate PR in the future.

I think @maqp's suggestion about the URL description label makes sense. If you start a server with the autostart timer, it shows you the URL, but doesn't have the clear sentence explaining what people should do with it:

Screenshot from 2019-03-22 16-43-54

It only shows that sentence after the server has started.

Screenshot from 2019-03-22 16-43-41

- [ ] Can you make it show that the URL description label before the server has started?

I also think it makes sense to make the button text and tooltip consistent for both startup and shutdown timers.

- [ ] Right now when the shutdown timer is set, the server button text says something like `Stop Sharing (85727s remaining)`. Can you change it to split seconds into minutes, hours, and days? So that would be `Stop Sharing (23h, 48m, 47s remaining)`.

- [ ] And right now, if you mouse-over the button while waiting for it to shutdown the tooltip says `Auto-stop timer ends at [datetime]`, only the datetime is formatted as `YYYY-MM-DD HH:MM:SS`. Can you change the the text to `Auto-stop timer set to [datetime]` and change the datetime format to be consistent with the other places we display dates with spelling out the month and 12 hour clock? (In the future maybe we can make the date formatting configurable, but for now I think it's better to just keep it consistent everywhere.)

- [ ] Right now when you're waiting for the startup timer, the server button says `Scheduled for [datetime]. Click to cancel.` Can you change it to be consistent with the button text with shutdown timer? So instead it can say for example `Scheduled for 4h, 17m, 52s. Click to cancel.`,

- [ ] And during the startup timer, the server button doesn't have a tooltip. Can you make the tooltip say `Auto-start timer set to [datetime]`?

Also I found a new issue I think related to the UI not checking the settings properly or something. Here's how to reproduce:

- Open onionshare-gui, open settings, check auto-start timer (but leave auto-stop timer unchecked)
- Add files and start a share, then stop it
- Open settings, uncheck auto-start timer
- Click "Start sharing" again
- It's starting as if the auto-start timer is on, showing you the URL early and the button says `Scheduled for [datetime]. Click to cancel.` Instead, it should just start immediately and not use the auto-start timer.

- [ ] Can you fix this, so that when you flip the setting back and forth, the ServerStatus object gets  updated on the changes?

And finally, I noticed that the CLI argument names for the startup and shutdown timers aren't consistent:

```
$ ./dev_scripts/onionshare -h
OnionShare 2.0 | https://onionshare.org/
usage: onionshare [-h] [--local-only] [--stay-open] [--startup-timer <int>]
                  [--shutdown-timeout <int>] [--stealth] [--receive]
                  [--config config] [--debug]
                  [filename [filename ...]]

positional arguments:
  filename                  List of files or folders to share

optional arguments:
  -h, --help                show this help message and exit
  --local-only              Don't use Tor (only for development)
  --stay-open               Continue sharing after files have been sent
  --startup-timer <int>     Schedule this share to start N seconds from now
  --shutdown-timeout <int>  Stop sharing after a given amount of seconds
  --stealth                 Use client authorization (advanced)
  --receive                 Receive shares instead of sending them
  --config config           Custom JSON config file location (optional)
  --debug                   Log OnionShare errors to stdout, and web errors to
                            disk
```

It's `--startup-timer` and `--shutdown-timeout`, but the settings in to GUI call them "auto-start timer" and "auto-stop timer".

- [ ] Can you change these CLI arguments to `--auto-start-timer` and `--auto-stop-timer` instead?

- [ ] And finally, this might be a pain but do you think it's worthwhile to rename all of the variables through onionshare to stop using the terms `startup_timer`, `shutdown_timer`, or `shutdown_timeout` (which are used all over the place, I just grepped for them) and switch to `autostart_timer` and `autostop_timer`? It's not necessary for making it work, and it won't change the functionality at all, but it will make the code a bit cleaner going forward. What do you think? If you don't think it's worth it, I'd be ok with merging without doing that.

mig5:
@micahflee do you want the actual settings keys renamed too (that end up in the json)? That to me seems the most complete solution, but technically has a minor backwards incompatibility issue in that users with the autostop timer setting enabled, might 'lose' that setting on upgrade. Not the end of the world, could include such a fact in release notes.

mig5:
@micahflee ready for testing again. I think I have covered all the outstanding issues.

mig5:
This is also maybe a poor moment to suggest, but I can't help but feel the timer implementations are very complicated/hacky, and impacted the codebase somewhat negatively.

If you can think of a cleaner/more modular way to do this, I'd welcome it. Happy for it to just be a PoC.

maqp:
@mig5:

>Maybe because there's special characters in the comment that aren't readable?

You were probably referring to the ⓘ symbol which I used to distinguish the SW's line 

"**Anyone** with this OnionShare address can **download** your files using the **Tor Browser** ⓘ"

You were seeing it correctly.

Anyway, I edited the [earlier comment](https://github.com/micahflee/onionshare/pull/934#issuecomment-474792474), hopefully it's now more clear. I also updated it according to the latest commit.

PR #939, opened 2019-03-10
Fix conflict with files unrelated to translation

emmapeel2:
I am not sure why weblate was having issues with stdeb.cfg.
I have merged develop with
`git merge -X theirs github-micah/develop`
and the conflict was removed.
Please merge so weblate hopefully stops complaining


emmapeel2:
wait i will do it the other way (checkout develop, merge weblate)

Issue #816, opened 2018-11-05
learn the URL programmatically


weaselp:
I'm thinking of using onionshare in the Debian Tor package regression tests. The plan is to launch an onionshare using the Tor to be tested, and try to fetch the file that we share.

What I'm looking for is some way for the onionshare process to a) signal that it has finished setting up the .onion and b) communicate the URL to me.

I could, of course, redirect stdout to a file and grep that in a while loop, but that sounds icky.

Maybe onionshare could, optionally, write the URL to a file (name or fd) provided on the command line once it's ready?

Cheers,

mig5:
Hi @weaselp - it's not a bad idea. Currently the CLI version of OnionShare will wait until it gets an .onion, so you wouldn't have to mess around with loops, and the onion address is actually printed to stdout, so you could potentially even grep the output of the onionshare CLI command that you executed.. still quite awkward.

However, there's also another option: I don't know if it's a practical solution for you, but OnionShare supports 'persistent' onion addresses. You could generate that persistent address via a regular OnionShare attempt and then keep the saved onionshare.json settings file that will then contain the private key of the .onion. If you run your tests using this same onionshare.json settings file, you'll get the same .onion address every time, which removes some complexity for you (you can hardcode the address you expect to see the onion address on).

Another option: in the `develop` branch we have heaps of GUI unit tests already (see in the `tests` directory). Some of these already test waiting for an .onion address, loading the .onion and downloading (or uploading) files. You could just run our pre-existing tests to achieve this? You need to run pytest with the `--runtor` flag to fire any of the tests in 'Tor connected' mode. 

Here is such a test: https://github.com/micahflee/onionshare/blob/develop/tests/onionshare_share_mode_download_test.py

The only other catch is that the Tor-connected tests don't use the 'bundled' Tor mode in OnionShare, they assume you're running Tor Browser in the background. But you could copy our test suite and  adjust the settings to use 'bundled' mode. See line 41 of https://github.com/micahflee/onionshare/blob/develop/tests/TorGuiBaseTest.py. Might save you some duplicate work?

In the meantime, I'll mark this as an enhancement, as we might indeed add at least a CLI flag to save the onionshare address (I'd actually like to go further than this, though, and maybe have a thin API that returns things like the onion address via JSON, as we could use that for our tests too).

mig5:
> However, there's also another option: I don't know if it's a practical solution for you, but OnionShare supports 'persistent' onion addresses.
[...]

PR #942, opened 2019-03-12
Save the onion address, slug and optionally hidservauth (if applicable) to the json settings if --debug is passed

mig5:
I was wondering what you thought of this solution to #816 

I thought if the user passed the `--debug` flag, we could simply save the new onion address, slug and optionally hidservauth to the json file. This is simpler than doing any other sort of json output since we already have our json right there.

It wouldn't interfere with persistence in the settings json, as these have their own keys prefixed with `debug_`

```
user@onionshare:~$ cat .config/onionshare/onionshare.json | json_pp | grep debug_
   "debug_slug" : "trout-playmate",
   "debug_hidservauth" : "",
   "debug_url" : "http://tglebv3kegvpkfh76ljhhq43zw4idu2wrzgrjzygtkbkg4whpztmoqyd.onion/trout-playmate",
   "debug_onion" : "tglebv3kegvpkfh76ljhhq43zw4idu2wrzgrjzygtkbkg4whpztmoqyd.onion",
```

It would allow the programmer to start an onionshare via CLI, perhaps with a wrapper Python script that calls it via subprocess (note: perhaps with a configurable timeout as per #941), wait for that length of time, then try a json lookup of these keys from the settings file (since we know the settings file's name and location) to obtain the values.

Felt easier to bundle it with 'debug' mode than add yet another CLI flag or other logic. But I'm open to it being rejected, it feels like it will rarely be used :)

And, of course, this is just for CLI use, since it's assumed only advanced users with unusual use cases would need it, e.g for programmatic purposes alone..

### DIFF
--- a/onionshare/__init__.py
+++ b/onionshare/__init__.py
@@ -18,7 +18,12 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 
-import os, sys, time, argparse, threading
+import argparse
+import json
+import os
+import sys
+import threading
+import time
 
 from . import strings
 from .common import Common
@@ -173,6 +178,15 @@ def main(cwd=None):
         else:
             url = 'http://{0:s}/{1:s}'.format(app.onion_host, web.slug)
 
+        # Save the onionshare details to json for programmatic use
+        if debug:
+            common.settings.set('debug_onion', app.onion_host)
+            common.settings.set('debug_url', url)
+            if not common.settings.get('public_mode'):
+                common.settings.set('debug_slug', web.slug)
+            if stealth:
+                common.settings.set('debug_hidservauth', app.auth_string)
+            common.settings.save()
         print('')
         if mode == 'receive':
             print(strings._('receive_mode_data_dir').format(common.settings.get('data_dir')))
--- a/onionshare/onion.py
+++ b/onionshare/onion.py
@@ -535,6 +535,10 @@ class Onion(object):
         except:
             pass
         self.service_id = None
+        debug_settings = [ 'debug_onion', 'debug_url', 'debug_slug', 'debug_hidservauth' ]
+        for key in debug_settings:
+            self.settings.set(key, '')
+        self.settings.save()
 
         if stop_tor:
             # Stop tor process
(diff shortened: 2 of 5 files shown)


micahflee:
Oh this is pretty cool. But I think it actually should be it's own separate CLI argument instead of tacking this functionality onto `--debug`.

The reason is because people may use onionshare with `--debug` for sensitive use-cases, just because they want to see what's going on. It's not intuitive that `--debug` will always save the the onion address, even when the persistence setting is disabled. For example I could see a situation where someone's computer is searched, and the onion address in their `settings.json` could potentially be used against them.

(And actually, now that I write this I wonder if `--debug` should be renamed to `--verbose`/`-v`, because that's what it actually does...)

But I would be into merging this if we include this functionality with a different CLI arg, like maybe `--programatic` or `--scripted` or something.

This is an out-there idea, but if we have a dedicated flag for programatic uses of OnionShare, another way to solve this same problem would be to have `stdout` print stuff in a totally different way for that flag. Like, each line could be a JSON object that's easily parsable, so the output could look something like this:

```
$ onionshare --programatic ~/files-to-share
{"status": "starting", "version": "2.0"}
{"status": "connected_to_tor"}
{"status": "ready", "slug": "trout-playmate", "hidservauth": "", "url": "http://tglebv3kegvpkfh76ljhhq43zw4idu2wrzgrjzygtkbkg4whpztmoqyd.onion/trout-playmate", "onion": "tglebv3kegvpkfh76ljhhq43zw4idu2wrzgrjzygtkbkg4whpztmoqyd.onion"}
{"status": "page_load", "path": "/"}
{"status": "page_load", "path": "/trout-playmate"}
{"status": "download_started"}
{"status": "download_finished"}
{"status": "closed_automatically"}
```

This would obviously take a lot more work and consideration, and thus maybe isn't worth it since I'm not sure using onionshare programmatically will be used very much.

Or `--programatic` could just do exactly as this PR does, except with `programatic_` prefixes in the settings instead of `debug_` prefixes. What do you think?

mig5:
I'll close it. I was not totally sold on the idea myself. I think a proper JSON output is warranted but that probably involves onionshare having a sort of query-able API of its own (not just JSON to stdout, which is not that easy to parse if the program doesn't exit) - a bigger feature needed.

PR #959, opened 2019-04-19
Rename --debug to --verbose

### LINKED ISSUE #958: Rename --debug to --verbose
Really all the `--debug` option does (for both CLI and GUI) is increase the verbosity of what gets outputed to `stdout`. It will be more clear if that option is renamed to `--verbose`, and just `-v`.

micahflee:
Resolves #958


onionshare/web/web.py
```diff
@@ -193,7 +193,7 @@ def generate_slug(self, persistent_slug=None):
             self.slug = self.common.build_slug()
             self.common.log('Web', 'generate_slug', 'built random slug: "{}"'.format(self.slug))
 
-    def debug_mode(self):
+    def verbose_mode(self):
         """
         Turn on debugging mode, which will log flask errors to a debug file.
```
mig5: There are a few references to `debug` in the code comment/other attributes in this function, but it's only superficial.

If you really wanted to be consistent you could apply this diff:

```
diff --git a/onionshare/web/web.py b/onionshare/web/web.py
index adadf7d..57e0fd4 100644
--- a/onionshare/web/web.py
+++ b/onionshare/web/web.py
@@ -195,10 +195,10 @@ class Web(object):
 
     def verbose_mode(self):
         """
-        Turn on debugging mode, which will log flask errors to a debug file.
+        Turn on verbose mode, which will log flask errors to a file.
         """
-        flask_debug_filename = os.path.join(self.common.build_data_dir(), 'flask_debug.log')
-        log_handler = logging.FileHandler(flask_debug_filename)
+        flask_log_filename = os.path.join(self.common.build_data_dir(), 'flask.log')
+        log_handler = logging.FileHandler(flask_log_filename)
         log_handler.setLevel(logging.WARNING)
         self.app.logger.addHandler(log_handler)
 
```

But I don't consider it a blocker.

PR #960, opened 2019-04-19
Support old linux

### LINKED ISSUE #918: Support slightly old linux versions
Failed when I'm trying to install deb in Ubuntu 16.04 with message:
Package 'python3-distutils' has no installation candidate

micahflee:
Resolves #918.

This adds support for Ubuntu 16.04 and Debian stretch.

We already no longer include the `python3-distutils` dependency. This also adds a try/catch block when disabling the flask banner, in case we're running an old version of flask that doesn't have a banner. And it skips all GUI and tor tests is the version of pytest is too old for the normal skip code to work.


micahflee:
@mig5 before reviewing it, can you look at #918 and let me know if you think there's anything else this PR might be missing, in order to support these older versions of Ubuntu and Debian? And are there other older linux distros we should also support, that aren't currently supported, which I should add to this PR?

PR #855, opened 2018-12-24
Website sharing

hiromipaw:
Hi,

I have managed to find time to develop this feature also for the gui component.
There are some adjustments that I will still need to do, but I just wanted to show you the functionality. 
I had started t code this before the new developments, so some of the ui is actually a bit funky at the moment, but I am not too concerned about that since we would probably need to reconsider a few things about this website publishing mode anyways.

Looking forward to see what you think about this.

PR #961, opened 2019-04-19
Static websites

### LINKED ISSUE #791: Static website sharing with OnionShare
With @holantonela have been thinking of how OnionShare could be used to share a static website. 

I have started to hack this functionality for the cli. 

https://github.com/hiromipaw/onionshare/tree/website-sharing

Please consider that there are still quite a few things to clean up and a few consideration to make, but since I believe there isn't better explanation of a feature than a few lines of code, I thought this could be a nice way to start a conversation about this.

Please let me know what you all think.

hiromipaw:
Hi,
There is still a bit of cleanup to do and tests to add. But this is almost everything we discussed during our meeting.
Sorry it took me so long to wrap this up. I am happy I finally got time to do this.

(Edit by mig5: Fixes #791 (just linking the issue for Github auto-closing of ticket purposes later)) )


micahflee:
Hey hiro, I'm excited to start reviewing this one!

Is this a new PR that replaces #855, so I can close that one?

Also there are some merge conflicts, could you resolve them?

hiromipaw:
Hey @micahflee! I think the only thing missing atm is testing. Not sure we want to push it for this version or we can slack a bit about it :wink:

onionshare/__init__.py
```diff
@@ -51,6 +51,7 @@ def main(cwd=None):
     parser.add_argument('--connect-timeout', metavar='<int>', dest='connect_timeout', default=120, help="Give up connecting to Tor after a given amount of seconds (default: 120)")
     parser.add_argument('--stealth', action='store_true', dest='stealth', help="Use client authorization (advanced)")
     parser.add_argument('--receive', action='store_true', dest='receive', help="Receive shares instead of sending them")
+    parser.add_argument('--website', action='store_true', dest='website', help=strings._("help_website"))
```
mig5: As of OnionShare 2.1, we no longer are using translatable strings in the CLI version (see the above strings which are now just hardcoded in English).

Are you able to replace your translatable strings to be hardcoded English strings like the above?

And in fact, `help_website` is already missing from the locale `en.json` so it doesn't work via CLI right now :)

onionshare/__init__.py
```diff
@@ -168,6 +172,15 @@ def main(cwd=None):
         print(e.args[0])
         sys.exit()
 
+    if mode == 'website':
+        # Prepare files to share
+        print(strings._("preparing_website"))
```
mig5: As above, please make this a hardcoded English string rather than a translatable `strings._()`

onionshare_gui/mode/website_mode/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+import os
+import secrets
```
mig5: 'secrets' module is shipped only with Python 3.6+. This means I can't run it on Fedora 29  (v3.5) and probably other distros.

Is there a way we can have backwards compatibility for older Python 3? 

@micahflee achieved this in his other project PassphraseMe by using `random` on older Python. See https://github.com/micahflee/passphraseme/commit/ce9a2005a8a01b6cc57c92d9a827337aef322b86

Having said that, I can't see where `secrets.choice` is actually being used, so maybe it's redundant and can be removed? Perhaps because you use the `web.slug` as the password for HTTP auth?

onionshare_gui/onionshare_gui.py
```diff
@@ -317,6 +367,7 @@ def reload_settings():
                         self.timer.start(500)
                     self.share_mode.on_reload_settings()
                     self.receive_mode.on_reload_settings()
+                    self.website_mode.on_reload_settings()
                     self.status_bar.clearMessage()
 
             # If we switched off the auto-stop timer setting, ensure the widget is hidden.
```
mig5: In this section that beings 'If we switched off the auto-stop timer setting'.... we have this block of code:

```
            # If we switched off the auto-stop timer setting, ensure the widget is hidden.
            if not self.common.settings.get('autostop_timer'):
                self.share_mode.server_status.autostop_timer_container.hide()
                self.receive_mode.server_status.autostop_timer_container.hide()
            # If we switched off the auto-start timer setting, ensure the widget is hidden.
            if not self.common.settings.get('autostart_timer'):
                self.share_mode.server_status.autostart_timer_datetime = None
                self.receive_mode.server_status.autostart_timer_datetime = None
                self.share_mode.server_status.autostart_timer_container.hide()
                self.receive_mode.server_status.autostart_timer_container.hide()
```

You will need to add an equivalent section for the `self.website_mode.server_status.autostart_timer and related attributes. Otherwise, if autostop or autostart timer is disabled in the Settings dialog whilst using Website Mode, the timers still appear in the GUI.

mig5:
Fixes #791 (just linking the issue for Github auto-closing of ticket purposes later)

micahflee:
This is so cool @hiromipaw! I _love_ the use of http basic auth too, instead of the slug in the URL. Actually, I wonder if we should use http basic auth instead of a slug URL for all of onionshare, not just for website sharing... but that should be a separate PR.

Also thank you for reviewing this @mig5! I agree with all of the things he brought up.

The first thing I noticed is in CLI mode, there's no validation on supplying filenames. If I just run `./dev_scripts/onionshare --website` it works, but loading the site gives me a 404 error. If you don't specify any files, it should display the usage again, like it does what using share mode (just running `./dev_scripts/onionshare` without specifying files).

The next thing is the directory listings. It lists files, but not folders, and there's no nested folders. For example, when I run this from the onionshare source tree: `./dev_scripts/onionshare --website share`, to share the files in the `share` directory. This is what the website looks like:

Screen Shot 2019-05-10 at 10 11 47 AM

The directory listing has all of the files in it, but it doesn't include the folder: `images`, `locale`, `static`, or `templates`. However, I know they're there. When I click "download" on `wordlist.txt` it links to `/download/wordlist.txt`, and if I go to `/download/static/img/ajax.gif` it actually loads that gif, so I know that the folders are recursively there even though they're not displayed in the directory listing.

Could you make the directory listing also show folders, and let you click through them recursively?

Also, I think this directory listing issue has a larger problem. I think there shouldn't need to be the `/download/[path]` route at all. Instead, you should get `wordlist.txt` by just loading `/wordlist.txt` -- right now, that displays the directory listing again (so do other routes that should be 404 errors, like `/asdfasdf`). And going to `/locale` should show another directory listing of just what's in the `locale` folder.

Another weird thing. Each time I refresh the page, the "Total size" in the top right corner gets higher. It started out as 1.7 MiB, but after refreshing a few times, it's up to 2.2mb.

Screen Shot 2019-05-10 at 10 14 31 AM

And finally, I think website sharing mode can't have static files itself, or any built-in paths of its own (like `/download`). For example https://micahflee.com/ is a static website, and when I try sharing its files this is what it looks like in onionshare:

Screen Shot 2019-05-10 at 12 11 19 PM

The reason is because the stylesheet is at `/static/style.css` (which is https://micahflee.com/static/style.css), but onionshare is reserving the whole `/static` folder for its own resources (its own stylesheet, images, etc.), which means websites you share via onionshare can't include their own `/static` folder. Similarly, what if a static website wants to have a folder called `/download`? If we want static website to be able to include whatever arbitrary filenames people want to use, then onionshare can't clobber any of them.

This also means that the pages that actually have resources (like the directory listing page, the 404 page, etc) need to include all of their resources inline. So the stylesheet needs to be right in the html file, and the images need to be data URIs. Or, alternatively, onionshare could create an random folder name for its static resources, like `/static_[random-20-chars]/` or something.

micahflee:
Also, since you get to drag arbitrary files and folders into onionshare to share as a website, what happens when multiple files have the same filename? Like what if I drag two separate index.html files in it -- which one gets rendered? Maybe in the case of a filename collision it should throw an error and not let you start the server.

micahflee:
Hey @hiromipaw check it out! I just pushed a bunch of commits into the [hiromipaw-static-websites](https://github.com/micahflee/onionshare/tree/hiromipaw-static-websites) branch that implements some of this. Let me see if I can make a PR into your branch...

micahflee:
@mig5 and @hiromipaw made a good suggestion in the Keybase chat about this feature. Instead of having website sharing mode be separate, maybe this should just be integrated into share mode. I think this is a great idea, but here are a few things to consider:

Right now if you have an `index.html` file in any folder you're sharing, it has special meaning: it gets rendered instead of the directory listing. But in share mode, you might not always want this. If you share a bunch of files, and one of them is `index.html`, maybe you _want_ a directory listing instead, so the user can see all of the files. And in website sharing mode, in some situations maybe you want directory listing disabled.

So if we do this, I think there needs to be a "Directory Listing" setting, with the choice being a dropdown or radio button with options like:

- "Show directory listings when there isn't an index.html"
- "Always show directory listings"
- "Never show directory listings"

Otherwise, there's no chance of having a "download" button that downloads all of the data.

The other thing to consider is zipping files up. In share mode, the first step is to zip up all of the files into a single compressed file that's stored in a temporary folder, and then to let people access that, but so far website mode just skips that step, and let's people download file directly.

I can picture people using website sharing mode to make _huge_ amounts of data available, like terrabytes of video files, for example. In this case, someone can just browse it and download the individual videos they want, and onionshare will serve them from their original location (including, e.g. a USB hard drive). But if you click "download", it will have to compress all of the data into a temporary folder first, and then let you download the zip.

Hmm, from the root directory listing there could be a "download all" button in the corner, and there could also be a "download" button next to each folder, and it could lazily compress folders on an as-needed basis, so e.g. wait until the first time someone clicks a download button.

Anyway, these are my initial thoughts. But I like the idea.

micahflee:
Another thought about this feature -- right now we use `send_from_directory` to serve files, but this doesn't use gzip compression. I think each file we serve should use gzip compression, which might speed up downloads a lot, especially over onion services.

We already have some gzip code here https://github.com/hiromipaw/onionshare/blob/static-websites/onionshare/web/share_mode.py#L304 but maybe it would be better to stream the compression on-the-fly instead of compressing each file.

mig5:
The idea I had, was that 'Website mode' could be nothing more than a checkbox in the Settings area under Share settings: 'Enable Website mode?'

If this mode is checked:

1) index.html files (or other html files) are rendered if present
2) there is no 'download' button for the overall share (to avoid the second issue you mentioned)

If this mode is *not* checked:

1) directory listings are used (no html rendering)
2) a 'download' button is present to download the .zip of the share 
(so basically, the same as existing Share mode, except that thanks to this PR's features, there's also an individual 'download' button for each file)

Would that address the issues? Basically with Website mode you turn on HTML rendering, at the cost of losing the Download (zip) button. Vice versa for regular share mode, but under both modes, technically each individual file can be 'downloaded'.

hiromipaw:
@mig5 I am not sure I see website mode as a checkbox, because the idea with website mode is that you publish a website. Going way forward I see he possibility for a quick and dirty html editor in website mode, where someone could quickly sketch a web page (like with lektor crm mode).

Personally I think it makes sense that if you are on website mode, in order not to break the functionality, a directory listing page is shown if there are no html files. But this is only because we are familiar with how a web server works.  I think from an application point of view we could even make the choice to display an error message. Something like "add an html file or switch to share mode".

At the same time I think we shouldn't push for users to use the website mode to share files that you can download independently. I'd rather see the share mode give this possibility. Maybe in this case a checkbox in settings could be useful. Something like: "allow visitors to download individual files" -> yes/no

I am curious to hear what does @holantonela think?

mig5:
> @mig5 I am not sure I see website mode as a checkbox, because the idea with website mode is that you publish a website. 

> We could merge them in the future, but the switcher needs to live close to the drag and drop area. The user option could change every time users are sharing/publishing data so the use frequency could be high. And, having it in the general settings is not intuitive at all.

Sure, it doesn't have to be in the general Settings area. We already have UI stuff near the drag-n-drop for the AutoStop and AutoStart timers (when those features are enabled), we could have something similar for activating 'website' mode just before starting a share, without needing to go to General Settings.

> At the same time I think we shouldn't push for users to use the website mode to share files that you can download independently. I'd rather see the share mode give this possibility. 

Absolutely yes,  this was what I was (trying, poorly) to say :) 

That's why I felt, if this 'download independently' functionality was in Share mode, then the two modes could be 'merged' because all that really remained in 'Website Mode' (for now), was the rendering of .html files if present. 

But I guess it doesn't have to be merged. The problem, though, is if we released OnionShare with a third tab 'Website Mode', and then in a *later* release, removed it (by merging the functionality), this might cause confusion for users who got used to that mode being its own tab.

I think I've created a bit of noise here, sorry. Probably should pull back to let the focus be on Micah's earlier review :)

micahflee:
You make a good point @holantonela that share mode and website mode have different use-cases, so even if the programming logic is almost the same, the UI should probably display them as separate tabs. This also gives us the flexibility to build more features into the modes separately if we want (like a future html editor for website mode, that isn't part of share mode).

But that said, we should make sure that the functionality between the two modes shares as much code as possible. Here's what I think we need to do to start getting there:

Share mode (and receive mode while you're at it) need to switch how slugs work. If public mode is disabled, then they should use HTTP basic auth just like website mode. This is required for share mode to support directory listing. So all of the basic auth stuff needs to be moved from `onionshare/web/website_mode.py` into `onionshare/web/web.py` and to work for all modes. (This will also probably require fixing up the tests so they still pass.)

`WebsiteMode.set_file_info` right now builds up `self.files` and `self.root_files`, which is used in the logic to choose when to render directory listings and when to render html files. And `WebsiteMode.directory_listing` displays the actual directory listing. Both of these functions would need to be available to `ShareMode`, but `ShareMode.set_file_info` (which does a totally separate thing, compresses the files) would still need to exist. So to merge these two, maybe it would make sense to make a new `BaseShareMode` class which implements all of the shared functionality, and then make `ShareMode` and `WebsiteMode` inherit from `BaseShareMode`. (I could also help with this if you want @hiromipaw).

And finally, here's a list of other things that need to be worked on in the mean time:

- The CLI strings shouldn't be translated anymore
- Like @mig5 mentioned in one of his comments, the `secrets` module should fallback to `random.SystemRandom`, to we can support python < 3.6 (see [here](https://github.com/micahflee/onionshare/pull/961#discussion_r282712153))
- Like @mig5 mentioned, there's a tiny patch needed for auto-stop and auto-start timers (see [here](https://github.com/micahflee/onionshare/pull/961#discussion_r282713261))
- Desktop notifications: It shouldn't pop up a notification for each and every web request. Maybe only for loading `/`, it can see "Someone has loaded your website's homepage"
- History widget: This hasn't been mentioned yet, but the history widget (clicking the little arrow button to expand the widget on the right) I think should look more like web logs. Right now it says "Someone has visited your website on [date]" but I think instead each web request should be something like: `2019-05-20 10:42:58 GET / 200`, or `2019-05-20 10:43:02 GET /favicon.ico 404`
- Running `onionshare --website` without any files should display the usage -- listing files should be required
- Share mode and website mode should stop using static files in `/static` and instead be self-contained. So I think we could pass the content of `/static/css/style.css` into their templates as a variable, to include css inline. And we could also pass the base64-encoded content of images in `/static/img/*` to display them as data URIs. This way, if we're sharing a folder named `static` (in either share mode or website mode) it will just work
- Directory listing page should display the path that it's displaying, and if it's not the root should at least have a "Parent directory"/`..` link, but would be cool to have full breadcrumbs too

micahflee:
I'm working on making another PR into your PR on two parts of this that require a lot of refactoring: making it so all modes use HTTP basic auth instead of slugs, and refactoring ShareMode and WebsiteMode so they inherit a common class in order to share code.

While working on the HTTP auth part, I ran into a few issues that I wanted to lay out here just to be thorough.

The code that makes sure the slug is correct used to use a constant-time comparison function, but now that comparison code is wrapped up in the flask_httpauth module so it's not constant time. OnionShare did this to avoid attackers guessing the slug using timing attacks (e.g. if the slug is "glacier-vertebrae", a guess that starts with "g" will take slightly longer to compute than guesses that start with any other letter because invalid slugs will get rejected immediately, instead of looping one more time, so you know the first letter of the slug is "g"). But anyway, I don't think this matter because in non-public mode we automatically stop the server after 20 wrong guesses.

But the other problem is detecting wrong guesses. Before we would consider a 404 file not found error to be a wrong guess, but now it's actually simpler in a way, because a 401 unauthorized access error is a wrong guess, and we don't have to care about 404 errors. However, I'm having trouble defining an error handler for 401 errors because flask_httpauth has already defined this error handler (see [this code](https://github.com/miguelgrinberg/Flask-HTTPAuth/blob/master/flask_httpauth.py#L31)). This might be tricky, but I'll see if I can work around it.

(This was also something I missed in the first review of this PR -- that since website mode was using HTTP auth, the code to prevent guessing the slug wasn't working for website mode, because the slug worked in a different way.)

micahflee:
I just opened another PR into this PR https://github.com/hiromipaw/onionshare/pull/2 but it's not ready to merge yet. I just wanted to explain what I had done so far, and list what was still needed before it's ready.

This PR will implement HTTP basic auth for all modes, as well make templates self-contained so that they don't load any resources from `/static`. I'll let you know when it's done, and would also appreciate a code review before you merge :).

micahflee:
This PR is getting big, and some of the subtasks require refactoring a lot of OnionShare. So much so that I'm wanting to fix unrelated issues and merge them into this branch instead of `develop`, to avoid merge conflicts in the future.

So given all of that, I think it makes sense to go ahead and merge this PR into `develop`. Then we can open github issues for all of the subtasks, and have the flexibility to merge those in as separate, smaller PRs. This will also make it easier to ensure that all of these smaller subtask PRs get good code review individually before getting merged.

I just closed my work-in-progress PR that extends basic auth support to all modes (https://github.com/hiromipaw/onionshare/pull/2). Instead, after merging this PR into `develop`, I'll open a new PR for that specific feature (once it's done -- all I have left is fixing the tests so that should be soon).

Before merging this PR, we need the CircleCI tests to pass, and right now they fail in python 3.5 because of the use of the secrets module. @hiromipaw would you be able to fix the python 3.5 issue (described [here](https://github.com/micahflee/onionshare/pull/961#discussion_r282712153))? Then I'll give it one final review and if it looks good, merge it :).

micahflee:
Ok @hiromipaw @mig5, I just split the remaining website mode tasks into 5 separate issues all with the new "website mode" label. I'm working on #990 and nearly done, but it would be cool to merge this PR first, so that my PR for that could get merged directly to develop.

So when you get a chance @hiromipaw fix the python 3.5/secrets module issue. And if you want, go ahead and fix other issues too -- this PR can close some of the issues that I just opened. But it would be good to merge it soon.

onionshare_gui/mode/website_mode/__init__.py
hiromipaw: I did a test a while back and then forgot to remove the import.

Issue #943, opened 2019-03-14
Running as a user with no home directory


bentley:
On OpenBSD, we often provide [daemon control scripts](https://man.openbsd.org/rc.d.8) for packages that users install, so the daemon can be started automatically on bootup and flags can be saved in config files. Typically the daemon runs as a dedicated user created by our package manager with a home directory of `/var/empty`.

I tried something consistent with how we usually set up our packages: running as an unprivileged user `_onionshare`, executing `onionshare --config /etc/onionshare.json --receive`, and a `_onionshare`‐owned `data_dir` of `/var/onionshare`. Unfortunately, it fails because OnionShare always tries to create `~/.config`:

    Traceback (most recent call last):
      File "/usr/local/bin/onionshare", line 22, in <module>
        onionshare.main()
      File "/usr/local/lib/python3.6/site-packages/onionshare/__init__.py", line 41, in main
        common.load_settings()
      File "/usr/local/lib/python3.6/site-packages/onionshare/common.py", line 55, in load_settings
        self.settings = Settings(self, config)
      File "/usr/local/lib/python3.6/site-packages/onionshare/settings.py", line 48, in __init__
        self.filename = self.build_filename()
      File "/usr/local/lib/python3.6/site-packages/onionshare/settings.py", line 139, in build_filename
        return os.path.join(self.common.build_data_dir(), 'onionshare.json')
      File "/usr/local/lib/python3.6/site-packages/onionshare/common.py", line 143, in build_data_dir
        os.makedirs(onionshare_data_dir, 0o700, True)
      File "/usr/local/lib/python3.6/os.py", line 210, in makedirs
        makedirs(head, mode, exist_ok)
      File "/usr/local/lib/python3.6/os.py", line 220, in makedirs
        mkdir(name, mode)
    PermissionError: [Errno 13] Permission denied: '/var/empty/.config'

It would be nice if `onionshare_data_dir` could be set through a flag or a JSON option.

mig5:
This would hopefully be fully solved by #366 , but indeed we should not attempt to create the .config directory if we have already passed a custom settings json file.

mig5:
Long story but after some investigation, this regression actually relates to the localisation of CLI strings. We have to load settings 'early' and this has inadvertently resulted in triggering logic that tries the default settings json location.

Noting it for my own reference just because #879 will probably unblock this.

mig5:
[...]

PR #963, opened 2019-04-20
Unlocalize CLI strings

### LINKED ISSUE #879: Localizing CLI strings?
I got feedback from the person who's translating OnionShare to Russian that maybe we shouldn't localize the command line strings, and instead keep them in English. They said that it's very odd for command line tools to not be in English (that most Russian sysadmins and programmers who use the CLI a lot use it all in English), and that it looks weird to have the OnionShare CLI in Russia. They suggest we keep the CLI in English, but localize the GUI in Russian.

It seems like the options are:

1. Translate all strings, including CLI strings
2. Keep CLI strings in English, and only translate GUI strings
3. Keep CLI strings in English only for the Russian language, and translate all strings for all other languages

@emmapeel2 and anyone else knowledgeable about localization issues, what do you think?

mig5:
Fixes #879 

I had to `common.load_settings()` and `strings.load_strings()` in a few unexpected places (onion.py and onionshare_gui.py) after ripping the strings out of the CLI and GUI `__init__.py` files. I hope it's OK...

Unfortunately it doesn't deal with the problem in #943 though, because various functions in `common` that build data directories etc, still try to expand the user's home directory. If they don't have one, it bails out.

(I previously had hoped that avoiding loading settings 'early', which we previously did in order to load localised strings early, would prevent this lookup of the user's home dir if a user was passing in a custom `--config` json file, but it's not enough)

Ultimately it even bails on using the `--verbose` flag because it wants to create the flask.log file in the user's home dir.

Maybe that could be fixed later, right now it's a bit beyond my means.


micahflee:
I found a few more CLI strings that you missed (by doing `grep -r "print(" onionshare | grep strings`) and added my own commit to remove them as well.

PR #968, opened 2019-04-22
Optimize `Common.build_slug`

illia-v:
The method used to load the whole file with words into the memory and convert it to a list. It is not efficient.

This solution reduces the maximum resident set size of the method by ~500 kilobytes and its execution time by ~30%.


micahflee:
OnionShare no longer uses slugs

PR #996, opened 2019-05-25
Switch from slugs to HTTP basic auth

### LINKED ISSUE #990: Switch from slugs to HTTP basic auth
Website mode as implemented in #961 uses HTTP basic auth instead of a slug, and we should apply this to share mode and receive mode as well.

I've already started work on this in the `basic_auth_everywhere` branch: https://github.com/micahflee/onionshare/tree/basic_auth_everywhere

> Share mode (and receive mode while you're at it) need to switch how slugs work. If public mode is disabled, then they should use HTTP basic auth just like website mode. This is required for share mode to support directory listing. So all of the basic auth stuff needs to be moved from onionshare/web/website_mode.py into onionshare/web/web.py and to work for all modes. (This will also probably require fixing up the tests so they still pass.)

micahflee:
This PR resolves #990.

This takes the HTTP basic authentication code in `WebsiteMode` and moves it into `Web`, so that all modes use it instead of just website mode.

By removing the `/slug/` addresses, the flask code is actually much simpler and more straighforward now. We no longer have to deal with two separate routes for each route (e.g. `/download` for public mode and `/[slug]/download` for not public mode). Regardless of public mode, everything uses the same route.

OnionShare used to detect slug guessing attacks by counting how many 404 errors were made, and if it got to 20 automatically shutting down the server. This wasn't the best because it occasionally got shut down not because of an attack, but just enough legit 404 errors. But now that we're using HTTP basic auth, this rate limiting works much better. It only counts invalid password guesses (and counts them smarter too -- if someone guesses the same password twice, that only counts once) by counting 401 errors.

This also updates the `shutdown_slug` login that is used to shut down the server when the user clicks the stop button to use HTTP basic auth too, so that shutdown request actually works.

Because the slugs are legitimately being used as passwords now (with the username "onionshare"), this also renames the term "slug" to "password".

And when it displays URLs to copy and paste, the URLs now all look like `http://onionshare:[password]@[onionaddress].onion` when public mode is disabled.

One thing I really like about this is it actually adds a new protection against shoulder surfing. Once you paste a URL like that into Tor Browser and confirm you want to login to the site with that username and password, the username and password part of the URL don't appear in the URL bar. So if you're at an OnionShare site in Tor Browser and someone looks at your screen, the password is not displayed, where as before the slug used to be displayed.

Additionally, this PR also makes it so static resources are in an unguessable folder, like `/static_bt2oq32ye4z2glafaukej5yvxa/css/style.css`. This way we can still use static resources in website mode, because there is no risk of the website someone is creating having a file with the name `static_bt2oq32ye4z2glafaukej5yvxa`. (The random part at the end is 128-bits of random data.)

And the static resources are exempt from HTTP basic auth. This way, on a 401 Unauthorized Access error, the 401 template can still use static resources.

This was a major refactor, so many tests also needed to be fixed.


hiromipaw:
This is a :+1: from me.

hiromipaw:
@mig5 I think your observations are on point but I am wondering if, given that everyone knows the username, we should consider case 1 as a possible attack. Using a wrong username after all does not give any information regarding the password as the error thrown by the web server will be a 401 in any case. I do agree also that the DoS case might be a valid argument here, while at the same time, if someone knows the onion service address they can perform already a number of DoS attacks.

micahflee:
@mig5:

> I think the 'incorrect password' should be expanded to check the username too. Because I found if I put a false username, and either a false password or the correct password, the incorrect username meant that the request was not added to the bad guess 'count'. So an attacker could keep trying to guess a username/combo - if they pick a username other than 'onionshare' the attack will continue. Not a big risk but I think we should clamp down on any such attack.

Here's my thinking. The goal of this rate limiting is to prevent someone from guessing the 2-word password and thus accessing the OnionShare site while only knowing the onion address. Since the only valid username is "onionshare", I don't see any problem with just allowing guesses with other usernames through because they will never succeed in gaining access to the site that way.

>  What's the use case of 'the same incorrect password doesn't count twice'? If an attacker spun up a farm of 20 machines and somehow knew the onion address, and directed all 20 machines to attempt to login with the same credentials, only one failed attempt would occur despite multiple attacks. To me any incorrect password (whether or not it's the same password tried before) should be counted? It's not a big deal though, this is corner case stuff. But maybe even there's a DoS risk, in that swamping the onion with the same invalid password over and over will only count 'once', and maybe the attacker could try and use up available connections.

I actually was thinking of implementing this for 404 errors, before making this PR, but figured it made sense to include in this PR anyway. Basically, I want to reduce the risk of someone accidentally shutting down the OnionShare server, and to have it only shut down on an actual attack. The keyspace for a 2-word password is finite, and guessing the same password multiple times doesn't actually help the attacker work through the keyspace any faster, so I figured it's safe to ignore those guesses. This way if, for example, someone has the right onion address but the wrong password written down, and they keep trying to type the password over and over again, it won't shutdown the server, because they're not actually doing a brute force guessing attack.

> I found that the randomised static_ path for assets, remained the same across multiple shares. Should it perhaps build a new random static path for each new share? I don't think there's a big risk here either, but just felt like each new share should always be 'unique' like the onion address/password is, unless of course if persistence is in use (but even new persistent shares could safely use a new random static asset path).

This is a good catch. I also don't see a security risk here, but yeah it seems cleaner to generate a fresh static_path each share. I will fix.

@hiromipaw: Yeah I totally agree that if someone knows the onion address there are ways to do DoS attacks. But the simplest DoS attack would be to guess the wrong password 20 times, and force the server to shut down :).

micahflee:
Ok @mig5 can you give me a final review on this last commit?

mig5:
@micahflee it works well. There is a slight quirk in the verbose output because we have three web objects that get loaded on startup, and each needlessly generate a random static URL immediately (and then when you actually start a share, it generates yet again a new random URL but only for the current mode's Web object of course). It might be nicer to only generate the static path once a user clicks to start the share, but it's not a big deal..

```
user@onionshare:~/git/onionshare$ ./dev_scripts/onionshare-gui --verbose
OnionShare 2.1 | https://onionshare.org/
[May 30 2019 11:24:39 AM] Onion.__init__
[May 30 2019 11:24:39 AM] OnionShare.__init__
[May 30 2019 11:24:39 AM] OnionShareGui.__init__
[May 30 2019 11:24:39 AM] Settings.__init__
[May 30 2019 11:24:39 AM] Settings.load
[May 30 2019 11:24:39 AM] Settings.load: Trying to load /home/user/.config/onionshare/onionshare.json
[May 30 2019 11:24:39 AM] Web.__init__: is_gui=True, mode=share
[May 30 2019 11:24:39 AM] Web.generate_static_url_path: new static_url_path is /static_zgf6jbhb6opvlgwljtlreerici
[May 30 2019 11:24:39 AM] ShareModeWeb.__init__
[May 30 2019 11:24:40 AM] ShareMode.update_primary_action
[May 30 2019 11:24:40 AM] Web.__init__: is_gui=True, mode=receive
[May 30 2019 11:24:40 AM] Web.generate_static_url_path: new static_url_path is /static_yezzgrdr75esi5odqrey4el33m
[May 30 2019 11:24:40 AM] ReceiveModeWeb.__init__
[May 30 2019 11:24:40 AM] Web.__init__: is_gui=True, mode=website
[May 30 2019 11:24:40 AM] Web.generate_static_url_path: new static_url_path is /static_4cj6i6kilfzqekpxurvrwqecby
[May 30 2019 11:24:40 AM] WebsiteModeWeb.__init__
[May 30 2019 11:24:40 AM] WebsiteMode.update_primary_action
[May 30 2019 11:24:40 AM] TorConnectionDialog.__init__
[May 30 2019 11:24:40 AM] TorConnectionDialog.start
[May 30 2019 11:24:40 AM] TorConnectionThread.__init__
[May 30 2019 11:24:40 AM] TorConnectionThread.run
[May 30 2019 11:24:40 AM] Onion.connect
[May 30 2019 11:24:41 AM] Onion.connect: Connected to tor 0.3.5.8
[May 30 2019 11:24:41 AM] TorConnectionDialog._connected_to_tor
```

:+1:

PR #1003, opened 2019-06-12
Added python-flask-httpauth depend

wh1t3fang:
added flask httpauth depend to the install instructions for debian.  Otherwise, this exception appears.

ImportError: No module named 'flask_httpauth'

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -14,7 +14,7 @@ Install the needed dependencies:
 For Debian-like distros:
 
 ```
-apt install -y python3-flask python3-stem python3-pyqt5 python3-crypto python3-socks python-nautilus tor obfs4proxy python3-pytest build-essential fakeroot python3-all python3-stdeb dh-python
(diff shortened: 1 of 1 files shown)


mig5:
I think this is only for Python 2, so it wouldn't work.

But it's a good point, we need to be able to depend on a package if we want to use the switch to HTTP auth instead of slug.

mig5:
@micahflee  the problem is that `python3-flask-httpauth` as a package doesn't exist in many distro releases.

E.G it's in Debian Buster (stable), Testing and Sid, but not Debian Stretch ('old stable' but very much still recent - even Debian Jessie before that still gets LTS)

https://packages.debian.org/search?keywords=python3-flask-httpauth

And it's only in 2 Ubuntu releases (disco and eoan) which are very new

https://packages.ubuntu.com/search?keywords=python3-flask-httpauth


So I worry we will have packaging issues or will have to drop very recent versions of Ubuntu including bionic

PR #1017, opened 2019-09-01
Hide Tor settings with ONIONSHARE_HIDE_TOR_SETTINGS

### LINKED ISSUE #1014: Provide some way to hide the Tor connection prefs in the Settings dialog
Hi! In the context of Tails, we pre-configure OnionShare to use the system instance of tor. We need to do this because a tor process started by OnionShare (as the `amesia` user) would not be allowed to connect to the Tor network: our firewall would forbid it; and in any case, it's nice that bridges settings already configured by the user at the system level are automatically used for OnionShare connections.

But now we're facing a UX problem since recent versions of OnionShare expose Tor connection prefs in the _Settings_ dialog: as my colleague sajolida wrote on https://redmine.tails.boum.org/code/issues/16913, "in the context of Tails, these options are, at best, not useful, probably confusing, and possibly harmful if changing them can lead to a broken OnionShare". I suspect Whonix is affected by similar problems as well.

Could you please add a way for Tails to hide these prefs?

For example, if the `LOCK_DOWN_TOR_CONNECTION_SETTINGS` environment variable is set to 1, OnionShare would use whatever Tor connection settings are set in its configuration file, and would hide the corresponding prefs in the _Settings_ dialog.

micahflee:
Resolves #1014. If the environment variable `ONIONSHARE_HIDE_TOR_SETTINGS=1`, the Tor connection settings just don't get displayed in the GUI.

Screenshot from 2019-09-01 13-17-37

### DIFF
--- a/onionshare_gui/settings_dialog.py
+++ b/onionshare_gui/settings_dialog.py
@@ -18,7 +18,11 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from PyQt5 import QtCore, QtWidgets, QtGui
-import sys, platform, datetime, re
+import sys
+import platform
+import datetime
+import re
+import os
 
 from onionshare import strings, common
 from onionshare.settings import Settings
@@ -27,6 +31,7 @@ from onionshare.onion import *
 from .widgets import Alert
 from .update_checker import *
 from .tor_connection_dialog import TorConnectionDialog
+
 
 class SettingsDialog(QtWidgets.QDialog):
     """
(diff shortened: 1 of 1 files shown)


maqp:
@micahflee @mig5 @Baccount 

In https://redmine.tails.boum.org/code/issues/14649 the Tails developers talk about delaying the inclusion of OnionShare 2.1:

> This will become FT work for 5.0, because Debian Bullseye will include OnionShare 2.x. If someone finds time to do it earlier, for example in order to fix #16913, or for some other reason, great!

The ticket #16913 in question is titled "Hide Tor settings in OnionShare" that IIUC relates to this ticket. 

The onion-grater settings that were a problem for Tails 4 should now be easy to solve (I wrote some regexes for the Tails ticket #14649) so I think there's still time to include OnionShare 2.1 in Tails 4.

Issue #1007, opened 2019-07-09
Install fails on Raspbian due to torbrowser-launcher not being installable


LosEagle:
There is onionshare version in Raspbian repos which cannot be installed due to
```
The following packages have unmet dependencies:
 onionshare : Depends: torbrowser-launcher but it is not installable
E: Unable to correct problems, you have held broken packages.
```
As far as I understand, onionshare does not actually need torbrowser to function. System tor should be enough so why does it depend on this package? Is there any workaround for this whatsover?

mig5:
According to https://github.com/micahflee/onionshare/blob/develop/stdeb.cfg OnionShare doesn't depend on torbrowser-launcher directly.

It is apparently already fixed by the downstream Debian maintainer,  see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=916889 

My guess is Raspbian is not based on new enough version of Debian to benefit from the change in Debian.

>  Is there any workaround for this whatsover?

I recommend cloning the git repo, checking out the master branch, and running `install/build_deb.sh`, from which you'll get your own .deb package which you can then install with `dpkg -i`.

LosEagle:
Raspbian seems to lack certain python dependencies which prevents me from building it there directly. I've tried building it in virtual Lubuntu environment with success but it still can't be installed without those dependencies. I guess I'm going to have to wait for newer version of Raspbian or Flatpak release then :(

micahflee:
No longer an issue

Issue #910, opened 2019-02-19
Flatpak support


Eonfge:
# Intro
Flatpak is a framework for distributing desktop applications on Linux. It has been created by developers who have a long history of working on the Linux desktop, and is run as an independent open source project.

[Introduction: Reasons to use Flatpak](http://docs.flatpak.org/en/latest/introduction.html#reasons-to-use-flatpak)

# Reason
In short, this: https://repology.org/metapackage/onionshare/versions

A good 80% of all Linux versions out there is running an older, possibly insecure version of Onionshare. As such, it would be preferable to move current deployment effort over to Flatpak. This would initially take a bit more effort, but in the long run it would ensure that all distributions always have the latest version of your application.

micahflee:
I agree, the Linux packaging situation is a mess. How does Flatpak compare with Snapcraft? https://snapcraft.io/

RoestVrijStaal:
I'd recommend AppImage, since at user's end, it doesn't require any "package manager", nor (activation of a) "runtime" nor (elevated) commands to run the application, in contrast with Flatpak or Snappy.

 https://docs.appimage.org/introduction/index.html

Eonfge:
Uhm. Flatpak vs Snap is still an unsettled battle. It's clear now that AppImage doesn't seem to cut it because it misses a centralized upgrading and distribution mechanism, but as far as FlatPak and Snap goes, all I can do, is point to research of others.

**Large overview by AppImage themselves:**
https://github.com/AppImage/AppImageKit/wiki/Similar-projects#comparison
_In my point of view there is some bias in this breakdown, but most elements are accurate. Reader discretion advices._

**Comparison Flatpak and Snap cross distro support (2018):**
https://kamikazow.wordpress.com/2018/06/08/adoption-of-flatpak-vs-snap-2018-edition/

**Theming and desktop integration (2017):**
https://aozoeky4dglp5sh0-zippykid.netdna-ssl.com/wp-content/uploads/2017/07/corebird-app-theming.jpg

**Theming and desktop integration (2018):**
https://www.youtube.com/watch?v=V_3e_VWCGlU

[...]

PR #1018, opened 2019-09-01
Add python3-distutils as a dependency

### LINKED ISSUE #1010: Dependency issue in Ubuntu 18.04
I got a report that this happens when trying to run OnionShare in Ubuntu 18.04 LTS:

```onionshare-gui 
  File "/usr/bin/onionshare-gui", line 21, in <module>
    import onionshare_gui
  File "/usr/lib/python3/dist-packages/onionshare_gui/init.py", line 29, in <module>
    from onionshare.common import Common
  File "/usr/lib/python3/dist-packages/onionshare/init.py", line 27, in <module>
    from .onion import *
  File "/usr/lib/python3/dist-packages/onionshare/onion.py", line 21, in <module>
    from stem.control import Controller
  File "/usr/lib/python3/dist-packages/stem/control.py", line 265, in <module>
    import stem.descriptor.microdescriptor
  File "/usr/lib/python3/dist-packages/stem/descriptor/init.py", line 55, in <module>
    import stem.util.system
  File "/usr/lib/python3/dist-packages/stem/util/system.py", line 70, in <module>
    import distutils.spawn
ModuleNotFoundError: No module named 'distutils.spawn'
```

micahflee:
Resolves #1010.

This just adds `python3-distutils` as a dependency. That package didn't exist in debian stretch, but that's no longer stable, and it does exist in buster, as well as several old ubuntus. So I think it's fine to depend on it now.

(Also, I removed redundant packages from `Build-Depends` that were already in `Depends3`.)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -14,7 +14,7 @@ Install the needed dependencies:
 For Debian-like distros:
 
 ```
-apt install -y python3-flask python3-stem python3-pyqt5 python3-crypto python3-socks python-nautilus tor obfs4proxy python3-pytest build-essential fakeroot python3-all python3-stdeb dh-python python3-flask-httpauth
+apt install -y python3-flask python3-stem python3-pyqt5 python3-crypto python3-socks python-nautilus tor obfs4proxy python3-pytest build-essential fakeroot python3-all python3-stdeb dh-python python3-flask-httpauth python3-distutils
 ```
 
 For Fedora-like distros:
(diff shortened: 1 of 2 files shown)


micahflee:
I think I'm ok with dropping support for stretch for the next OnionShare release. It's not like stretch will ever include OnionShare 2.2 anyway, and hopefully soon-ish we'll have a Flatpak package anyway (#910) which should run on stretch, and simplify all these linux packaging woes.

An alternative also would be to implement our own HTTP basic auth, which isn't all that complicated, but using an existing python module is definitely preferred.

mig5:
Yeah, it might lead to more issues like #1007 (e.g other Debian specialist flavours like Raspbian if people can't build their own deb) but maybe Flatpak can help there too. And there's always pip and  the `dev_scripts` solution for those special cases, which should work just fine if less 'pristine' than a .deb

PR #1019, opened 2019-09-01
Update deps

micahflee:
This updates pip dependencies, and also for macOS and Windows it uses a newer version of Python3 and PyQt5. Also, it seems that PyInstaller can be installed in macOS fine, so I'm switching the instructions to install from pip instead of build from source.


micahflee:
I just re-ran the test in CircleCI and they all passed. Sweet.

Issue #997, opened 2019-05-25
Write tests for website mode


micahflee:
Right now website mode (#791) code is merged into develop, but it's not tested yet. We should have some tests before releasing.

micahflee:
Fixed in #1020

Issue #993, opened 2019-05-24
Website mode's history widget should look like web logs


micahflee:
From https://github.com/micahflee/onionshare/pull/961#issuecomment-494088431:

> History widget: This hasn't been mentioned yet, but the history widget (clicking the little arrow button to expand the widget on the right) I think should look more like web logs. Right now it says "Someone has visited your website on [date]" but I think instead each web request should be something like: `2019-05-20 10:42:58 GET / 200`, or `2019-05-20 10:43:02 GET /favicon.ico 404`

micahflee:
Fixed in #1020

PR #1020, opened 2019-09-01
[WIP] Share code between share mode and website mode

### LINKED ISSUE #991: Make ShareMode and WebsiteMode share code
From https://github.com/micahflee/onionshare/pull/961#issuecomment-494088431:

> `WebsiteMode.set_file_info` right now builds up `self.files` and `self.root_files`, which is used in the logic to choose when to render directory listings and when to render html files. And `WebsiteMode.directory_listing` displays the actual directory listing. Both of these functions would need to be available to `ShareMode`, but `ShareMode.set_file_info` (which does a totally separate thing, compresses the files) would still need to exist. So to merge these two, maybe it would make sense to make a new `BaseShareMode` class which implements all of the shared functionality, and then make `ShareMode` and `WebsiteMode` inherit from `BaseShareMode`.

micahflee:
This resolves #991. This is a fork of @hiromipaw's PR #1002, but is slightly more complete -- so if this gets merged, that will also get merged automatically.

- I renamed `BaseModeWeb` to `SendBaseModeWeb` because this is only a base class for the send modes (as opposed to receive mode)
- I moved a bunch of share mode and website mode specific code out of the base class and into the share and website classes
- I added back `Web.generate_static_url_path` so that each share gets a different random static URL path
- Also it doesn't quite fit here, but I tweaked the css to change the style of directory listing links


mig5:
> I don't reproduce this in OnionShare 2.1 so I think it's some regression in the sharing of Share and Website mode code, with the FileList.

I also don't reproduce it on `develop` branch, so indeed the FileList bug must be specific to this PR.

micahflee:
>     1. As mentioned in Keybase Chat:
> 
> if I have 'stop sharing after files have been sent' enabled (as is the default), but share a .txt file (or in fact, any file, such as a .png image) with myself in normal Share Mode, I can click the text file hyperlink and view the contents of the file in Tor Browser. Or, in the case of an image, the browser displays the image, and I can right-click and save it to my computer, likewise I can copy/paste the contents of a .txt file and save it to my computer. But the Share keeps running because this viewing of the individual file didn't trigger the 'download' logic as such. That seems contrary to expectations of 'stop sharing after first download' which really means 'stop sharing after the user obtains the file (in some form)', do you agree?
> 
> Maybe the individual file viewing should be for Website Mode only after all? A shame as I know we liked the idea of 'download individual files' when in Share Mode as that's been a requested feature before. Or I wonder if we can trigger the 'download' logic when an individual file is viewed 'directly' in the browser? It's still a GET request after all.

This is a really good point. I don't think that it will work to have the stop download login happen when downloading an individual file, because what if there are 50 files. It doesn't make sense for the server to stop after someone downloads any 1 of them, but what if someone downloads 49 of them but doesn't download the last? The server will never stop.

So as I said in keybase, I think the best solution is to add a new checkbox, "Allow downloading individual files" that can only be enabled if "Stop sharing after files have been sent" is unchecked. And the "Allow downloading individual files" checkbox can be disabled unless "Stop sharing after files have been sent" is unchecked, which prevents users from even using the feature at all unless they've manually unchecked the "Stop sharing ..." checkbox.

This means that there will have to be a new way of rendering the templates -- one where the files are links, and one where they aren't links. I still like the idea of allowing the Tor Browser user to navigate the file tree though. And if "Allow downloading individual files" isn't checked, then I need to make sure it actually doesn't work, not just removing the links to the files.

I'll work on implementing this.

>     1. I noticed a bug in Share mode with the FileList, whereby if I:
> 
> a) share a file (in this example, verbose.txt),
> b) then stop the share,
> c) click the [ X ] widget to remove the file from the FileList
> d) add a different file (in this case, tor.png) and start the share
> 
> ... the original file is still listed at the onion service when I view it in Tor Browser, in addition to the new file. Even though in the OnionShare GUI, I only see the new file in the FileList widget.
> 
> two-files-instead-of-just-tor-png
> 
> If I rinse/repeat and replace the file list with a third file, I now see three files in the subsequent share.
> 
> three-files-instead-of-just-tor-png
> 
> In short, the list of files to share is somehow not being reset internally on subsequent shares, even though it is so visually in the UI.
> 
> I don't reproduce this in OnionShare 2.1 so I think it's some regression in the sharing of Share and Website mode code, with the FileList.

Good catch! I'll try to see what's going on and fix this.

> Finally, this is just a small note but we have a mix of terminology in the code now, 'Send' used as part of class names, but which is inside 'share_mode' and which is also 'Share' in the UI. Should we standardise on 'Send' in the UI do you think, now that there is 'Receive' (Send and Receive seem a better match of terms? One for Glenn?) Or, standardise by renaming things like SendBaseModeWeb to ShareBaseModeWeb ? But it's not a big deal.

Hmm, maybe we should rename share mode to send mode, though I do think it's nice that it's called "Share" since it's OnionShare. But actually I used the word "Send" in "SendBaseModeWeb" specifically as a way to describe the common functionality between share mode and website mode -- both share and website are sending files, where-as receive mode is receiving them.

I'm ok with leaving it how it is for now. But I'm also curious if @glenn-sorrentino has an opinion :).

mig5:
> And if "Allow downloading individual files" isn't checked, then I need to make sure it actually doesn't work, not just removing the links to the files.

We should be able to do that in a unit test by trying to check the webpage HTML (e.g make sure there's no href on the line that matches a file name being shared)  Tests FTW. We can also test the checkbox isn't checkable if another one is enabled etc, I think we do that in some other tests already (maybe the 'Legacy mode' stuff).

micahflee:
I have fixed the bug where if you share files, stop the server, change the files, and start the server again, the first files are still there.

I also added a the new "Allow downloading of individual files" setting, and made it work as described above. If it's checked, and "Stop sharing after files have been sent" is unchecked, you can't download individual files.

The one thing I didn't do is write new unit tests to test this new functionality.

micahflee:
Here's what the settings dialog looks like:

Screenshot from 2019-09-01 20-57-58

micahflee:
To do:
- If someone downloads an individual file, this should probably be shown somewhere in the GUI (lthough making a progress bar could be quite complicated)
- If someone downloads an individual file, does it get http gzipped? I don't think it does, and it probably should
- Write some tests

onionshare/web/share_mode.py
```diff
@@ -16,9 +16,8 @@ class ShareModeWeb(SendBaseModeWeb):
     """
     def init(self):
         self.common.log('ShareModeWeb', 'init')
-        # If "Stop sharing after files have been sent" is unchecked and "Allow downloading of individual files" is checked
-        self.download_individual_files = not self.common.settings.get('close_after_first_download') \
-            and self.common.settings.get('share_allow_downloading_individual_files')
+        # Allow downloading individual files if "Stop sharing after files have been sent" is unchecked
+        self.download_individual_files = not self.common.settings.get('close_after_first_download')
```
mig5: @micahflee we might not need this overridden `init()` class at all now? I can't see where this attribute `self.download_individual_files` is evaluated now.

micahflee: No I think we still need it. It gets used later in the `directory_listing_template` and `render_logic` methods, basically to tell the template if the filename should be a link, and to prevent people from downloading the file if they just type in the URL, even though it's not a link.

mig5: Yep my bad, I was viewing an individual commit instead of the full diff :sigh:

mig5:
>     * If someone downloads an individual file, this should probably be shown somewhere in the GUI (lthough making a progress bar could be quite complicated)

I pushed a commit to my copy of this branch that registers a simple HistoryItem (similar to the simplified HistoryItem of the Website Mode). This works based on the `web.REQUEST_LOAD` event, which only passes the path requested, and so it filters out a couple of paths are considered 'reserved' such as / , /favicon.ico, /download and the randomised static url path.

There's probably a better solution (what if someone really does want to share and view a file called favicon.ico or 'download' ?) but thought I'd offer it as some inspiration. The commit is here:

https://github.com/mig5/onionshare/commit/173b2d3f5ed1cf37fdb33e675e337eb4350498eb

hiromipaw:
I have noticed a few issues.

First very small:

The page view counter isn't reset when the user click on stop sharing and start sharing something else. Maybe it isn't even an issue. But I just thought I'd mention. 

Second issue:

I think something needs a reset if you stop sharing and then want to share something else. Otherwise you keep sharing the thing you shared the first time.

I don't think these are generally blockers to merge this PR into develop. Maybe it would be nice to fix these before the next release.

Everything else works great. 

Thanks @micahflee.

mig5:
@micahflee I have also now added tests for:

1) Testing that the hyperlink to an individual file is present and that the individual file is fetchable, when in stay_open mode

2) That the above is not true when stay_open is false

3) simple website test (publish an index.html and check that the onion service shows the *contents* of that file as the website, rather than a file listing page)

micahflee:
Thanks for writing tests @mig5! I just merged your code into this branch, and I also just made it so downloading individual files doesn't use flask's `send_from_directory` and instead streams the file from disk manually, making it possible to have a progress bar, and also gzips it if the browser supports it.

So the latest to do list, including the bug @hiromipaw found:

- Improve the IndividualFile History widget, which I realized really is just #993, so we might as well resolve it in this PR too. It's possible now to give this widget a progress bar.
- The page view counter isn't reset when the user clicks on stop sharing and starts sharing something else.

@hiromipaw, you wrote:

> I think something needs a reset if you stop sharing and then want to share something else. Otherwise you keep sharing the thing you shared the first time.

What do you mean? Like if you click stop sharing, you think it should automatically remove the files to share from the file list? It currently just doesn't do that, in case like you stop sharing because you forgot to add one more file or something, and if you want to start over you have to manually remove all the files.

micahflee:
Also, if you're in share mode and you have multiple files, and one of them is called `download`, if someone clicks on it to try to download it, it makes a GET request to `/download` and instead downloads the entire set of files, zipped up. (If you're only sharing one file called `download` it does the same, but not zipped up, so ironically it actually works as expected.)

The obvious way to solve it would be similar to the static files path, instead of `/download` make the path something unguessable, like `/[random-hex]_download`. But I'm kind of leaning towards just keeping this broken edge case, because I like that `/download` is the same every time you share files. This way without loading it in a browser first, you can easily wget it, which I've done before...

What do you all think?

mig5:
I think it's fine to leave it as `/download`, no-one's complained about it up til now :) I wonder about `favicon.ico`, maybe it would be nice to ship a favicon.ico so we don't have to deal with the fact that it 404s (simplifying code since I think you wrote an exception to avoid that tripping the 'guessing URLs' logic). And maybe if someone tries to share a favicon.ico it renames the file to have an underscore or something, before making the share?

mig5:
> The page view counter isn't reset when the user clicks on stop sharing and starts sharing something else.

It is for me?

When I stop the share after viewing the file once, I see the counter as '1' with a tick. But if I start a new share, it's reset.

mig5:
> > The page view counter isn't reset when the user clicks on stop sharing and starts sharing something else.
> 
> It is for me?
> 
> When I stop the share after viewing the file once, I see the counter as '1' with a tick (even after I stop the share). But if I start a new share, it's reset.
> 
> I don't mind seeing the counter after clicking Stop - it's useful for when the share auto-stops and maybe the computer was unattended. You might want to see the total count after it's stopped.

My bad, I thought you meant the in progress/completed counters, not the ToggleHistory indicator.

I have pushed a fix (and a test) for this in my branch, commit https://github.com/mig5/onionshare/commit/09f2f572982f39a411ae82891d677d929c29a52f

micahflee:
I found another bug. The files listed in the directory listing don't stick around between server stop/starts, but the files in the zip file still down. Steps to reproduce:

- In share mode, add `file1.txt` and start the server
- Stop the server, add `file2.txt`, and start the server again
- Download the files
- You'll download a zip with both `file1.txt` and `file2.txt` in it, when you should only be downloading `file2.txt`

I confirmed it's a bug in this branch. It doesn't happen in `develop`.

micahflee:
I've done a bunch more work on #993 and it's starting to look pretty cool. (It's not done yet, and I've introduced some more bugs, but I'm stopping for the night.)

micahflee:
I think this is very nearly done, and when merged it will resolve #991, #993, and #997. In fact, if there are no bugs introduced or anything big I'm missing, I would be fine with merging this now and doing the rest of the final changes in separate PRs.

To recap:

- This introduces `Web.SendBaseModeWeb`, which is the base class for both `Web.ShareMode` and `Web.WebsiteMode`, and it shares most of the code. This allows share mode to take advantage of all of the directory listing stuff that website mode has.
- Thanks to @mig5, there are a lot of tests!
- In share mode, when "Stop sharing after files have been sent" is disabled, individual file downloading works. If it's enabled, then you can still browse directory listing but you can only download the whole zip, not individual files. Here's what the settings dialog looks like:

Screen Shot 2019-09-08 at 4 53 03 PM

- Now all modes (share mode, receive mode, and website mode) display `IndividualFileHistoryItem` widgets in their history, which provides more information to the OnionShare user about how people are actually using their web server:

Screen Shot 2019-09-08 at 12 15 36 PM
Screen Shot 2019-09-08 at 12 23 29 PM
Screen Shot 2019-09-08 at 12 19 22 PM

The only thing I think is left to consider is the "in progress" and "completed" indicator icons in the top right of the history widgets, for all modes. Right now they still work the way they used to work:

- In share mode, "in progress" increments when a download starts (as in, downloading all the files being shared), and "completed" increments when the download finished -- but nothing happens when an individual file is downloaded
- Similarly, in receive mode "in progress" and "completed" only change when files get uploaded, but not when any other requests get made and displayed to the user
- In website mode, there's currently an "in progress" widget but it's always 0, because it never gets incremented

I think maybe there should be a third icon representing web requests (maybe something [like this](https://www.freepik.com/free-icon/internet-world_705410.htm#page=1&query=web&position=0)?) that increments basically every time a new `IndividualFileHistoryItem` appears.

micahflee:
Ok I just added the web requests indicator icon. @mig5 ready for a final review!

micahflee:
Aha, tests are failing because the progress indicator now increments for each web request in addition to downloads/uploads. Let me fix that...

mig5:
I've also added a route for `/favicon.ico` to send our actual favicon.ico, so that it stops 404ing, reducing noise. The exception is for website mode, so that a website can actually have its own favicon.

The only catch is that this therefore makes the favicon.ico a 'reserved' word, e.g it won't be possible to share a file called favicon.ico in Share Mode unless it's with other files and so fetchable as a zip... maybe we could rename a file if found to be called favicon.ico before sending it?

micahflee:
Found (and fixed) another bug. Because `Web.error404` requires passing in `history_id` now, the web server was crashing when loading a 404 error in receive mode. In share and website modes, all routes are caught, and there's logic that calls `error404` for some of them. But this wasn't the case in receive mode.

As soon as I confirm tests pass, I think this looks ready to merge to me too!

Issue #973, opened 2019-05-02
[Feature request] restrict upload file size


sashaman:
probably especially interesting for public/persistent shares

PR #1021, opened 2019-09-03
Upload max size

mig5:
Took a pass at #973

Might be a better way of doing it, not sure.

Note that I discovered the early return in the ReceiveModeRequest object, is not handled too well when it fires. A subsequent interaction with the Flask Request system doesn't like that there are some other attributes that only  get defined afterward (beyond the early return). 

Instead, raising an Exception seems a bit better and stands a chance of the user being told about the error in a flash() message. The trackback in the OnionShare user's terminal (if running via `dev_scripts` for example) is also a bit less messy. Try swapping out that small change back to a `return` to see what I mean.


mig5:
I have fixed the conflicts and the showing of the max upload size in the HTML. I can't figure out how to resolve the latter issue, that was what I tried to explain in the original description of the PR (The only way I could get Flask to not throw a trackback error when handling the request, was to to raise an Exception instead of a `return`, which in doing so automatically renders the 'error' logic in your javascript, but in a way that is not configurable).

Every attempt I've made to try and render a flash message on that error after computing the size of the request's content length, has failed.. let me know if there's an obvious fix I'm missing

micahflee:
Cool, well I'll try to take a stab at it after we finish the 2.2 release.

PR #1024, opened 2019-09-12
Don't avoid not having the meaning of this phrase backwards.

ln5:


micahflee:
Thanks for contributing this @ln5. Do you mind though instead of a git pull request making this change directly in Weblate, which is the system we use for localization?

See:
- https://github.com/micahflee/onionshare/wiki/Translating
- https://hosted.weblate.org/projects/onionshare/translations/

Thanks.

ln5:
I did something in Weblate.
Seems like I managed to create a suggestion for a change to the Swedish translation of gui_settings_close_after_first_download_option: https://hosted.weblate.org/translate/onionshare/translations/sv/?q=dela+efter+att&search=ftx&source=1&target=1&type=all&offset=1#suggestions
LMK if this helps.

PR #1025, opened 2019-09-15
Breadcrumbs

### LINKED ISSUE #994: Directory listing should have breadcrumbs
From https://github.com/micahflee/onionshare/pull/961#issuecomment-494088431:

> Directory listing page should display the path that it's displaying, and if it's not the root should at least have a "Parent directory"/`..` link, but would be cool to have full breadcrumbs too

micahflee:
Resolves #994.

This adds breadcrumbs to both share mode and website mode.

breadcrumbs

### DIFF
--- a/onionshare/web/send_base_mode.py
+++ b/onionshare/web/send_base_mode.py
@@ -90,9 +90,15 @@ class SendBaseModeWeb:
             'status_code': 200
         })
 
+        breadcrumbs = [('☗', '/')]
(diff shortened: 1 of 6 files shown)


micahflee:
> We can do that later though (just trying to get us into the habit of adding tests as we introduce new features)

Definitely a good habit to get into! I should make myself more in that habit too... but I think for the moment I'm just going to merge this one :)

PR #1030, opened 2019-09-16
Make setting the Content-Security-Policy header optional so it doesn't break website mode shares

### LINKED ISSUE #1029: Make the Content Security Policy header optional (for Website Mode)
Right now, we enforce strict CSP headers in the Flask web app.

```
        self.security_headers = [
            ('Content-Security-Policy', 'default-src \'self\'; style-src \'self\'; script-src \'self\'; img-src \'self\' data:;'),
            ('X-Frame-Options', 'DENY'),
            ('X-Xss-Protection', '1; mode=block'),
            ('X-Content-Type-Options', 'nosniff'),
            ('Referrer-Policy', 'no-referrer'),
            ('Server', 'OnionShare')
        ]
```

We should make the first one enabled by default, but optional in the Settings, so that users wanting to use Website Mode can disable it if it breaks their sites (3rd party javascript libraries, CDNs, etc)

mig5:
Fixes #1029


mig5:
Thanks and sorry, I really should've thought of all of that, it was sub-standard.. I think I have addressed all those issues now.

Issue #953, opened 2019-04-12
Should Onionshare be notarized for macOS?


Baccount:
This will help give users more confidence that the software they download and run, no matter where they get it from, is not malware by showing a more streamlined Gatekeeper interface.
> With the public release of macOS 10.14.5, we require that all developers creating a Developer ID certificate for the first time notarize their apps.

https://developer.apple.com/news/?id=04102019a

micahflee:
Good idea. Here is further documentation on how the process is actually done:
https://developer.apple.com/documentation/security/notarizing_your_app_before_distribution

micahflee:
I want to add notarizing the app as part of the macOS release process, but to start learning it I decided to notarize the OnionShare 2 installer. I figured this issue is a good place to start documenting what I learn.

[These docs](https://developer.apple.com/documentation/security/notarizing_your_app_before_distribution/customizing_the_notarization_workflow#3087734) say:

> The notary service generates a ticket for the top-level file that you specify, as well as each nested file. For example, if you submit a disk image that contains a signed installer package with an app bundle inside, the notarization service generates tickets for the disk image, installer package, and app bundle.

The OnionShare release script makes a codesigned `.pkg` installer file, and inside that is a codesigned app bundle. So I assume I can just notarize the `.pkg` and it will notarize both the installer and the app bundle.

I made an app-specific password in my Apple ID account, and saved it to my login keychain under the name "onionshare-notarize".

Then I ran this to notarize it:

```
$ xcrun altool --notarize-app --primary-bundle-id "com.micahflee.onionshare" --username "micah@micahflee.com" --password "@keychain:onionshare-notarize" --file OnionShare-2.pkg 
2019-04-19 14:08:53.912 altool[27377:660911] No errors uploading 'OnionShare-2.pkg'.
RequestUUID = 42504659-e533-47cb-8bd1-87ff12dd0275
```

I can see all of my notarization jobs like this:

```
$ xcrun altool --notarization-history 0 --username "micah@micahflee.com" --password "@keychain:onionshare-notarize"

Notarization History - page 0

Date                      RequestUUID                          Status      
------------------------- ------------------------------------ ----------- 
2019-04-19 21:08:49 +0000 42504659-e533-47cb-8bd1-87ff12dd0275 in progress 

Next page value: 1555708129000
```

And I can find detailed info like this:

```
[...]

PR #1033, opened 2019-09-21
Update dependencies and build instructions

### LINKED ISSUE #1027: Update dependencies, and Tor binaries
For version 2.2, we should update to Qt 5.13.1, as well as other pip dependencies. And right now we're getting Tor binaries from Tor Browser 8.0.8, but the latest Tor Browser is 8.5.5.

micahflee:
Resolves #1027


BUILD.md
```diff
@@ -100,7 +100,7 @@ Open a command prompt, cd to the onionshare folder, and install dependencies wit
 pip install -r install\requirements.txt
 ```
 
-Install the Qt 5.13.0 from https://www.qt.io/download-open-source/. I downloaded `qt-opensource-windows-x86-5.13.0.exe`. In the installer, you can skip making an account, and all you need `Qt` > `Qt 5.13.0` > `MSVC 2017 32-bit`.
+Install the Qt 5.13.0 from https://www.qt.io/offline-installers. I downloaded `qt-opensource-windows-x86-5.13.1.exe`. In the installer, you can skip making an account, and all you need `Qt` > `Qt 5.13.1` > `MSVC 2017 32-bit`.
```
mig5: @micahflee tiny nitpick: 'Install the Qt 5.13.0' should now be 'Install the Qt 5.13.1'. Maybe we could rephrase this to 'Install version 5.13.1 of Qt'.

BUILD.md
```diff
@@ -50,7 +50,7 @@ Download and install Python 3.7.4 from https://www.python.org/downloads/release/
 
 You may also need to run the command `/Applications/Python\ 3.7/Install\ Certificates.command` to update Python 3.6's internal certificate store. Otherwise, you may find that fetching the Tor Browser .dmg file fails later due to a certificate validation error.
 
-Install Qt 5.13.0 for macOS from https://www.qt.io/offline-installers. I downloaded `qt-opensource-mac-x64-5.13.0.dmg`. In the installer, you can skip making an account, and all you need is `Qt` > `Qt 5.13.0` > `macOS`.
+Install Qt 5.13.0 for macOS from https://www.qt.io/offline-installers. I downloaded `qt-opensource-mac-x64-5.13.1.dmg`. In the installer, you can skip making an account, and all you need is `Qt` > `Qt 5.13.10` > `macOS`.
```
mig5: 'Install Qt 5.13.1' (not 5.13.0) - also is '5.13.10' a typo?

micahflee:
I'd totally be into including Tor 0.4.1.6, but I just checked and it's not in the latest Tor Browser alpha yet (9.0a6), which means the only way to use it right now is to build from source. Which we're going to need to start doing anyway for #953 -- but we haven't yet, so I'm fine with sticking with Tor 0.4.1.5 until the next release.

PR #1060, opened 2019-10-19
Use ==/!= to compare str, bytes, and int literals

cclauss:
https://docs.python.org/3/whatsnew/3.8.html#changes-in-python-behavior

The compiler now produces a SyntaxWarning when identity checks (is and is not) are used with certain types of literals (e.g. strings, numbers). These can often work by accident in CPython, but are not guaranteed by the language spec. The warning advises users to use equality tests (== and !=) instead.


micahflee:
Thanks!

PR #1063, opened 2019-10-21
Use python 3.6 f-strings

micahflee:
Python 3.6 supports f-strings, which are much nicer to use compared to `.format()` for string formatting. This PR switches to using f-strings wherever possible, which basically means anywhere we need to do string formatting that isn't on a localized string -- I still use `.format()` for those.

More about f-strings: https://realpython.com/python-f-strings/

Basically, it looks like this:

```python
print(f"Hello, {first_name} {last_name}. Nice to meet you.")
```

Instead of this:

```python
print("Hello, {} {}. Nice to meet you.".format(first_name, last_name))
```

Because f-strings are only available in python 3.6+, this PR removes support for python 3.5, so removes the python 3.5 tests from circleci as well, and only tests with 3.6 and 3.7.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,14 +7,13 @@ workflows:
   version: 2
   test:
     jobs:
-      - test-3.5
       - test-3.6
       - test-3.7
 
 jobs:
-  test-3.5: &test-template
+  test-3.6: &test-template
     docker:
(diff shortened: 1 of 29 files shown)


micahflee:
True. I have high hopes though that nearly all of our Linux support problems will go away once we finally do flatpak packaging :).

Issue #1074, opened 2019-11-28
Tor tests


micahflee:
While working on tabs (#1064) I've refactored all of the GUI tests. I think they're much cleaner and easier to implement new tests. But I didn't rewrite the Tor GUI tests We rarely run the Tor GUI tests because they aren't enabled in CI, and in fact they're currently failing, even for version 2.2.

Is it worth re-implementing Tor GUI tests using the new way tests have been refactored? Or should we just leave it and not worry about Tor GUI tests?

imstanz:
As a user...
I do want to run any kind of test.
Tho, as of today - I'm not there yet :)

Issue #1073, opened 2019-11-28
New design for tabs


micahflee:
Adding tab support (#1064) is a major issue that includes a lot of refactoring, and it will also involve doing a lot of design work. This issue is for the design work, and will have a separate PR from the tabs themselves.

PR #1075, opened 2019-11-28
Add support for tabs

### LINKED ISSUE #1064: Support for tabs
I'd like to make the same OnionShare window support multiple modes simultaneously, similar to browser tabs. So you can have a persistent receive tab open, a few persistent website tabs, but still be able to share files if you want.

Each tab will have its own settings. Right now the global settings dialog has a left side that has general and onion settings (which apply to all modes), and then settings for sharing, receiving, and website modes. The global settings dialog should still have all of these, but they should be renamed to default settings. Then, when you open a new tab and choose the mode (for example, a share files tab), it will start with the default share settings, but you can modify them for that tab.

If you quit OnionShare and open it again, you should start out with all of your persistent tabs open automatically.

If you run `onionshare-gui` and there’s no OnionShare window open, it should open a new one. If there is a window open, it should bring that window to the front. If you pass filenames into your `onionshare-gui` command, it should open a new tab with those files already filled out. (Basically, the same way web browsers behave.) This will require figuring out a way to communicate between the new `onionshare-gui` process and the already open `onionshare-gui` process.

When using bundled Tor, right now the Tor data directory is set to a random temporary folder. This means each time you load OnionShare you have to re-build the Tor cache. If we can guarantee only one OnionShare process running at a time (by opening a new tab in the existing process instead of opening a separate process), this means we can finally use a persistent Tor data dir (like Tor Browser does), which means opening OnionShare and connecting to Tor will be much quicker, and also more reliable for bridge users.

However the CLI version, `onionshare`, should still use randomized Tor data dirs. Then way you can have multiple tabs terminals open, running multiple `onionshare`s, and they won’t clobber each others’ data dirs.

micahflee:
This implements tabs, which resolves #1064!

This is a major PR that refactors a lot of stuff. There's still some more work to be done on tabs, but I'd like to merge this sooner rather than later so it's not holder up other work.

This includes reworking the design (#1073), which is being discussed in #design in the keybase team, with some great UX work thanks to @glenn-sorrentino. Also deciding if we should make Tor GUI tests work or not (#1074).

There's a lot to cover in explaining this PR, but the main different is this: Before, the main window had the option of choosing between three different modes (share, receive, website) by selecting big purple buttons at the top. Now the main window has a tabs widget. You can create new tabs, close tabs, and re-order them. And each tab is basically like the old main window: you can select a mode. But once you've selected a mode, you can't chose the mode of that tab -- you just have to use a different tab if you want a new mode.

I refactored the code so and moved most of what was in the `MainWindow` class into a `Tab` class. So we still have only one `MainWindow`, but we have as many `Tab`s as we want.

Another big change is how persistence works. When you turn on persistence in a tab (by checking "Save this tab, and automatically open it when I open OnionShare") then it saves that tab to a file in the `~/.config/onionshare/persistent` dir. If you close OnionShare and then re-open it, all persistent tabs automatically open.

Now that there are tabs, and in particular persistent tabs, it doesn't make sense to have two separate OnionShare windows anymore -- you don't want them both to open full of the same persistent tabs. So now, when `onionshare-gui` runs it tries to detect if there's another `onionshare-gui` process already open, and if there is it opens a new tab in that process and then quits itself. (Process detection is done using `psutil`, so it introduces a new python dependency.)

The way this interprocess communication is implemented using a concept called "events". There's a folder (`~/.config/onionshare/events`) that OnionShare watches for changes (using `watchdog`, another new python dependency). Specifically, it watches the file `~/.config/onionshare/events/events` for incoming events (one JSON string per line). So, when a second `onionshare-gui` process wants to open a tab in the first process, it does this by writing JSON to that file. (Note that testing this only works if you install OnionShare system-wide.)

Now that we know there will only ever be one `onionshare-gui` open at a time, this means we can rely on only one Tor process being run at a time, which means finally we can have the Tor data directory stay persistent -- in this case, `~/.config/onionshare/tor_data/`. This stores the Tor cache and makes it so it's quicker and more reliable to connect to Tor, especially if you're using bridges. However, it's still possible that a user could have one GUI open but multiple `onionshare` CLIs going in different terminals, so the CLI version still creates its own Tor data dir in a temporary folder.

Another major refactor is the tests. I've reimplemented all of the logic from the old CLI and GUI tests, but in a way that supports tabs, and I've also written new tests that test the tab functionality itself (`test_gui_tabs.py`). I didn't reimplement Tor tests. The tests are much nicer now in that writing a new test involves writing a function in one of the test classes, rather than making a copy of a whole file and modifying it.

I haven't done thorough testing in Windows and Mac yet...


mig5:
After that last commit, it tried to connect to built-in Tor method but got this error in the modal dialog:

```
OnionShare could not connect to Tor:
socket connection failed ([Errno 2] No such file or directory)

Try changing how OnionShare connects to the Tor network in the settings.
```

After this, I tried socket mode (Tor is running as a daemon in my Qube), which succeded to connect, but then crashed immediately with this:

```
  File "/home/user/git/onionshare/onionshare_gui/tab/tab.py", line 385, in timer_callback
    mode = self.receive_mode
AttributeError: 'Tab' object has no attribute 'receive_mode'
Aborted
```

If I try to run `dev_scripts/onionshare-gui` again, I get the above error straight away.

micahflee:
Good job immediately finding bugs lol. I think I just fixed the second issue you're having with the commit I just pushed.

For the Tor error, I'm not sure what's causing this, but you can fix it by deleting `~/.config/onionshare/tor_data/` and re-opening OnionShare. I should figure out the underlying cause, but also OnionShare could also wipe this if it ever has an issue connecting to Tor.

onionshare/__init__.py
```diff
@@ -149,42 +193,86 @@ def main(cwd=None):
...
+        if mode == "receive":
+            if data_dir:
+                mode_settings.set("receive", "data_dir", data_dir)
+        if mode == "website":
+            mode_settings.set("website", "disable_csp", disable_csp)
+    else:
+        # See what the persistent mode was
+        mode = mode_settings.get("persistent", "mode")
+
+    # In share an website mode, you must supply a list of filenames
```
mig5: minor code comment typo 'In share an website mode'

onionshare/__init__.py
```diff
@@ -284,10 +371,7 @@ def main(cwd=None):
             print("")
 
     # Start OnionShare http service in new thread
-    t = threading.Thread(
-        target=web.start,
-        args=(app.port, stay_open, common.settings.get("public_mode"), web.password),
-    )
+    t = threading.Thread(target=web.start, args=(app.port,))
```
mig5: is trailing comma a problem or not a big deal?

onionshare/mode_settings.py
```diff
@@ -0,0 +1,143 @@
...
+"""
+import os
+import pwd
+import json
+
+
+class ModeSettings:
+    """
+    This stores the settings for a single instance of an OnionShare mode. In CLI there
+    is only one TabSettings, and in the GUI there is a separate TabSettings for each tab
```
mig5: 'TabSettings' or 'ModeSettings' ?

onionshare/web/web.py
```diff
@@ -60,10 +60,16 @@ class Web:
     REQUEST_OTHER = 13
     REQUEST_INVALID_PASSWORD = 14
 
-    def __init__(self, common, is_gui, mode="share"):
+    def __init__(self, common, is_gui, mode_settings, mode="share"):
+        """
+        tab_settings_get and tab_settings_set are getter and setter functions for tab settings
```
mig5: I can't find these functions 'tab_settings_get' and 'tab_settings_set'?

onionshare_gui/tab/mode/mode_settings_widget.py
```diff
@@ -0,0 +1,278 @@
...
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+from PyQt5 import QtCore, QtWidgets, QtGui
```
mig5: QtGui doesn't get used in this file according to pyflakes

onionshare_gui/tab_widget.py
```diff
@@ -0,0 +1,225 @@
...
+        self.observer.schedule(self.event_handler, self.common.gui.events_dir)
+        self.observer.start()
+
+    def move_new_tab_button(self):
+        # Find the width of all tabs
+        tabs_width = sum(
+            [self.tabBar().tabRect(i).width() for i in range(self.count())]
+        )
+
+        # The current positoin of the new tab button
```
mig5: minor typo 'positoin'

onionshare_gui/threads.py
```diff
@@ -47,29 +59,28 @@ def run(self):
...
                 self.mode.app.start_onion_service(
-                    await_publication=False, save_scheduled_key=True
+                    self.mode.settings, await_publication=False, save_scheduled_key=True
                 )
                 # wait for modules in thread to load, preventing a thread-related cx_Freeze crash
                 time.sleep(0.2)
                 self.success_early.emit()
                 # Unregister the onion so we can use it in the next OnionThread
-                self.mode.app.onion.cleanup(False)
+                self.mode.app.start_onion_service(self.mode.settings)
```
mig5: Can we clarify why we're calling `start_onion_service` when we're meant to be unregistering the onion?

mig5: I think that's why I get the error 'There was an error with Tor: ADD_ONION response didn't have an OK status: Onion address collision' when trying to schedule a share. We need to unregister the onion here so we can re-use it later for a scheduled share

mig5:
One more bug:

When opening OnionShare with a persistent tab (a persistent v3 onion), if I check 'Use a legacy address' in the advanced settings, it will continue to use the persistent v3 address.

I guess clicking 'Use a legacy address' should remove a persistent v3 onion. But maybe it should throw a modal warning first in case the user changes their mind, before the private key is removed from settings?

onionshare_gui/main_window.py
```diff
@@ -0,0 +1,292 @@
...
+                def update_available(update_url, installed_version, latest_version):
+                    Alert(
+                        self.common,
+                        strings._("update_available").format(
+                            update_url, installed_version, latest_version
+                        ),
+                    )
+
+                self.update_thread = UpdateThread(
+                    self.common, self.common.gui.onion, self.common.gui.config
```
mig5: `self.common.gui.config` is an unknown attribute and crashes Mac (probably Windows too)

onionshare/__init__.py
micahflee: Actually the trailing comma is required! It's because args must be a tuple, and if I just pass in `(app.port)` it evaluates to `app.port`, which ends up passing in an integer instead. :)

onionshare/mode_settings.py
micahflee: Good catch. I called this `TabSettings` at first, but renamed it to `ModeSettings` because it made more sense, especially since CLI doens't have tabs but still uses it.

onionshare/web/web.py
micahflee: Good catch. I changed things to use `ModeSettings`, so I don't need that comment at all anymore

micahflee:
I hit this error again and I just figured out why:

Screenshot from 2019-11-28 15-47-59

It turns out, onionshare is launching its own `tor` subprocess, but that subprocess immediately fails:

```
Nov 28 19:52:41.000 [warn] It looks like another Tor process is running with the same data directory.  Waiting 5 seconds to see if it goes away.
Nov 28 19:52:46.000 [err] No, it's still there.  Exiting.
Nov 28 19:52:46.000 [err] set_options(): Bug: Acting on config options left us in a broken state. Dying. (on Tor 0.4.1.6 )
Nov 28 19:52:46.000 [err] Reading config failed--see warnings above.
```

And, sure enough, when I look at my process list there's a an extra `tor` process running with the same data dir, from a previous run on onionshare.

So it looks like actually this is actually the bug that proactivesvcs reported in keybase the oter day:

> I noticed that onionshare leaves a child tor.exe behind on Windows, after exiting.

Apparently, quitting onionshare doesn't always clean everything up.

micahflee:
In regards to the `There was an error with Tor: ADD_ONION response didn't have an OK status: Onion address collision` bug, I think this patch should fix that issue:

```
diff --git a/onionshare_gui/threads.py b/onionshare_gui/threads.py
index 95b6dab..bfa4f72 100644
--- a/onionshare_gui/threads.py
+++ b/onionshare_gui/threads.py
@@ -76,7 +76,7 @@ class OnionThread(QtCore.QThread):
                 time.sleep(0.2)
                 self.success_early.emit()
                 # Unregister the onion so we can use it in the next OnionThread
-                self.mode.app.start_onion_service(self.mode.settings)
+                self.mode.app.stop_onion_service(self.mode.settings)
             else:
                 self.mode.app.start_onion_service(
                     self.mode.settings, await_publication=True
```

But do you know how to reliably reproduce it? I've seen it before by mistake, but I'm not sure how to trigger it, to test if this fixes it. (I mean it really should though)

mig5:
> But do you know how to reliably reproduce it? I've seen it before by mistake, but I'm not sure how to trigger it, to test if this fixes it. (I mean it really should though)

Hmm, you're right, I don't actually always reproduce it.

But I've found yet another problem - the address that is shown whilst scheduled, is not the onion address that ends up used after the schedule timer runs out. Even with your fix...

See this output:

```
[Nov 29 2019 03:17:12 PM] Web.generate_password: persistent_password=None
[Nov 29 2019 03:17:12 PM] Web.generate_password: built random password: "panther-chill"
[Nov 29 2019 03:17:12 PM] OnionShare.start_onion_service
[Nov 29 2019 03:17:12 PM] Onion.start_onion_service: port=17636
[Nov 29 2019 03:17:12 PM] Onion.start_onion_service: key_type=NEW, key_content=ED25519-V3
[Nov 29 2019 03:17:12 PM] ModeSettings.set: updating catfish-crabgrass-eastward: share.filenames = ['/home/user/git/onionshare/BUILD.md']
[Nov 29 2019 03:17:12 PM] ModeSettings.set: updating catfish-crabgrass-eastward: general.service_id = jyaiywb3mtt5jt3ridjgrqqj6bbxjltkvdldaqryyk7sxh4kvpzykuid
[Nov 29 2019 03:17:12 PM] Onion.stop_onion_service: onion host: jyaiywb3mtt5jt3ridjgrqqj6bbxjltkvdldaqryyk7sxh4kvpzykuid
[Nov 29 2019 03:17:14 PM] Tab.copy_url
[Nov 29 2019 03:17:59 PM] Settings.load
[Nov 29 2019 03:17:59 PM] Settings.load: Trying to load /home/user/.config/onionshare/onionshare.json
[Nov 29 2019 03:17:59 PM] Mode.start_server: Starting a scheduled onion thread
[Nov 29 2019 03:17:59 PM] OnionThread.__init__
[Nov 29 2019 03:17:59 PM] OnionThread.run
[Nov 29 2019 03:17:59 PM] Web.generate_static_url_path: new static_url_path is /static_7i3qwilohv3oam3lymnvwnp6ja
[Nov 29 2019 03:17:59 PM] OnionShare.start_onion_service
[Nov 29 2019 03:17:59 PM] Onion.start_onion_service: port=17636
[Nov 29 2019 03:17:59 PM] Onion.start_onion_service: key_type=NEW, key_content=ED25519-V3
```

The first address jyaiywb3mtt5jt3ridjgrqqj6bbxjltkvdldaqryyk7sxh4kvpzykuid is the one that is generated 'in advance', and then we stop that service immediately (with your patch).

But then when the timer runs out, we generate a new onion with key_type=NEW. That is incorrect, we should have generated one with the private key retained from the earlier quick start/stop (it's a bit like when using persistence).

So even with your patch there is still a bug somewhere. Will look into it..

mig5:
OK, it looks like we've accidentally lost some of the logic in onionshare/onion.py wherever there were conditions regarding `if not save_scheduled_key'. That attribute is still passed to `start_onion_service`, but it is never evaluated anywhere.

So your patch is needed, yes - but we also need to reinstate the logic that does not generate a *new* onion private key if we were scheduling an onion with a pre-existing private key.

mig5:
Actually it was this snippet that got lost:

```
        elif self.scheduled_key:
            key_content = self.scheduled_key
```

But there were similar settings related to `save_scheduled_key` and Client Auth (on v2 onion).

Maybe we should just merge this and can fix that up later separately... what do you think?

mig5:
I've actually already fixed it now locally on this branch, but I need to re-instate the similar fix for Client Auth too.

micahflee:
Oh wow thanks, I would definitely appreciate your commits in this branch to fix the auto-start/auto-stop timer issues. You originally wrote that code and I probably didn't full understand it when I was refactoring -- I wonder why the tests are passing though?

Also I just pushed a commit that helps with with orphan `tor` subprocess issue, I think. I think I do a slightly better job at making sure the `tor` subprocess is actually killed on cleanup, and it appears to work fine. But what I also discovered is that if onionshare crashes, and doesn't close cleanly, then the cleanup is never run, and there's an orphan `tor`.

So this means in most cases quitting onionshare and starting it again should be fine, but if it crashes and you start it again, it will hit this error. Which isn't the best. One option is before connecting to tor we can try to see if there's an orphan tor running already and kill it if there is... but I'm not sure if that's the most elegant. Thoughts?

mig5:
> I wonder why the tests are passing though?

Because the onion host is always 127.0.0.1 (no Tor tests - ha ha we are learning our lesson..)

> One option is before connecting to tor we can try to see if there's an orphan tor running already and kill it if there is... but I'm not sure if that's the most elegant. Thoughts?

I'm worried if there's a chance the user running OnionShare might be running a tor process separately for a totally different reason. We wouldn't want to kill that if it's not 'ours'.

Perhaps we need to create a 'lock' file in our tor state dir that contains the pid of the tor process, and remove that file on each 'clean' closure of the app. And check for the lock file on startup - if it's there, we should destroy the pid that that file contains (and then delete that lock file before starting up proper). Is that safer?

I'm still working on quirky bugs with scheduled onions but getting there. When it's ready I'll push it in.

micahflee:
> I'm worried if there's a chance the user running OnionShare might be running a tor process separately for a totally different reason. We wouldn't want to kill that if it's not 'ours'.

I think we can detect the exact one, actually. The process will include the filename of the `torrc` file, like for example I ran `onionshare-gui`, and in another terminal:

```
$ ps aux |grep ".config/onionshare/tor_data/torrc"
user       31641  1.9  0.2  36148 31816 pts/1    S+   21:13   0:00 /usr/bin/tor -f /var/home/user/.config/onionshare/tor_data/torrc
user       32638  0.0  0.0   5100   888 pts/6    S+   21:13   0:00 grep --color=auto .config/onionshare/tor_data/torrc
```

mig5:
OK, I've pushed up the fix for scheduled shares. It's a bit awkward because we both were editing that `cleanup` function, hehe. I've retained your changes while re-instating some bits. We actually need to retain that `stop_tor` attribute because we need a way to wipe *all* ephemeral onions without killing tor. The `stop_onion_service` function only removes the one service id that it knows about. Scheduled shares are a bit different somehow (can't exactly work out why, I think we aren't 'setting' the service_id in the Settings object before reusing it or something..)  we need to be more brutal about removing all onions so that we don't risk accidentally re-using them (which leads to collisions) .. 

There might still be bugs or more elegant ways of handling this, but this at least reinstates the original and working behaviour around scheduling, including for Client Auth on v2.

micahflee:
> I would actually advocate for merging the cleanup and stop_onion_service function in the Onion class, so that stop_onion_service always iterates over all known service_ids and destroys them.. I can't think of a case where we would want to keep any service_ids (onions) in the connected Onion session when 'stopping'..

Now that we have tabs, all tabs share the same Onion object, and the same tor subprocess. So if you have five tabs with running servers, and you click stop and one of them, if we destroy all service_ids, we end up killing the onion service for all five tabs.

mig5:
oohhh, crap.. I didn't realise the ramifications. Feel free to revert my commit(s)... will have to find some other way..

mig5:
Think I fixed it! We weren't updating the service_id in the 'general' settings if it was already present (with the previous service id from a scheduled share) which was why there was a mismatch/collision if you stopped a scheduled share and then started a *new* scheduled share.

So I've reinstated your original patch for stopping the share, but now we don't use the cleanup() function that obliterates all service ids.

micahflee:
About [this comment](https://github.com/micahflee/onionshare/pull/1075#issuecomment-559628625), where if you use a persistent v3 address, you can't then go and make it a v2 address, I think a good solution would be to not let you change the legacy/client auth settings in the future after setting them once, in the same tab. I like the idea of each tab having its own onion address (even if it's not persistent), and if you want a new one you make a new tab. So I think once there's data in `mode_settings["persistent"]["private_key"]`, then you can't change settings for that tab.

About the auto-start timer, there's still an issue. Because the `Onion` object is shared across all tabs, we can't store any tab-specific state in it, or tabs might clobber each other. The patches you added did indeed fix the problem by using `self.scheduled_key` for a single tab with an auto-stop timer. But what if we have two tabs open and they both have auto-stop timers set? The onion addresses that are displayed at the beginning aren't same as the once that are displayed at the end. Instead of using `self.schedule_key`, we need to store this scheduled key inside the `mode_settings`, which is where tab-specific settings live -- I was actually trying to store it in `mode_settings["persistent"]["private_key"]`, since that's where we'll need to store it anyway, but I apparently did it in a buggy way.

So here's my to do list for this PR, I think:

- [ ] Make it so after you start a server once, you can't change the legacy/client auth settings
- [ ] Fix the auto-start timer issue by removing all tab-specific state from the Onion object
- [ ] Before starting bundled Tor, detect if an existing bundled Tor process is currently running (the exact one, with the path to the same torrc file in the command), and if so try killing it first

micahflee:
- [ ] When you start a server with the auto-stop timer checked, then stop it, then uncheck auto-start timer, and start it again, it acts like auto-start timer is still checked
- [ ] In some situations you still get "There was an error with Tor: ADD_ONION response didn't have an OK status: Onion address collision" -- need to figure out exactly how to reproduce it

micahflee:
I started a PR to fix the tests (#1099). I finally got it to actually run the tests properly, but a few tests were failing now. However, this branch completely refactors the tests, and some of the failing ones were obsolete, so I decided to merge the tests branch into this PR.

Now all of the tests pass, but the pytest process itself closes with a segfault.

micahflee:
There's still one segfault. I managed to fix the others by successfully closing threads, so maybe there's a thread I'm missing. Almost there.

Then, I'm thinking of just opening issues for the remaining tasks in this PR, and getting this PR merged into develop sooner.

mig5:
@micahflee I started a share with these settings checked:

[x] Stop sharing after files have been sent...
[x] Save this tab, and automatically open it when I open Onionshare
[x] Start onion service at scheduled time

Then I stopped the running share, unchecked the auto-start option, and tried to start, but I reliably get the `There was an error with Tor: ADD_ONION response didn't have an OK status: Onion address collision` error every time since - even if I uncheck all other options.

I seem to only be able to resolve if I close OnionShare altogether after removing persistence. In other words, if I had persistence enabled, and I start OnionShare and use auto-start, then I reliably reproduce the collision after each stop/start. 

*Without* auto-start, but with or without persistence enabled, I don't seem to reproduce it.

I think we still aren't clearing/unregistering the onion service ID via the Stem controller based on the autostart (maybe autostop, didn't test) setting.

That said, given the tests are passing, and the size of this PR, and that this bug might not even be related to Tabs directly but to refactoring of the Onion object or threads(?) I'm not against merging it in and tackling such issues in smaller changesets... would be good to get back to smaller changes. Up to you.

SaptakS:
I am planning to review this PR in general by EOD. But in general I agree with @mig5 that it would be better to merge this PR if the tabs features are mostly working and dealing with other issues in smaller separate issues and PR.

SaptakS:
So apart from what @mig5 mentioned more or less seems to be working for me. There is another small UI thing that I noticed which might not be something new introduced in this PR and might have been carry forwarded (though since the UI has changed a little, difficult to verify).

So I made a `Receive Mode` tab persistent. Then closed it. Then started the GUI again.
So when I start, at the very beginning for a split second, the entire `Receive Mode` is shown like this:
Screenshot from 2020-04-04 17-34-04

But after that split second, only half of the UI vanishes but other half remains:
Screenshot from 2020-04-04 17-33-55
Now either this is a bug, or this is a feature to prevent users from accessing the UI. But at the same time, given that we might revamp the UI, I feel this might not be super important, especially since it doesn't affect the user workflow at all.

I haven't found any huge blocker bug as such.
cc: @micahflee

micahflee:
Thank you @mig5 and @SaptakS! I just added your comments to #1105 ... and with that, I'm going to merge this PR!

PR #1086, opened 2020-01-04
Update org.onionshare.OnionShare.desktop

milotype:
Added croatian (hr) translation for Comment and Keywords


milotype:
Just to inform you – I've also translated OnionShare into croatian at Weblate. So, whenever you feel like it, you can implement that as well.

PR #1096, opened 2020-03-09
Add chat feature to onionshare

SaptakS:


holantonela:
🔥

micahflee:
FYI, the `python3-flask-socketio` package is available in debian bullseye: https://packages.debian.org/bullseye/python3-flask-socketio

And ubuntu eoan (19.10): https://packages.ubuntu.com/eoan/python3-flask-socketio

It doesn't look like it's packaged in fedora yet. But I'm hoping we'll just have a flatpak soon.

SaptakS:
@micahflee Oh! That's great news!

SaptakS:
It's a very basic UI right now. Have been trying to focus mostly on the UX (which might also need some help)

image


Right now for the user list, I am using a background thread for every user session which broadcasts the username. That way, every user can create their own user list in the frontend. I am using this approach to avoid storing any kind of data. An alternative would be to maybe have a sqlite db storing all the usernames. I think for a small number of users, this background thread process works fine. But let me know if anyone feels that would cause some kind of performance problem.

Also, for the time, the randomly generated username uses the same code as the password generator. I am open to ideas for creating better, interesting random names. 

Also, curious if adding colors for every user makes sense like in etherpad. Or a dedicated color for incoming and outgoing message. Or should we stick to the IRC style UI that I build for the first iteration.

SaptakS:
One thing that I realized I need to do is stop the background thread and leave some kind of message when someone closes the tab and disconnects from the socket.

Another suggestion by @micahflee  that I am considering implementing is browser notifications when someone joins.

SaptakS:
Something else I noticed is when using over onion network, the socket tends to disconnect sometimes if the internet is not too good. I think there are some ways of keeping the socket alive longer. But I am open to suggestions of how to do that and any UX/UI implementations to make things more meaningful.

I will stop blabbering now and let everyone have a look. :sweat_smile:

kushaldas:
@SaptakS please update the requirements files too :)

emmapeel2:
about UX, when trying to scroll up to see last messages, even when I was still holding the slider on the side, the window will continuously scroll down to  the bottom.
That made it very hard to read previous messages.

ZeroDot1:
@SaptakS That's a really good idea. I like it.

SaptakS:
> @SaptakS please update the requirements files too :)

@kushaldas Yes. I will update that. As discussed with @micahflee I am still not sure which route we will go down for the requirements. So once I make the final PR, I will definitely add that.

SaptakS:
> even when I was still holding the slider on the side, the window will continuously scroll down to the bottom.

I actually made the feature such that if a new message (or status message) comes in, it scrolls to the bottoms of the page. Never realized it will have a side effect. I will definitely look into it, such that if someone is scrolling on top, new messages or anything don't effect that. Thanks for pointing out @emmapeel2

4jNsY6fCVqZv:
@SaptakS Since the initial description of this PR does not provide any information about it, I want to ask why Onionshare needs an integrated chat at all? There are already so many good decentralised p2p communication solutions. And I imagine that chat is also a huge component that needs to be maintained in the future. I would also be interested to know which security concept your solution offers.

share/static/js/chat.js
```diff
@@ -1,65 +1,123 @@
...
+    }
+  });
+  return userListHTML;
+}
+
+var getScrollDiffBefore = function() {
+  return $('#chat').scrollTop() - ($('#chat')[0].scrollHeight - $('#chat')[0].offsetHeight);
+}
+
+var scrollBottomMaybe = function(scrollDiff) {
```
SaptakS, line 92: @emmapeel2 this should address the issue that you raised UX wise. Right now the UX logic for scrolling to bottom is something like this:

- If the current user is sending a message, even if they have scrolled up, on sending a message, the chat scrolls to bottom.
- If the current user's chat window is already scrolled to the bottom, any incoming status or chat message will still scroll to the bottom.
- If the current user has scrolled up in the chat window, any incoming status or chat message will not change their scroll position.

I think that's the common norm in chat? Let me know if that sounds good.

onionshare/web/chat_mode.py
```diff
@@ -80,40 +91,33 @@ def update_username(message):
...
+                    "old_name": current_name,
+                    "new_name": session.get("name"),
                 },
-                room=session.get("room")
+                room=session.get("room"),
             )
 
-
-
-class UserListWorker(object):
```
SaptakS: Removed the background worker logic by an instance variable logic. Given that we can assume that onionshare chat will be running on a single server, and not clustered over different servers, the instance variable will solve the purpose of holding an user list (Point me if I am wrong). This would reduce the load because no background thread.

share/static/js/chat.js
```diff
@@ -1,65 +1,123 @@
...
+
+    // Update username
     $('#update-username').on('click', function() {
       var username = $('#username').val();
+      current_username = username;
       socket.emit('update_username', {username: username});
     });
+
+    // Show warning of losing data
+    $(window).on('beforeunload', function (e) {
```
SaptakS, line 45: I am not 100% sure this is a good idea. Bit of conflicted here. On one hand, I feel it's important to give users a warning that closing or refreshing will lead them to lose all messages. On the other hand, if someone refreshed too many times (don't know why someone would do that), or every time someone closes the tab, this warning will be shown. So that might get frustrating. @holantonela @emmapeel2 would really love some feedback from you on this.

SaptakS:
> @SaptakS please update the requirements files too :)

Added the dependencies to poetry. :)

onionshare/web/chat_mode.py
```diff
@@ -0,0 +1,123 @@
...
+
+        @self.web.app.route("/")
+        def index():
+            history_id = self.cur_history_id
+            self.cur_history_id += 1
+            session["name"] = (
+                session.get("name")
+                if session.get("name")
+                else self.common.build_username()
+            )
```
SaptakS, line 46: This logic is a little broken and needs to be fixed. So `flask-socketIO` seems to create a fork of the actual flask session and when modified, it doesn't modify the actual session but the forked session.

>If a SocketIO handler modifies the session, the modified session will be preserved for future SocketIO handlers, but regular HTTP route handlers will not see these changes. Effectively, when a SocketIO handler modifies the session, a “fork” of the session is created exclusively for these handlers. The technical reason for this limitation is that to save the user session a cookie needs to be sent to the client, and that requires HTTP request and response, which do not exist in a SocketIO connection.

Hence if a user updates their name in chat, on refresh, their old name is shown, instead of the new name, and also causes the user list to break a bit.

There are two ways I can go about it.
1. Maintain a key: value dictionary of originally set name in the flask session as key and the updated username and any future data (like if we decide to add colors to each users) related to that user in the value. On refresh, we just reassign all those values from the dict based on the originally set name. This is a bit tricky.
2. Send an ajax request instead of a socket emit when updating the username, so that there is a HTTP Request that can modify the actual flask session variables.

I am leaning towards 2 right now.

4jNsY6fCVqZv:
@SaptakS Could you imagine giving me some feedback on my unresolved question? https://github.com/micahflee/onionshare/pull/1096#issuecomment-602017015

kushaldas:
> @SaptakS Could you imagine giving me some feedback on my unresolved question? [#1096 (comment)](https://github.com/micahflee/onionshare/pull/1096#issuecomment-602017015)

Because a lot of users (like me) wants to have a chatting option and we already have Onionshare installed on the box.

SaptakS:
> @SaptakS Since the initial description of this PR does not provide any information about it, I want to ask why Onionshare needs an integrated chat at all? There are already so many good decentralised p2p communication solutions. And I imagine that chat is also a huge component that needs to be maintained in the future.

The idea of a chat in Onionshare is to provide the onionshare users to have a temporary, onionshare-y way of chatting inside the onion network. It's a nice to have feature, and it's not going to be a huge signal-like chat application.

>  I would also be interested to know which security concept your solution offers.

Firstly, it's not going to be a persistent chat. Secondly, the entire chat functions inside the onion network. The chat server runs in the person's computer (just like in all other onionshare feature) and uses the same method as rest of onionshare uses. 

I will let @micahflee jump in to address any other question you may have. :)

matt1tk:
> 
> 
> @SaptakS Since the initial description of this PR does not provide any information about it, I want to ask why Onionshare needs an integrated chat at all? There are already so many good decentralised p2p communication solutions. And I imagine that chat is also a huge component that needs to be maintained in the future. I would also be interested to know which security concept your solution offers.

Easy integration to quickly send files and chat with someone in the same app. Few decentralized p2p chat programs use tor, only other one I can think of, off the top of my head, is ricochet.im which does not include file sending yet.

matt1tk:
> 
> 
> > @SaptakS Since the initial description of this PR does not provide any information about it, I want to ask why Onionshare needs an integrated chat at all? There are already so many good decentralised p2p communication solutions. And I imagine that chat is also a huge component that needs to be maintained in the future.
> 
> The idea of a chat in Onionshare is to provide the onionshare users to have a temporary, onionshare-y way of chatting inside the onion network. It's a nice to have feature, and it's not going to be a huge signal-like chat application.
> 
> > I would also be interested to know which security concept your solution offers.
> 
> Firstly, it's not going to be a persistent chat. Secondly, the entire chat functions inside the onion network. The chat server runs in the person's computer (just like in all other onionshare feature) and uses the same method as rest of onionshare uses.
> 
> I will let @micahflee jump in to address any other question you may have. :)

Does this provide plausible deniability as well, a la off-the-record? As well as some sort of HMAC?

micahflee:
> Does this provide plausible deniability as well, a la off-the-record? As well as some sort of HMAC?

Yes, they will be plausibly deniable.

- Because it's an onion service, there is no way for an attacker to eavesdrop on the messages. The closest they can get is if they run a malicious Tor rendezvous node that's chosen for the onion service, they'll be able to spy on encrypted onion traffic. So, there's no capturing ciphertext to decrypt later on.
- Messages never get written to disk at all (right @SaptakS?), so there will be no record of the conversation if a computer is searched either. It's entirely ephemeral.

No, there won't be any sort of signatures. Basically, as long as you securely distribute the OnionShare URL to others, they can join the chat and beyond that there's no user authentication. So, like the rest of OnionShare, you have to make sure you're securely sharing URLs.

micahflee:
@SaptakS I just merged a bunch of PRs into develop, and now something is broken in `ChatMode` when I load the URL in Tor Browser. You can merge `develop` into this branch to reproduce.

```
  File "/home/user/code/onionshare/onionshare_gui/tab/tab.py", line 477, in timer_callback
    mode.handle_request_individual_file_started(event)
  File "/home/user/code/onionshare/onionshare_gui/tab/mode/__init__.py", line 478, in handle_request_individual_file_started
    self.toggle_history.update_indicator(True)
AttributeError: 'ChatMode' object has no attribute 'toggle_history'
Aborted (core dumped)
```

(Also when you merge in `develop` there's an annoying merge conflict with `poetry.lock` -- the easiest solution IMO is to just delete `poetry.lock` and re-run `poetry install` to create a new one.)

micahflee:
Another issue with merging in develop. jQuery has been updated to 3.5.1, so the path to load it in `chat.html` needs to be updated as well.

micahflee:
There's no `<noscript>` tag. When the security slider is set to Safest and javascript is disabled, nothing works. I think instead, there should the OnionShare logo and a brief sentence like: "Chat requires JavaScript, so you must set your Tor Browser security level to Safer or Standard to join."

SaptakS:
> Another issue with merging in develop. jQuery has been updated to 3.5.1, so the path to load it in `chat.html` needs to be updated as well.

I am aware of this change. Have to rebase and fix the path.

> There's no <noscript> tag. When the security slider is set to Safest and javascript is disabled, nothing works. I think instead, there should the OnionShare logo and a brief sentence like: "Chat requires JavaScript, so you must set your Tor Browser security level to Safer or Standard to join."

My mistake I forgot the `<noscript>` tag. Will add it

I will rebase and fix the conflicts and errors.

micahflee:
I decided to take a stab at some of this myself. I see that allowing the user to set their name in a modal dialog is more complicated than I thought because the name is initially set on the server side. So instead I just did a few smaller UX changes, and opened issues for the future: #1150 and #1151 

I think this is good enough to merge!

SaptakS:
@micahflee Sounds a good way to go. Sorry I got a little busy lately. I will look into the above issues very sooooon!

Issue #1079, opened 2019-12-12
onionshare doesn't work with Python 3.8 (dictionary keys changed during iteration)


myf:
this issue presents to anyone who's currently using archlinux which comes with `python 3.8`. Launching onionshare in the commandline will fail with the following message:
```
$~ onionshare some_file
OnionShare 2.2 | https://onionshare.org/
OnionShare could not connect to Tor in the background:
dictionary keys changed during iteration
```
After further investigation, this bug is related to the python package `stem` which interfaces with tor. `stem-1.7.1` is not compatible with python3.8.
The discussion of this issue is documented under this ticket https://trac.torproject.org/projects/tor/ticket/32552 . To build a working version, you'll have to build `stem` from master instead `stem1.7.1`:
```
% git clone https://git.torproject.org/stem.git
% cd stem
% sudo python setup.py install
# then go back to onionshare
% sudo pip install --upgrade stem
```
We may have to wait for the new release of stem for the fix to be patched via pypi. But before that, this is a work around if you are using archlinux and/or python 3.8

micahflee:
Stem 1.8 is out https://stem.torproject.org/change_log.html#version-1-8

This may fix the problem...

micahflee:
It worked! I'm using fedora 32 with python 3.8 and stem 1.8.0, and it stem successfully works.

mig5:
Fixed in #1098

Issue #1094, opened 2020-03-08
Switch from requirements.txt to poetry


micahflee:
[Poetry](https://github.com/python-poetry/poetry) is a good way of keeping track of cross-platform python package management.

While switching to poetry, also update all python dependencies.

SaptakS:
I am taking a  stab at this.

mig5:
I feel like we can close this since we are definitely using poetry now in develop branch.

PR #1098, opened 2020-03-17
Switches from pipenv and requirements.txt to poetry to manage dependencies

SaptakS:
Refs #1094 

This needs cross-platform testing. I have tested it in linux and seems to be working great. I am on my way to test with windows. @micahflee it would be great if you can test in mac OSx.

I have also for the time kept a fixed version in the dependency instead of using `^` in the version because that's how it was in requirements.txt and I didn't want to bring any breaking changes by allowing updates. But if it is safe to do `^` then I will add that in `pyproject.toml`


pyproject.toml
```diff
@@ -0,0 +1,50 @@
+[tool.poetry]
+name = "onionshare"
+version = "2.2"
```
micahflee: Can you add a note to the release documentation (https://github.com/micahflee/onionshare/blob/develop/BUILD.md#changelog-version-and-signed-git-tag) to update the version in this file with each release?

pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+description = "OnionShare lets you securely and anonymously send and receive files. It works by starting a web server, making it accessible as a Tor onion service, and generating an unguessable web address so others can download files from you, or upload files to you. It does _not_ require setting up a separate server or using a third party file-sharing service."
+authors = ["Micah Lee <micah.lee@theintercept.com>"]
+license = "GPLv3+"
+
+[tool.poetry.dependencies]
+python = "^3.7"
```
micahflee: Actually sadly there's a bug where onionshare doesn't work with python 3.8 (#1079) though I'd love to fix that. So this should just be "3.7"

micahflee: Or, however you do >= 3.7 and < 3.8

pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+atomicwrites = "1.3.0"
+attrs = "19.1.0"
+more-itertools = "7.2.0"
+pluggy = "0.13.0"
+py = "1.8.0"
+pytest = "5.1.2"
+pytest-faulthandler = "2.0.1"
+pytest-qt = "3.2.2"
+six = "1.12.0"
+urllib3 = "1.25.3"
```
micahflee: None of these are tied to any specific version, they can all have `^` -- and while you're at it, I would be fine bumping all of the versions to their newest versions, so long as the tests pass

pyproject.toml
```diff
@@ -0,0 +1,50 @@
+[tool.poetry]
+name = "onionshare"
+version = "2.2"
+description = "OnionShare lets you securely and anonymously send and receive files. It works by starting a web server, making it accessible as a Tor onion service, and generating an unguessable web address so others can download files from you, or upload files to you. It does _not_ require setting up a separate server or using a third party file-sharing service."
+authors = ["Micah Lee <micah.lee@theintercept.com>"]
```
micahflee: Can you change my email here to micah@micahflee.com? :)

pyproject.toml
micahflee: You can update to all the latest versions with `poetry update`

pyproject.toml
micahflee: Actually, I'm hoping that upgrading to stem 1.8.0 will allow onionshare will work with python 3.8 as well. If that's the case, then this PR will also fix #1079

SaptakS: Can do that in this PR itself.

pyproject.toml
micahflee: Actually, while troubleshooting in #1079 I discovered that if you set the version for all of the dependencies to `*`, and then run `poetry install` and `poetry update`, it updates all dependencies in `poetry.lock`, and fixes the python 3.8/stem problem.

pyproject.toml
SaptakS: Done

pyproject.toml
SaptakS: Changed everything to `*` and then did a `poetry update` to update the `poetry.lock`

SaptakS:
@micahflee I haven't done the linux build part yet. Do you mean we should use poetry instead of apt and dnf for python dependencies in `BUILD.md`? I am not super clear on what you mean. :sweat_smile:

PR #1099, opened 2020-03-22
Fix tests to install latest dependendencies instead of using apt packages

micahflee:
This testing with python 3.6, 3.7, and 3.8, instead of just 3.6 and 3.7. It no longer installs onionshare dependencies from apt packages.

Instead, it installs Qt 5.14.0 binaries downloaded from Qt's website. It uses a [noninteractive controller script](https://doc.qt.io/qtinstallerframework/noninteractive.html) to control the installer without a GUI.

Then, it installs the rest of the dependencies using poetry.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ workflows:
 jobs:
   test-3.6: &test-template
     docker:
-      - image: circleci/python:3.6.6
+      - image: circleci/python:3.6-buster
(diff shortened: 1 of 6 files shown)


micahflee:
I'm going to fix the tests directly in #1075, because they've been refactored so much. (And try to get #1075 merged into develop soon because it's such a monster PR)

PR #1106, opened 2020-04-06
auto-start/stop fixes

mig5:
Fixes logic that checks if auto-start timer setting is enabled, so that it checks real settings and not widget instantiation. 

This was the issue that resulted in shares still being 'scheduled' even after the auto-start setting had been unchecked. This was also causing the knock-on effect of 'ADD_ONION' collision errors (because OnionShare was trying to schedule a share that it shouldn't, resulting in a duplicate attempt to register the service id)

Also, I noticed that if the timers (both autostart and autostop) were enabled in settings, and settings were persistent, then on startup of OnionShare, the setting would be checked, but the dateTime() widget was not shown. The setting would have to be disabled and then re-enabled to activate the dateTime widget. This was just because we explicitly `hide()` the widget on startup, whereas in fact we should check if the setting was enabled first.

### DIFF
--- a/onionshare_gui/tab/server_status.py
+++ b/onionshare_gui/tab/server_status.py
@@ -269,7 +269,7 @@ class ServerStatus(QtWidgets.QWidget):
                     self.common.gui.css["server_status_button_working"]
                 )
                 self.server_button.setEnabled(True)
-                if self.autostart_timer_datetime:
+                if self.settings.get("general", "autostart_timer"):
                     self.server_button.setToolTip(
                         strings._("gui_start_server_autostart_timer_tooltip").format(
                             self.mode_settings_widget.autostart_timer_widget.dateTime().toString(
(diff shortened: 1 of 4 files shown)


onionshare_gui/tab/server_status.py
```diff
@@ -279,15 +279,6 @@ def update(self):
                     )
                 else:
                     self.server_button.setText(strings._("gui_please_wait"))
-
-                if self.settings.get("general", "autostart_timer"):
```
mig5, line 283: This might look weird in a diff but in fact it was a duplicate chunk of code here that isn't needed anymore due to the change in the conditional above.

mig5:
Tests are failing during QT Binary install process, at:

```
[1954] At page: CredentialsPage ('Qt Account  Your unified login to everything Qt')
```

Seems to be blocked waiting for something.. wanting to login to Qt somehow?

SaptakS:
@mig5 I remember that the Qt binary installation indeed needs a login even in the headless mode. I think @micahflee added the credentials in Circle CI variables.

SaptakS:
I am wondering if it is because of the `QT_EMAIL` and `QT_PASSWORD` being set by @micahflee in his own CircleCI config but since this CI is currently running on your forked repo's branch rather than in the upstream repo, it can't find the variables in your Circle CI. Maybe if you can check setting `QT_EMAIL` and `QT_PASSWORD` in your own Circle CI and if it passes, that will kind of prove my theory. :sweat_smile:

mig5:
Thanks @SaptakS sounds like that's what it is! I'll check it out...

mig5:
Thanks @SaptakS all fixed!

SaptakS:
Apart from the temporary fix, maybe it is worthwhile to modify the Circle CI settings to run CI at a PR level rather than a forked branch level maybe? I am not 100% sure how but I think it can be done. Else it would need every contributor with a forked repo to technically add `QT_EMAIL` and `QT_PASSWORD` for the tests to pass. Though not an immediate requirement but might be good for the long run. cc: @micahflee @mig5

PR #1113, opened 2020-05-12
update archlinux onionshare URL

Kcchouette:


kpcyrd:
I think this should be changed to suggest `pacman -S onionshare` instead, the PKGBUILD is only relevant if you're building the package yourself :)

mig5:
@kpcyrd this is for the `BUILD.md` which really is specifically about 'building the package yourself' e.g for developers.

The README.md has a link to the wiki page about installing, in which ArchLinux is included in the 'available via package manager' table https://github.com/micahflee/onionshare/wiki/How-Do-I-Install-Onionshare#other-linux

PR #1120, opened 2020-05-27
Add a Delete All button in the File List area for share modes

### LINKED ISSUE #1116: Ability to remove multiple items after sharing files
After clicking on `Stop sharing`, the list of files that I was sharing earlier continues to appear. Removing them using the `x` mark next to each file name works, but it would be nice to be able to remove all files at once.

mig5:
Fixes #1116

### DIFF
--- a/share/locale/en.json
+++ b/share/locale/en.json
@@ -12,6 +12,7 @@
     "gui_add_files": "Add Files",
     "gui_add_folder": "Add Folder",
     "gui_delete": "Delete",
+    "gui_file_selection_clear_all": "Clear All",
     "gui_choose_items": "Choose",
     "gui_share_start_server": "Start sharing",
     "gui_share_stop_server": "Stop sharing",
--- a/share/locale/en.json
+++ b/share/locale/en.json
@@ -12,7 +12,7 @@
     "gui_add_files": "Add Files",
     "gui_add_folder": "Add Folder",
     "gui_delete": "Delete",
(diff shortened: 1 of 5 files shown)


share/locale/en.json
```diff
@@ -12,6 +12,7 @@
     "gui_add_files": "Add Files",
     "gui_add_folder": "Add Folder",
     "gui_delete": "Delete",
+    "gui_file_selection_clear_all": "Clear All",
```
SaptakS, line 15: Given that we use the word "Delete" to remove each file from the UI, should we use "Delete All" instead of "Clear All" to make it clear that both have the same effect?

mig5: Yeah, this was tricky, because we use 'Clear All' in the toggled History view, for history items. I originally had 'Delete' and then I thought it might be better to make it consistent with the History 'Clear All' button.

But, I think I agree with you: from a UX perspective, 'Delete' is more consistent with the 'Delete' button for individual files, and users probably associate 'Delete' with files and 'Clear' with 'history' (like 'clearing' browser history). So I'm changing it back :) Thanks for the review.

SaptakS: Honestly I feel "Remove" makes the intent clear the best. Because it's clear you are removing the files from share list. Delete sometimes might feel like delete from where. But given that people are already familiar with what Delete means in respect to onionshare, I think this is fine.

PR #1122, opened 2020-05-28
#1095 Fix crash for shares larger than 2GB. #1089 add ETA/bytes uploaded in progress bar of ReceiveMode items

### LINKED ISSUE #1089: Receive Mode does not show ETA/upload size in progress bars
When sending/receiving large files, or if the connection is particularly bad, it would be nice to have an estimation of transfer speed and the remaining time to transfer the data. This is useful to evaluate how the connection perform, and if the users need to put some countermeasures in place (ie, if the download will take too much time anyway to complete in a reasonable time, exchaging receiver and sender may help if upload and download bandwidths of the internet provider are asymmetrical).

Since the app is in Python, you may be interested in taking a look at [tqdm](https://github.com/tqdm/tqdm), which is a library that is focused on providing easy but precise progress bars. Disclaimer: I was involved with tqdm in the past. (but I'm not advising tqdm to everyone, you can look at my github history, it's just that here I think it could be a good addition in terms of functionality to get estimated download times and with low effort by reusing a library).

### LINKED ISSUE #1095: Onionshare shuts down unexpectedly on initiation of download > 2 Gb
HI Everyone,

I'm using onionshare 2.2 on a windows 7 64 install.  I've been playing around with sharing different filesizes, and I've noticed that as soon as someone tries to download from a shared URL via TOR, onionshare on the server end (where the file is being uploaded from) crashed as soon as the recipient clicks the "download files" button.

There's no error message of any kind,  onionshare simply closes and that's it.  

It successfully starts tranferring for files up to ~1.9gb no problem, btu as soon as the 2gb  boundary is crossed it crashes.

This is also the same if I try to share two files that total more than 2gb in size.

Has anyone else noticed this, and is there a work around for this behavior?

mig5:
Fixes #1089 
Fixes #1095 

The progress bar crash in #1095 has something to do with 32 bit integers in C/Python (I don't quite understand it) - a 2GB file, expressed in bytes, exceeded that integer and caused the crash. 

I copied the fix from the other open source project mentioned in that ticket, and it seems to work.

See also https://www.networkworld.com/article/3010974/whats-so-special-about-2147483648.html

### DIFF
--- a/onionshare_gui/tab/mode/history.py
+++ b/onionshare_gui/tab/mode/history.py
@@ -119,7 +119,7 @@ class ShareHistoryItem(HistoryItem):
         self.progress_bar.setAttribute(QtCore.Qt.WA_DeleteOnClose)
         self.progress_bar.setAlignment(QtCore.Qt.AlignHCenter)
         self.progress_bar.setMinimum(0)
-        self.progress_bar.setMaximum(total_bytes)
+        self.progress_bar.setMaximum(total_bytes / 1024)
         self.progress_bar.setValue(0)
         self.progress_bar.setStyleSheet(
             self.common.gui.css["downloads_uploads_progress_bar"]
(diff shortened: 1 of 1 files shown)


onionshare_gui/tab/mode/history.py
```diff
@@ -461,7 +463,7 @@ def update(self, downloaded_bytes):
         self.downloaded_bytes = downloaded_bytes
 
         self.progress_bar.setValue(downloaded_bytes)
```
SaptakS, line 465: Shouldn't this line also be `self.progress_bar.setValue(downloaded_bytes)` ?

mig5:
I *think* I have fixed it, but if anything it's almost accidental, as I feel a bit muddled between all the total_bytes vs downloaded_bytes and which ones need to be converted to MB or why others don't seem to.

So I'd appreciate others closely looking at this code (not just the bits I changed but those around it). Maybe there's a more DRY way than all these `/ 1024` calculations.

PR #1124, opened 2020-05-28
#1123 Load the ajax.gif in receive.js via relative directory, now that static path is not... static

### LINKED ISSUE #1123: ajax.gif has hardcoded /static URL in receive.js
We have a dynamically-generated 'static' path now, but the receive.js still tries to load an ajax.gif from the static/img dir with hardcoded 'static' folder name, causing a 404.

```
127.0.0.1 - - [28/May/2020 14:07:43] "GET /static/img/ajax.gif HTTP/1.1" 404 -
```

mig5:
Fixes #1123

This was more complicated than I expected (I'm not a Javascript guy). Stack Overflow to the rescue. It works though.


mig5:
@SaptakS thank you for the review! 

As a sysadmin by trade, frontend javascript is not my skill.

Your solution is much more elegant and (even better :) ) it works, please see my latest commit.

PR #1127, opened 2020-05-31
Add QR Code for onion URL

### LINKED ISSUE #1115: Show onion address as QR code [suggestion]
I just used onion share to send videos from Laptop to Phone. I would have like at that moment a QR code to scan instead of sending the link to myself. Food for thought!

mig5:
Fixes #1115

Note: on my iPhone, unfortunately some observations that I don't think we can hope to overcome:

1) The main Camera app detects the QR Code but it always wants to open it in Safari (no option to just copy the URL, and then open it in Onion Browser)

2) I have a separate app 'QR Code Scanner'. This app *does* let me simply copy the URL. I can paste it to someone in a Signal message, for example. But when I paste it into Onion Browser, the HTTP auth user/pass is missing, and I have to manually enter those. 

These are ultimately limitations of the app(s) abilities to recognise the HTTP auth string before the URL. It's not an OnionShare or QR Code issue.

I'm dubious, as a result, about the usefulness of this feature, but maybe it's easier on Android, and at least sending it via Signal (or equivalent) works great, if you've got a QR Code app that allows copying the URL to the clipboard.

onionshare_qr_code


mig5:
P.S sorry about the weird changes to the poetry.lock etc. I'm new to poetry, and I don't know why it changed some category from 'dev' to 'main'. All I had done was run `poetry add qrcode` in order to get it added to the toml file, but in the process, it seems to have changed some other stuff. Maybe it's ok, not sure.

PR #1128, opened 2020-05-31
Fail early if any of the tests fail, rather than continue and possibly exit 0

mig5:
I was watching some tests run, and some failed at https://circleci.com/gh/mig5/onionshare/365?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

In this case it was a transient error probably only affecting that one virtual environment. However, the overall tests ended up reporting a success globally, and so a green tick to Github too, which I did not expect.

That's because the test script iterates over each .py file individually and runs pytest against it, but if one fails, it will happily keep going on the others, and if the last one succeeds, it'll return success.

This change makes it `exit 1` as soon as a failure happens. It might mean more errors on transient CircleCI build issues, but we don't want to risk missing any *real* test failures accidentally.

I guess `set -e` would achieve the same thing, feel free to close and do that if it suits your style more.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -20,7 +20,7 @@ do
 	shift
 done
 
-pytest $PARAMS -vvv ./tests/test_cli*.py
+pytest $PARAMS -vvv ./tests/test_cli*.py || exit 1
 for filename in ./tests/test_gui_*.py; do
-	pytest $PARAMS -vvv --no-qt-log $filename
+	pytest $PARAMS -vvv --no-qt-log $filename || exit 1
 done


mig5:
Hah, irony, it failed again on the same Receive Mode upload test during this PR's tests:

https://circleci.com/gh/mig5/onionshare/367

I re-ran it and it worked for all three Python variants this time... https://circleci.com/workflow-run/4b8bc707-5289-4128-9398-e64c6bc3fddf

I guess we have a time-based variable here that is causing intermittent failures for uploads (it's not specific to Python version). Possibly something to do with checking for whether the uploaded file exists, due to the timestamp-based folder structure.

At least the whole process reports a Failure now though! :)

Also not sure why tests don't report in here sometimes, something to do with me having hooked up my fork to CircleCI I guess. I might remove that as it's getting annoying that it doesn't always trigger here.

PR #1132, opened 2020-06-07
#1126 update the status bar each time the active tab is changed

### LINKED ISSUE #1126: With tabs, the status bar indicator gets overridden by status of other tabs
Now that we have tabs, if you start, schedule or stop a share/receive mode tab, the status, as indicated in the Server Status bar at the bottom of the application, is set and seen in any tab.

In other words, the Server Status is not per tab, it's a global status bar.

So if I start a Receive mode, I get 'Receiving', but if I open or switch to a 'Share' tab, it will continue to say Receiving. 

Conversely, if I stop a server, but navigate to a Share tab that is still running, it will now say 'Ready to receive' or 'Ready to share'.

Either the status bar needs to be set per tab, or it needs to no longer show the indicator (we already have the indicator icon in the tab title itself, maybe this is enough...). 

Or, as per #1082, perhaps the status bar should merely show feedback to the user that OnionShare is connected to Tor or disconnected (but I'm not sure that ticket is a good idea, we don't necessarily need to talk about another program called 'Tor' in an app called OnionShare for UX purposes)

mig5:
Fixes #1126 ! We now connect to the tab change signal and update the status bar to reflect the status of the current mode (or clear it if there is no mode set yet in the new tab).

### DIFF
--- a/onionshare_gui/tab_widget.py
+++ b/onionshare_gui/tab_widget.py
@@ -60,6 +60,7 @@ class TabWidget(QtWidgets.QTabWidget):
         # Use a custom tab bar
         tab_bar = TabBar()
         tab_bar.move_new_tab_button.connect(self.move_new_tab_button)
+        tab_bar.currentChanged.connect(self.tab_changed)
         self.setTabBar(tab_bar)
 
         # Set up the tab widget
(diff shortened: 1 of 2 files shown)


onionshare_gui/tab_widget.py
```diff
@@ -108,6 +109,24 @@ def move_new_tab_button(self):
...
+        try:
+            mode = self.tabs[tab_id].get_mode()
+            if mode:
+                # Update the server status indicator to reflect that of the current tab
+                self.tabs[tab_id].update_server_status_indicator()
+            else:
+                # If this tab doesn't have a mode set yet, blank the server status indicator
+                self.status_bar.server_status_image_label.clear()
+                self.status_bar.server_status_label.clear()
+        except KeyError:
```
SaptakS, line 125: :+1:

PR #1139, opened 2020-07-01
Flatpak support

### LINKED ISSUE #910: Flatpak support

micahflee:
Resolves #910

This makes all of the changes to OnionShare that are needed to support the [flatpak](https://flatpak.org/) packaging. The actual OnionShare flatpak repo is: https://github.com/micahflee/org.onionshare.OnionShare (for now, once it's ready I plan on publishing it in flathub's repo)

There are instructions for testing it in that repo. But in short, this is how you build and run it:

```sh
git clone https://github.com/micahflee/org.onionshare.OnionShare.git
cd org.onionshare.OnionShare/
flatpak remote-add --if-not-exists --user flathub https://flathub.org/repo/flathub.flatpakrepo
flatpak-builder build --force-clean --install-deps-from=flathub --install --user org.onionshare.OnionShare.json
flatpak run org.onionshare.OnionShare
```

This will build all of the dependencies from source (pyqt5, tor, obfs4proxy) and install all of OnionShare's poetry dependencies too.

Right now, the `org.onionshare.OnionShare` repo pulls OnionShare from the `910_flatpak` branch. Once we make the 2.3 release I will go through the [process](https://github.com/flathub/flathub/wiki/App-Submission) of getting OnionShare into flathub.org, and instead pull from the `v2.3` tag.


mig5:
This is my first time trying flatpak, I had to add this to my .bashrc before anything would work:

```
export XDG_DATA_DIRS="/home/user/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share
```

After waiting a looooooooong time for it to compile, it failed right at the end:

```
error: /home/user/git/org.onionshare.OnionShare/build/export/share/icons/hicolor/scalable/apps/org.onionshare.OnionShare.svg is not a valid icon: Image too large (1024x1024)

Export failed: Child process exited with code 1
```

I can't see the svg file in the original repo so I'm not sure how to fix or move past this? Any ideas?

micahflee:
It's strange that you ran into this problem and I didn't. Especially with an "Image too large" error, because the image itself is a vector graphic. What OS are you using? I've been testing in Ubuntu 20.04, but my guess is you're using fedora? Still, it shouldn't matter for flatpak...

In any case, I switched from using an svg to a 512x512 png. Maybe that will fix it for you.

Make sure to `git pull origin master` in the org.onionshare.OnionShare repo (it points to the latest commit in the `910_flatpak` branch now), and then re-run the build command:

```sh
flatpak-builder build --force-clean --install-deps-from=flathub --install --user org.onionshare.OnionShare.json
```

It should be much faster the second time because it shouldn't have to re-build Qt5. Hopefully this will work for you now.

> I can't see the svg file in the original repo so I'm not sure how to fix or move past this? Any ideas?

The svg file is actually in `install` in the OnionShare repo, and it gets put in the right place in `setup.py`.

mig5:
Thanks, that does appear to have fixed it!

> What OS are you using? I've been testing in Ubuntu 20.04, but my guess is you're using fedora? Still, it shouldn't matter for flatpak...

This is a Debian 10 Qubes appVM. I installed flatpak and flatpak-builder just with `apt` from the vanilla Debian repos.

```
user@onionshare:~/git/org.onionshare.OnionShare$ flatpak --version
Flatpak 1.2.5
user@onionshare:~/git/org.onionshare.OnionShare$ flatpak-builder --version
flatpak-builder 1.0.5
```

Things seem to work OK. I noticed that without persistence enabled, if I start and stop a share over and over, I keep getting the same HTTP auth string and Onion address, as if persistence was enabled. I'm not sure if that's a bug specific to flatpak work or (more likely?) specific to OnionShare itself. Might open a separate ticket for that and investigate more.

micahflee:
> Things seem to work OK. I noticed that without persistence enabled, if I start and stop a share over and over, I keep getting the same HTTP auth string and Onion address, as if persistence was enabled. I'm not sure if that's a bug specific to flatpak work or (more likely?) specific to OnionShare itself. Might open a separate ticket for that and investigate more.

Oh actually, if I recall correctly I think this is on purpose in OnionShare in general. I think I made it so each tab remembers its own onion address, but if you close the tab and open a new one you'll get a new onion address. I don't remember the decision making behind it, but I seem to remember thinking this would be more intuitive for how things should work at the time or something -- sort of like, if you refresh a page in Tor Browser your cookies are still set, or something.

As far as this PR in general goes though, I don't think we need to worry about any issues with actually building the flatpak package. Just the functionality of the resulting built package itself. Hmm, I wonder if I can run the tests within the flatpak container somehow...

mig5:
Yup, I ended up opening https://github.com/micahflee/onionshare/pull/1146 before realising your 'persistence' thing was deliberate.

Cool, well from a flatpak perspective it all looks good to me..

micahflee:
Awesome! I'm trying to figure out how to run the tests within the flatpak container, and no luck so far. But everything I've manually tested seems to work, so I'll go ahead and merge!

PR #1140, opened 2020-07-05
Adds responsiveness in receive and share modes webapps

### LINKED ISSUE #886: OnionShare share and receive mode pages should be responsive/mobile-first
I just looked at OnionShare share and receive mode pages in Tor Browser for Android, and while both work great, neither use responsive design, so it's like viewing a desktop-only website on your phone. We should redesign these pages to support mobile browsers.

SaptakS:
Made some adjustments that fixes #886 mostly. 
I have mostly not changed the primary design and hence it is still the same as the current design, but I have made it responsive for smaller screens so they don't have irritating scrolls and all.

 I realised that in quite some places we were using `float` and `table` which though useful in their own rights have a tendency to break when used for layouting. So I have replaced those with a more flex based layouting.

Receive Mode
-------------------
Screenshot_2020-07-05 OnionShare(3)

Share Mode
-----------------
Phone screen:
Screenshot_2020-07-05 OnionShare(1)

Tab screen:
Screenshot_2020-07-05 OnionShare(2)

Laptop screen:
Screenshot_2020-07-05 OnionShare

I have done my bit of testing, but more the testing the merrier, might help to identify other parts which break.

### DIFF
--- a/share/static/css/style.css
+++ b/share/static/css/style.css
@@ -12,6 +12,13 @@ body {
   font-family: Helvetica, sans-serif;
 }
 
+.d-flex {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  align-items: center;
+}
+
 header {
   border-bottom: 1px solid rgba(0, 0, 0, 0.1);
   background: #fcfcfc;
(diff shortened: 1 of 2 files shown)


micahflee:
Hey thank you this is awesome. However sadly, I tested in both Tor Browser for Android and OnionBrowser for iOS, and it turns out that neither of these seem to render properly for small screens. But my guess is that there's nothing we can do about this -- in order to avoid browser fingerprinting, they probably don't act like mobile browsers.

Still though, I think it's reasonable to merge this PR and maybe better solve the problem some day.

Issue #1145, opened 2020-07-06
Persistent mode is stuck enabled


mig5:
In `develop` branch, something has happened whereby persistence mode is on even though the setting is off in the preferences.

Starting/stopping any mode re-uses the same onion address and auth string

```
[Jul 06 2020 10:51:05 AM] Web.generate_static_url_path: new static_url_path is /static_vmgl6i6wc673wbinhx5huctdx4
[Jul 06 2020 10:51:05 AM] Web.generate_password: saved_password=None
[Jul 06 2020 10:51:05 AM] Web.generate_password: built random password: "deceiving-primary"
[Jul 06 2020 10:51:05 AM] OnionShare.start_onion_service
[Jul 06 2020 10:51:05 AM] Onion.start_onion_service: port=17631
[Jul 06 2020 10:51:05 AM] Onion.start_onion_service: key_type=NEW, key_content=ED25519-V3
[Jul 06 2020 10:51:09 AM] ModeSettings.set: updating spoiled-planner-washhouse: general.service_id = mcv3dblva6ytfjpscnz3rhrwikzohsdbje6vcase4a2cwfmu2mthc6yd
[Jul 06 2020 10:51:09 AM] ModeSettings.set: updating spoiled-planner-washhouse: onion.private_key = 8EhOBFeCjjR1pSrkW30ZrOfyG2DNy3h2GE4byNmBX3W3ozJLrIe735Wfk2Hvnp96NL1xNu79CJEc7MHWv4bHhQ==
```

Then I stop:

```
[Jul 06 2020 10:51:14 AM] Onion.stop_onion_service: onion host: mcv3dblva6ytfjpscnz3rhrwikzohsdbje6vcase4a2cwfmu2mthc6yd
```

Then I start (remember, persistent mode is *unchecked* both times)

```
[Jul 06 2020 10:51:15 AM] Mode.start_server: Starting an onion thread
[Jul 06 2020 10:51:15 AM] OnionThread.__init__
[Jul 06 2020 10:51:15 AM] OnionThread.run
[Jul 06 2020 10:51:15 AM] Web.generate_static_url_path: new static_url_path is /static_ipva2mng3svplcsnfsj2c5xeku
[Jul 06 2020 10:51:15 AM] Web.generate_password: saved_password=deceiving-primary
[Jul 06 2020 10:51:15 AM] Web.generate_password: saved_password sent, so password is: "deceiving-primary"
[...]

PR #1146, opened 2020-07-06
Fix forced persistence

mig5:
This 'fixes' #1145 , however it looks to me like this may have been a deliberate change on your part?

76d109747e7caa861ee1a05e70334679d1942b04

> Move private_key, hidservauth_string, and password from "persistent" mode settings to "onion" mode settings; and make it so the onion settings are always saved in each tab, even if the tab is not persistent, so if you stop and start a service in the same tab it has the same onion address and password

This to me sounds like every tab is persistent during runtime. If you really wanted it that way then feel free to reject this. It felt like a bug to me, in that I expect onion addresses and related settings (auth password etc) to be totally ephemeral unless I enable persistent mode, but maybe you figured that's not what the average user expects.

### DIFF
--- a/onionshare_gui/tab/mode/mode_settings_widget.py
+++ b/onionshare_gui/tab/mode/mode_settings_widget.py
@@ -209,8 +209,11 @@ class ModeSettingsWidget(QtWidgets.QWidget):
             self.tab.tab_id, self.persistent_checkbox.isChecked()
         )
 
-        # If disabling persistence, delete the file from disk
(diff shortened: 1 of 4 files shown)


mig5:
I branched from the wrong branch. But anyway, I think this was deliberate on your part.. so it can be ignored... 

If you change your mind, the patch is at  https://github.com/mig5/onionshare/commit/7a16f86ad8c8f373f6c8d8c2d84f8357bb28509f

PR #1167, opened 2020-08-25
OnionShare documentation website

### LINKED ISSUE #1164: Rewrite and finalize documentation for version 2.3
The documentation should be able to localize, and I'm leaning towards [sphinx](https://www.sphinx-doc.org/en/master/) and [Read the Docs](https://readthedocs.org/) (or self-hosted).

Now that I'm starting to rewrite the docs, I like the idea of doing this each release -- updating the documentation to include descriptions of all the newest features. In order to do that I need to take screenshots (and then refresh those screenshots in each version).

So the tasks for this issue are:

- [x] Version bump to 2.3 and update changelog (since the version is in the settings screenshot)
- [x] Rewrite documentation, including taking new screenshots
- [x] Remove old screenshots from the readme (we don't need them in multiple places)
- [x] Update any links within the app to point to the new https://docs.onionshare.org/ site (which is where I assume the docs will live) instead of the old github wiki

micahflee:
Resolves #1164 

Here is a live copy of the website to review: https://micahflee.keybase.pub/docs.onionshare.org/


4jNsY6fCVqZv:
@micahflee 
If you also want to offer a FLOSS-compatible direct Flatpak download at https://onionshare.org/dist/ - see discussion at https://github.com/micahflee/onionshare/issues/910#issuecomment-559288616 - then it would be great if you could add this as a hint in the Linux installation section.
How do you feel about this?

micahflee:
> @micahflee
> If you also want to offer a FLOSS-compatible direct Flatpak download at https://onionshare.org/dist/ - see discussion at [#910 (comment)](https://github.com/micahflee/onionshare/issues/910#issuecomment-559288616) - then it would be great if you could add this as a hint in the Linux installation section.
> How do you feel about this?

Oh yes, good idea. It's hard to write that documentation though before we publish the Flatpak package because I'm not entirely sure how it will work yet. I think maybe we should make the release without that, and then update the documentation shortly after the release with more information.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # OnionShare Changelog
 
+## 2.3
+
+* Major new feature: Multiple tabs, including better support for persistent services, fasting Tor connections
```
mig5, line 5: @micahflee 'fasting'

mig5 (approved):
@micahflee looks great - I didn't read through every file in the diff as it's pretty intense, but I read through the documentation and it looks very good.

I wondered if we need a bigger section on 'Stealth Mode / Client Auth' other than the passing remark under 'Legacy Addresses' - even though it's only possible with Legacy Addresses right now - but it's probably ok, and we shouldn't block this PR on any improvements to be made to the docs, which will always be a source of continuous improvement.

Only saw one small typo which is actually in the CHANGELOG rather than the new docs site itself.

micahflee:
> I wondered if we need a bigger section on 'Stealth Mode / Client Auth' other than the passing remark under 'Legacy Addresses' - even though it's only possible with Legacy Addresses right now - but it's probably ok, and we shouldn't block this PR on any improvements to be made to the docs, which will always be a source of continuous improvement.

I was considering writing a long section on client authentication but decided against it since using legacy addresses is not recommended and about to get deprecated. And when we add client authentication for v3 onion services, the instructions will be totally different anyway, so I figured we can rewrite them in the docs for that version instead.

4jNsY6fCVqZv:
@micahflee Thanks for the revision of the documentation! I'm fine with it, when the Flatpak under Linux part will be extended later.

> It's hard to write that documentation though before we publish the Flatpak package because I'm not entirely sure how it will work yet.

Could you elaborate on that, please?

PR #1176, opened 2020-09-07
Adds UI/UX enhancements in different sections of tab layouts

### LINKED ISSUE #1162: Increase width of the URL container
Follow up to #1154 , #1073 

- [ ] The URL container should be moved a little below (below both text and image_
- [ ] should expand the entire width.

SaptakS:
Things added:

- Moved the Onion URL address below to take full width
- Increased Size of Title text
- Changed color of title text to purple
- Reduced the size of images in share and website
- Added title to cards in the main tab page
- Reduced size of images
- Adds Onionshare logo in the main page

Fixes #1162


micahflee (changes requested):
Nice! So I have a little feedback. Here are screenshots of the 4 modes:

Screenshot from 2020-09-07 15-22-45
Screenshot from 2020-09-07 15-22-47
Screenshot from 2020-09-07 15-22-50
Screenshot from 2020-09-07 15-22-52

It's nice that share and website mode no longer feel cramped and the URL bar stretches all the way across. But this isn't the case with receive or chat mode.

- [ ] Can you make it so that in all modes the URL bar stretches all the way across at the bottom of the window, and looks consistent?

There is a lot of blank space beneath the history widget (in share and website mode now, but in receive mode too soon). I know that it's because the ServerStatus widget is vertically stacked and has the server button at the top.

- [ ] Could you make it so the history widget stretches all the way down?

I know this might be difficult because it might involve moving the button itself outside of the ServerStatus widget and into a separate widget. But maybe another way of doing it is, when the server is started, you can move the red stop button directly to the right of the URL bar? Sort of like this:

Screenshot from 2020-09-07 15-36-57

But I don't know I'm open to other suggestions too.

SaptakS:
Good point. I remember in our discussion @glenn-sorrentino had said to make the history full height in receive mode. So in a way, I missed to do the same for share and website mode. So @micahflee  in that case, the question is, do we want to let the history pane take the entire height in share and website mode also like in receive mode? That removes the issue of blank space and moving button also.

In receive mode also, the URL is much more stretched compared to the current develop one. I also feel, in chat mode, it won't look that good with a single button being in the extrem left bottom. And will look disoriented.

glenn-sorrentino:
Where’d we land with this? I’d recommend that all of the panels use full height.

micahflee (approved):

PR #1179, opened 2020-09-14
Fix Debian and Fedora packaging

micahflee:
This fixes the `build_deb.sh` and `build_rpm.sh` scripts, and the build instructions for building Debian and Fedora packages.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -80,13 +80,13 @@ Install the needed dependencies:
 **For Debian-like distros:**
 
 ```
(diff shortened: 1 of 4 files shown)


mig5:
Just a note that those packages don't exist on Debian 10 (stable), which only came out a year ago

```
E: Unable to locate package python3-nautilus
E: Unable to locate package python3-socketio
E: Unable to locate package python3-flask-socketio
```

Will that cause a problem for things like Tails?

micahflee:
Oh interesting -- I was doing all my testing in ubuntu 20.04, where those packages do exist.

But I think it's ok. The version of OnionShare in Tails is still incredibly old, and the Tails project is indeed considering starting to use Flatpak to have more modern software, and that's sandboxed with Linux containers: https://gitlab.tails.boum.org/tails/tails/-/issues/15874 -- I don't know how far they are on that though.

But in any case, I'm fine with merging this as is. Debian 10 users (and anyone else) can use flatpak.

PR #1181, opened 2020-09-18
Norwegian change

emmapeel2:
We had two ways of calling Norwegian in our translation files and that was creating problems on the translation platform.

I have removed the new way because the other was already used on the app and in other translations in the weblate platform (unrelated to OnionShare, but the users configure their language like that).

Please review, as I am not sure if this is the only place where the change is needed.


emmapeel2:
Oops, I have sneaked a translation update on the merge request: 538a754
Let me know if I have to make a new merge without it, as it is unrelated to this merge request (too many repos!)

micahflee (approved):

comradekingu:
nb_NO is indeed the correct one. See the arrows on http://people.skolelinux.org/pere/blog/Spr_kkoder_for_POSIX_locale_i_Norge.html :)

PR #1191, opened 2020-09-23
Installation documentation reworked

comradekingu:
> Then install OnionShare from Flathub by following `the instructions here <https://flathub.org/apps/details/org.onionshare.OnionShare>`_.

This link doesn't work…

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -1,8 +1,8 @@
 Installation
 ============
 
-Install on Windows or macOS
----------------------------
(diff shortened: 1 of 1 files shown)


micahflee:
> > Then install OnionShare from Flathub by following `the instructions here <https://flathub.org/apps/details/org.onionshare.OnionShare>`_.
> 
> This link doesn't work…

The link won't work until it's released and available on Flathub, but it should be the correct link.

PR #1206, opened 2020-10-31
Make CLI arg --autostop-sharing false by default

whew:
`args.autostop_sharing` defaults to `True` and the argument parser's action is `store_true` which means the argument will always be `True`.
This PR sets the default to `False` which is probably what was intended.


micahflee:
Thanks for this, but I'm fixing it in a slightly different way in #1236

PR #1212, opened 2020-11-10
Remove confusingly-worded QR code label

### LINKED ISSUE #1203: Translator's suggestion to change wording re: QR code device
https://hosted.weblate.org/translate/onionshare/translations/en/?checksum=d209859ffcab3ae8

Original string `gui_qr_code_description`

> Scan this QR code with a QR reader, such as the camera on your phone, in order to more easily share the OnionShare address with someone.

Suggestion:

> A device’s camara is not a QR reader, per se. Should we rather talk about an app?
> Also maybe rather “on your device” since OnionShare is not limited to phones.

micahflee:
Fixes #1203


SaptakS (approved):
LGTM

PR #1216, opened 2020-11-10
Fix username updation to both the sessions

### LINKED ISSUE #1210: Can't change back to previous nickname you had after reloading chat page (v2.3)
When you reload the chat link,you can't change to previous nick you had and therefore you first need to change to a different nickname,then you can change back to nickname you actually had. When that this bug happens,other people can steal your nick.

SaptakS:
Fixes #1210 

Updating the socket session only after a success response is received from the POST request to update the flask session. That way, the conflicting race condition in connected_users is avoided.


micahflee (approved):
Nice, seems to work for me

PR #1230, opened 2020-11-23
Allow directory listing work to with or without trailing slash

### LINKED ISSUE #1053: Website mode issues with folders
There seems to be some confusion (for lack of a better word) for folders in the list under "Publish Website".

After some testing:
* If a single folder is present in the list (i.e. the only one added), the contents of the folder are served. Contents of sub-folders generate 404s.
* If only files are present in the list they appear in a directory listing and load as root sub-resources (is this the right definition?)
* When the contents of the list are mixed (i.e. at least two items and at least one is a folder), I get a 500 error page for the root in the browser (actual response content) although the history panel shows a 200. I also get a 404 for any html content in any of the two folders.

All this suggests that only two cases are workable (undocumented behavior?):
1. Add only files: they will be server under the root (if an `index.html` is present, it will also be the default content)
1. Add only one folder: only the files in it will be served (sub-folders are *not accessible*)

micahflee:
Resolves #1053

### DIFF
--- a/cli/onionshare_cli/resources/templates/listing.html
+++ b/cli/onionshare_cli/resources/templates/listing.html
@@ -1,55 +1,59 @@
 <!DOCTYPE html>
 <html>
- <head>
-    <title>OnionShare</title>
-    <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1"> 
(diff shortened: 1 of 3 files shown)


SaptakS (approved):
Looks to be working for me

PR #1236, opened 2020-11-25
Prevent incomplete downloads in share mode, close after first download

### LINKED ISSUE #929: OnionShare closes before downloading is complete
When I attempt to share a file using the CLI version of OnionShare 2.0 on Ubuntu 18.04 the file that is received is a truncated version of the original.

Attempting to resend the file will result in a different amount of data truncated.

This bug is only present in the CLI version, the GUI version behaves as expected.

When the --stay-open option is used the whole file will be downloaded correctly but OnionShare will report that the download has completed long before Tor Browser does.

The problem does not appear to occur when sharing vary small files but is consistent when sharing anything around 10 MB or larger.

micahflee:
Fixes #929.

When you start sharing files with the CLI with "close after first download" enabled (which is default), as soon as the files are done uploading from your computer, OnionShare quits. If you're using bundled Tor (which is default), it also kills the `tor` process.

However when you're done uploading files, the recipient might not yet be done downloading those files because the data may still be traversing the Tor circuit. When `tor` is killed, it also kills the circuit, which essentially cancels the download before it's done. This problem is very similar to the problem in #812, and that was solved in tor by not closing the rendezvous circuits when an onion service was closed (https://trac.torproject.org/projects/tor/ticket/28619).

So the solution here is, before OnionShare quits, it waits for all of the rendezvous circuits to close and _then_ kills `tor`. This works: it guarantees that there are no circuits in use, so therefore all downloads must be finished. The problem is it takes a long time, I think because you basically have to wait for a circuit timeout. So when a download finishes, it says:

```
Stopped because transfer is complete
Waiting for 1 Tor rendezvous circuit to close ...
```

And then you might need to wait a few minutes before finally:

```
Tor rendezvous circuits have closed
```

And OnionShare quits.

This also includes an alternate implementation of the PR #1206, which basically allows close after first download to be the default (which I think it should be), so that PR can be closed.

~This is still a draft PR because the GUI needs some work now. If you open the OnionShare GUI, start a share with "close after first download", download it, and then quit OnionShare after the download completes, the main window closes but the `onionshare` and `tor` processes are still open, and in the terminal it says you're waiting for the rendezvous circuits to close. If you then try opening OnionShare again, it thinks it's still open and tries opening a new tab, but nothing happens because it's actually just waiting to quit. So basically, this waiting for rendezvous circuits to close needs to be exposed in the UI, so the user can tell that OnionShare hasn't fully quit yet.~

`Onion.cleanup` now only cares about rendezvous circuits for share mode, all other modes are ignored. This makes it less likely the user will have to wait. And when quitting, a dialog pops up informing the user what it's waiting for.

Screenshot from 2020-11-27 14-41-50


micahflee:
I think I can improve this. Download errors are only a problem in share mode, so I think it would be reasonable to only keep track of share onion services, and make sure that their rendezvous circuits are closed, rather than all rendezvous circuits. This could greatly speed things up, and make it so many uses of OnionShare don't have a wait at all when closing.

micahflee:
I've been working on another potential way to solve this problem much better, but haven't gotten it to successfully work yet. Stem allows you to not just run `c.get_circuits()`, but also `c.get_streams()` (https://stem.torproject.org/api/control.html#stem.control.Controller.get_streams) to get a list of active Tor streams, so OnionShare could only kill `tor` once there aren't any more streams.

I've started to implement this but it's frustratingly not working consistently, so maybe just watching for the circuits to close is the best we can get.

mig5:
This is working for me. It took about 8 minutes for the rendezvous circuit to close, in my case. Which is a real shame. But we need file integrity, so I think this is a reasonable compromise. I wonder if we can control the circuit's 'TTL' so to speak, but I don't know the Tor internals well enough (let alone whether that's something that could be set via the control port), or whether it would lead to other issues/risks even if so.

micahflee:
> I've been working on another potential way to solve this problem much better, but haven't gotten it to successfully work yet. Stem allows you to not just run `c.get_circuits()`, but also `c.get_streams()` (https://stem.torproject.org/api/control.html#stem.control.Controller.get_streams) to get a list of active Tor streams, so OnionShare could only kill `tor` once there aren't any more streams.
> 
> I've started to implement this but it's frustratingly not working consistently, so maybe just watching for the circuits to close is the best we can get.

If we can get that to work, it will be much quicker. You'll only have to wait a few seconds after the download finishes, rather than waiting an arbitrary amount of time for a circuit to timeout. I actually did get it working several times, it just wasn't consistent and I couldn't figure out why... so I decided to fallback to this. But if you have some time it would totally be worth trying again.

Also, it sucks that you have to wait so long, but it _only_ applies if you use share mode with close automatically-- all the the other onion circuits are ignored. So that will at least minimize the situations where people have to wait a long time to close OnionShare.

In the meantime, I'm going to merge this PR.

PR #1238, opened 2020-11-29
Find tor correctly

### LINKED ISSUE #1237: Traceback on MacOS when opening SettingsDialog cog icon
On macOS 10.15.7, either via the 'official' v2.3.dev2 dmg or having built from source via `desktop/README.md`, I get this error when I try to click on the SettingsDialog (cog icon):

```
File "/Applications/OnionShare.app/Contents/Resources/app/onionshare/main_window.py", line 228, in open_settings
d = SettingsDialog(self.common)
File "/Applications/OnionShare.app/Contents/Resources/app/onionshare/settings_dialog.py", line 145, in __init__
) = self.common.get_tor_paths()
File "/Applications/OnionShare.app/Contents/Resources/app_packages/onionshare_cli/common.py", line 98, in get_tor_paths
    prefix = os.path.dirname(os.path.dirname(tor_path))
  File "/Applications/OnionShare.app/Contents/Resources/Support/lib/python3.8/posixpath.py", line 152, in dirname
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

I have also been trying to run develop branch via briefcase on Linux, and the SettingsDialog opens, but trying to click on the radio buttons to select Tor modes seems to freeze, or only occasionally 'switch' the active radio button. Not sure if is related or a separate issue yet. That's on Debian 10.

micahflee:
Fixes #1237 

This fixes the logic around finding the `tor` binary, and also displays a proper error message if you run `onionshare-cli` without tor installed (including telling you to install it with Homebrew, if you're using a Mac):

Screen Shot 2020-11-29 at 11 26 33 AM

And when running the desktop version of OnionShare, it just works whether or not `tor` is installed, because the desktop version ships with its own `tor` binary.


micahflee:
Oh yeah this also adds a nice helper script to run the desktop version with `-v` or other CLI args.

mig5:
OK, that definitely seems to fix the issue of finding tor. But now it hangs on the Tor Connection Dialog at 'Starting', with 0% progress shown in the terminal (thanks for the `dev.sh -d` ).

Very mysteriously, if I click 'Test Tor Connection' in the SettingsDialog, it actually connects fine!

But as soon as I hit save, and Tor tries to reconnect 'proper', it hangs on 0%.

micahflee:
Weird... and this happens consistently?

When it says `Connecting to the Tor network: 0% - Starting` the `Starting` is just being returned by the tor controller: https://github.com/micahflee/onionshare/blob/develop/cli/onionshare_cli/onion.py#L380-L394

```py
while True:
    try:
        res = self.c.get_info("status/bootstrap-phase")
    except SocketClosed:
        raise BundledTorCanceled()

    res_parts = shlex.split(res)
    progress = res_parts[2].split("=")[1]
    summary = res_parts[4].split("=")[1]

    # "\033[K" clears the rest of the line
    print(
        f"\rConnecting to the Tor network: {progress}% - {summary}\033[K",
        end="",
    )
```

Basically, it asks tor for the `status/bootstrap-phase`, parses the response, and `Starting` is the `summary` to begin with, and if it's never proceeding past that point then it looks like it's a problem with tor just not connecting.

The difference between OnionShare connecting to Tor on its own and testing Tor in the settings is that when you test Tor is uses a separate temporary data dir. So maybe the problem is something in your data dir is corrupted, which is preventing Tor from connecting?

Can you try deleting `~/Library/Application Support/OnionShare` and then running it again? This should force it to create a fresh data dir.

mig5:
> Can you try deleting ~/Library/Application Support/OnionShare and then running it again? This should force it to create a fresh data dir.

Yep, this solved it, I guess somehow the tor data was corrupt somehow. Very weird.

So this looks good to me, I can connect to bundled Tor on mac now!

mig5 (approved):

PR #1239, opened 2020-11-29
Translate Tor-related error messages

### LINKED ISSUE #1214: Handle localization from cli/onionshare_cli/onion.py
With #1208, the CLI version is not localized at all, so all of the strings are in the desktop version. Tor-related error messages are all only in English, but they should continue to get localized.

micahflee:
Fixes #1214 

A good way to test is with "Attempt auto-configuration with Tor Browser". Change your language to a mostly-translated language, like Español, open Tor Browser, and then set OnionShare to auto-configure. Close OnionShare, close Tor Browser, and open OnionShare again -- you should see the translated error message that Tor Browser is closed.

### DIFF
--- a/cli/onionshare_cli/__init__.py
+++ b/cli/onionshare_cli/__init__.py
@@ -332,7 +332,7 @@ def main(cwd=None):
         print("")
         sys.exit()
     except Exception as e:
-        sys.exit(e.args[0])
+        sys.exit()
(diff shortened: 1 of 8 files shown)


mig5 (approved):
Screenshot_2020-12-02_16-45-14

Yep, looks great!

PR #1243, opened 2020-12-13
Adjusts height of title based on length of text

### LINKED ISSUE #1240: Text on New Tab window gets cut off in other languages when phrases/words are longer
Our 'mode buttons' on the New Tab are of a fixed size. In some languages where phrases are less curt than in English, or if the words themselves are long, they get cut off.

Stretching the main window horizontally doesn't help.

Screenshot_2020-12-02_16-59-16

SaptakS:
Fixes #1240 

I have reduced the font size from 28px to 25px. Also adjusted the height of image a little. And added a logic to check if sizeHint width of the text is greated the geometry width, then adjust the height and starting y.

This is how it looks.
Screenshot from 2020-12-13 20-59-41

If it goes more than 2 lines in wordwrap, then it will again be an issue, but I checked in few different languages and looked okay.

### DIFF
--- a/desktop/src/onionshare/gui_common.py
+++ b/desktop/src/onionshare/gui_common.py
@@ -260,7 +260,7 @@ class GuiCommon:
                 QLabel {
                     text-align: center;
                     color: #333333;
-                    font-size: 28px;
+                    font-size: 25px;
                 }
                 """,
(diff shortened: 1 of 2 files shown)


micahflee (approved):
Works great!

PR #1245, opened 2020-12-13
"Stopped because auto-stop timer ran out" string should get displayed

### LINKED ISSUE #1183: "Stopped because auto-stop timer ran out" string never displays
I'm taking screenshots of all of the strings for weblate, and it turns out the string "Stopped because auto-stop timer ran out" (`close_on_autostop_timer`) never seems to show up.

The server does automatically stop though, so this isn't urgent or anything. I haven't figured out exactly what's happening but maybe this will help in debugging:

In `onionshare_gui/tab/mode/share_mode/__init__.py`:

```py
    def autostop_timer_finished_should_stop_server(self):
        """
        The auto-stop timer expired, should we stop the server? Returns a bool
        """
        # If there were no attempts to download the share, or all downloads are done, we can stop
        if self.history.in_progress_count == 0 or self.web.done:
            self.server_status.stop_server()
            self.server_status_label.setText(strings._("close_on_autostop_timer"))
            return True
        # A download is probably still running - hold off on stopping the share
        else:
            self.server_status_label.setText(
                strings._("gui_share_mode_autostop_timer_waiting")
            )
            return False
```

Update: The same thing happens with the string "Stopped because transfer is complete" (`closing_automatically`). 

Update: The "Disconnected from Tor." string (`gui_tor_connection_lost`) also never gets displayed. It should display when you have OnionShare open and connected to Tor, and then the Tor connection fails (like you disconnect from the internet).

micahflee:
Fixes #1183

### DIFF
--- a/desktop/src/onionshare/tab/mode/__init__.py
+++ b/desktop/src/onionshare/tab/mode/__init__.py
@@ -177,8 +177,7 @@ class Mode(QtWidgets.QWidget):
 
                     self.status_bar.clearMessage()
                     if not self.app.autostop_timer_thread.is_alive():
-                        if self.autostop_timer_finished_should_stop_server():
-                            self.server_status.stop_server()
+                        self.autostop_timer_finished_should_stop_server()
 
     def timer_callback_custom(self):
         """


SaptakS (approved):
Looks good to me.

Issue #1059, opened 2019-10-19
[Bug] Adding/editing resources when site is running lists them but refuses to serve (404)


seafox:
When site folder added before starting site, looks like its resources cached or enumerated internally. When a file added or edited after site is appeared online, the change remains unaccessable although new items are listed. Attempt to navigate on them causes 404 error.

In case of a many files and subdirs that approach will cause resource and performance problems, so it'd be better to allow all new and changed subdirs and files of added dir to be re-fetched on change/addition on demand.

mig5:
> although new items are listed

Not for me. No new changes occur at all while the share is running and this is deliberate. I am not 'live changes' is something we intend to support

seafox:
> > although new items are listed
> 
> Not for me. No new changes occur at all while the share is running and this is deliberate. I am not 'live changes' is something we intend to support or if it is even possible within Flask

Step-by-step to reproduce:
1. Make a dir. That dir would be the only one added before pressing share button.
2. Change to that dir. Place into the dir a few html files but not index.html. Create there also a dir for resources like images and CSS and place a few resources there.
3. Add the dir in #1 in OS website panel. Start sharing.
4. Navigate to the site. You'll see the resource dir and file listing.
5. Add a picture file to the shared dir mentioned in #1.
6. Refresh browser. Same look, no new added file shown.
7. Add the same new picture file to resource dir (that is a subdir of shared dir).
8. Now navigate in browser to that dir. You will see that new picture file name but will get 404 on clicking it.

What I meant by this issue:
1. Ability to make changes into files without need to restart sharing would be convenient, especially with ephemeral addresses.
2. Ability to add/rename resources to a running site without need to restart it would be a huge plus for public sites.
3. Current implementation which uses some rigid cache control would choke when site has significant amount of connected resources. Managing those files dynamically when site is online without interruptions would be essential.

imstanz:
Any progress planned for this enhancement?
It is excactly what I went thru, on my 1st 'publish website' try!

There's always some 'oops', 'sh*t' -- to notice, After it's published and the link has been passed out already!!

Then it's, stop share - add edited files - start share - pass out
link to -many peeps-,
and hope there's no more 'oops' in the html or css!!!

kpcyrd:
`WebsiteModeWeb` has a `files` dict that is populated during publish that holds a dict that maps every possible url to a filesystem path. I assume this is to prevent directory traversal vulnerabilities.

[...]

PR #1256, opened 2020-12-20
[WIP] Read files from disk with every request

whew:
Implements #1059.
I didn't go to great lengths to figure out the best way to do this, but I think it works.

<details>
<summary>Demo</summary>

onionshare
</details>

### DIFF
--- a/cli/onionshare_cli/web/send_base_mode.py
+++ b/cli/onionshare_cli/web/send_base_mode.py
@@ -169,7 +169,7 @@ class SendBaseModeWeb:
         # gzip compress the individual file, if it hasn't already been compressed
         if use_gzip:
             # tempfile.mkstemp automatically opens the file
-            gzip_file_descriptor, gzip_filename = tempfile.mkstemp("wb+")[0]
+            gzip_file_descriptor, gzip_filename = tempfile.mkstemp("wb+")
             # The only way to avoid an eventual "Too many open files" IOError
             # is either to use and close the file descriptor instead of the filename,
             # or to close the file descriptor here and use the filename instead.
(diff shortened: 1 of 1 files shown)


micahflee:
Wow, this is very cool! I will start taking a look soon, and maybe we can get it in the 2.3 release.

whew:
Right now the only change is removing the cache of gzipped filenames in `send_base_mode` so that it reads and compresses the file with each request.

BTW I just realized that this already happens if `should_use_gzip()` is `False`, the file is read from disk with each request so changes are reflected on the website. So if you modify a file after Onionshare has started you'll get different responses with `$ curl` and `$ curl --compressed`. Probably a bug.

There's no change yet to share mode's `/download` logic, it still compresses everything at the beginning only once. Is it a good idea to do that with every request? If you're sharing a lot of files it might take a long time / be CPU intensive if it has to create a ZIP file with every request. Maybe there should be some kind of check to see if files have been modified like `os.path.getmtime` or a hash?

whew:
Share mode's `/download` works now when you're sharing exactly 1 file.
Something I thought, since the size of the gzip file we're sending may change, `self.gzip_filesize` might be wrong which'll affect the GUI progress bar.

whew:
When you share more than 1 file or any directories, it zips them all at startup and sends that whenever `/download` is requested.

The way I see it, if any shared files are modified while the share is up, to be able to reflect the changes when someone requests `/download`, our options are either:
(1) delete the existing zip file and create a new one from scratch, or
(2) modify the existing zip file.

A potential problem is that this is happening when someone requests `/download`, and that request could time out if what we do takes too long. In the case where you're sharing lots of files or big files ([which apparently is a real use case](/micahflee/onionshare/issues/1295#issuecomment-785455947)), solution (1) would be impossible (I tested this, zipping two 1GiB files took 100 seconds).

That leaves (2) but I'm not that familiar with zip files in Python. According to [these StackOverflow answers from 2014](https://stackoverflow.com/q/25738523) you can't modify files in a zip archive directly. And you can't delete files from a zip archive either; that feature request has been in Python's issue tracker [since 2009](https://bugs.python.org/issue6818). There is [a PR from April 2020](python/cpython#19358) that implements deleting from zip files. something like [this](https://stackoverflow.com/a/35435548) maybe?

PR #1262, opened 2020-12-31
Update send_base_mode.py

Kcchouette:
In some case where file is not ascii encoded (eg UTF-8) we have UnicodeEncodeError
```
Error on request:
  File "werkzeug/serving.py", line 323, in run_wsgi
    execute(self.server.app)
  File "werkzeug/serving.py", line 315, in execute
    write(data)
  File "werkzeug/serving.py", line 276, in write
    self.send_header(key, value)
  File "http/server.py", line 517, in send_header
    ("%s: %s\r\n" % (keyword, value)).encode('latin-1', 'strict'))
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 39-43: ordinal not in range(256)
```
I adapt commit using https://github.com/pallets/flask/commit/0049922f2e690a6d58f335ca9196c95f1de84370

cheers


Kcchouette:
Oh, I'm letting you upport it to new branches

SaptakS:
@mig5 I have rebased and made few changes based on the requirements due to the big migration. This works as intended, and should be ready for review and merge.

micahflee (approved):

Issue #1186, opened 2020-09-22
New tabs layout cuts off some wording in macOS in share mode, website mode


micahflee:
But there's enough padding so that I think it could be not cut off, and still use the same font size.

<img width="1152" alt="Screen Shot 2020-09-21 at 9 14 25 PM" src="https://user-images.githubusercontent.com/156128/93843668-7abc5400-fc4f-11ea-93c9-abc694168d40.png">
<img width="1152" alt="Screen Shot 2020-09-21 at 9 14 36 PM" src="https://user-images.githubusercontent.com/156128/93843670-7e4fdb00-fc4f-11ea-993e-32fcfdd9bd89.png">

micahflee:
Since switching from PyQt5 to PySide2, this no longer happens, and it looks great in macOS

PR #1266, opened 2021-01-17
Fixes macOS layout while maintaining similar layout in linux

SaptakS:
In macOS, in share and website mode, the text gets cut off unless
it is full screen. This is mostly because of margin being applied
by css. Here I have tried to implement a similar layout using qt
instead of CSS so it behaves reasonably everywhere.

I was able to reproduce https://github.com/micahflee/onionshare/issues/1186. This
fix doesn't change the current layout much while fixing it in my mac as well.

Before (in debian):
Screenshot from 2021-01-09 02-09-34

After (in debian):
Screenshot from 2021-01-09 02-10-07

Before(in mac):
signal-2021-01-18-010320

After (in mac):
signal-2021-01-18-010604

### DIFF
--- a/desktop/src/onionshare/gui_common.py
+++ b/desktop/src/onionshare/gui_common.py
@@ -292,12 +292,10 @@ class GuiCommon:
                 QLabel {
                     color: #4E064F;
                     font-size: 48px;
-                    margin-bottom: 72px;
                 }""",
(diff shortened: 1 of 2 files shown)


micahflee (approved):
Nice!

PR #1268, opened 2021-01-23
MacOS dark mode implementations

### LINKED ISSUE #1250: Support macOS dark mode
Right now it's not all that supported. Also, the text is cut off.

Screen Shot 2020-12-13 at 8 17 24 PM

SaptakS:
Resolves #1250 

Following are some screenshots:
Screenshot_2021-01-24_at_3 46 52_AM
Screenshot_2021-01-24_at_3 45 02_AM
Screenshot_2021-01-24_at_3 42 23_AM

PS: Right now, I can detect if the system color mode is changed while the app is open, but it doesn't change the colors right now. But if the system is already in dark mode, and onionshare is opened, then it opens perfectly with all colors. I think on detecting the color change, we either need to relaunch the app or reinitialize the MainWindow. @micahflee @mig5 thought??


SaptakS:
Also updated the copy url button styles to default OS button style with some padding:

Screenshot from 2021-01-24 04-15-35

micahflee (approved):

micahflee:
I think this is great, and honestly I don't mind the dialog telling you to restart if you want the colors to change. If you're in the middle of something important you'll just keep it, and otherwise you'll restart.

PR #1272, opened 2021-02-09
Correct rendevouz to rendezvous

mig5:
Just a typo that stung some half-french eyes :)


micahflee:
The test fail was a false positive. I pushed it to my own branch and the tests passed.

micahflee (approved):

PR #1278, opened 2021-02-21
Delete unused Qt5 DLLs from the Windows release

### LINKED ISSUE #1276: Reduce file size of Windows binary
The new system of using briefcase and PySide2 is nice, but it ends up packaging *all* of PySide2, not just the parts that we need. The macOS build script deletes all of the unused Qt5 frameworks as part of the build process: https://github.com/micahflee/onionshare/blob/develop/desktop/package/macos/build.py#L72

We need to do this for the Windows version as well. Without doing this, the Windows installer is 148mb and after installing it's 374mb. I believe if we delete all the unused Qt5 stuff we can reduce the installer size down to about 110mb and the installed app down to about 280mb.

micahflee:
Resolves #1276 

Instead of the Windows OnionShare software taking 374mb, it now only takes 272mb, and everything still works.


SaptakS (approved):
I will trust you on this that you tested on windows and approve. :)

PR #1279, opened 2021-02-21
Set content-disposition to attachment for /download

### LINKED ISSUE #1277: Downloads should always have content-disposition: attachment
In share mode, when you click "Download Files", it links to `/download`, which responds with the content of the file being shared. We should explicitly set the response header `content-disposition: attachment` to force the file to download rather than get displayed in the browser.

I tested sharing a 101mb .msi file, and when I click download it tries displaying this huge file in the browser. It should always download it instead.

micahflee:
Resolves #1277


SaptakS (approved):
Looks good to me.

PR #1281, opened 2021-02-22
Fix the onion object when checking for a new release via the Settings Dialog

mig5:


SaptakS (approved):
Tested on MacOS. Looks good to me.

PR #1314, opened 2021-03-08
Adding some changes to have old style

Kcchouette:
Following some change during new 2.3 release, the theme is now totally different from before.

The goal of this PR is to revert back to style in screenshot in PR https://github.com/micahflee/onionshare/pull/1140

### DIFF
--- a/cli/onionshare_cli/resources/templates/listing.html
+++ b/cli/onionshare_cli/resources/templates/listing.html
@@ -24,7 +24,7 @@
   {% endif %}
 
   <table class="file-list" id="file-list">
-    <tr>
+    <tr class="heading">
       <th id="filename-header">Filename</th>
       <th id="size-header">Size</th>
       <th></th>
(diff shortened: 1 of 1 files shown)


Kcchouette:
Hello
I saw your PR #1140  @SaptakS

Do you plan of doing it for "website" mode?

Because your changes seems to have broke this mode and I don't want to change your work
eg I would change for "website" mode:

```
.file-list div.d-flex {
  border-bottom: 1px solid #e0e0e0;
  justify-content: flex-start;
}
```

into 

```
.file-list tr {
  border-bottom: 1px solid #e0e0e0;
  justify-content: flex-start;
}
```

Cheers

SaptakS:
@Kcchouette thanks for bringing this up. I would want to make the website mode have similar HTML as in the send html, basically replacing table layouting with flex layouts. If you want, feel free to add the similar changes from #1140 to this PR for website mode. It should be more or less same.

Kcchouette:
Ok I'll try theses few days, I'll ping you again for checking

Thanks

Kcchouette:
@SaptakS it was easier than I think so here the commit :fireworks:

SaptakS:
Thanks! I will look into it by tomorrow!

SaptakS (approved):

PR #804, opened 2018-10-05
range request support for share mode

heartsucker:
Fixes #736

PR #1321, opened 2021-03-30
Reviving the old range request PR

SaptakS:
This commit aims at reviving https://github.com/micahflee/onionshare/pull/804/commits on the newly structured and refactored code. I tried merging/cherry picking the commits to retain the commit history and contributions by @heartsucker but the code has changed substantially, so it was much easier to add and adopt the code from the PR to the new code. Also updated the tests and imports to remove most errors.

I have tested that simple download works, and I tried simple range request with cURL, but haven't tested yet using tor browser or any other tool.

### DIFF
--- a/cli/onionshare_cli/web/share_mode.py
+++ b/cli/onionshare_cli/web/share_mode.py
@@ -186,15 +186,15 @@ class ShareModeWeb(SendBaseModeWeb):
             use_gzip = self.should_use_gzip()
             if use_gzip:
                 file_to_download = self.gzip_filename
-                filesize = self.gzip_filesize
+                self.filesize = self.gzip_filesize
                 etag = self.gzip_etag
             else:
                 file_to_download = self.download_filename
-                filesize = self.download_filesize
+                self.filesize = self.download_filesize
                 etag = self.download_etag
 
             # for range requests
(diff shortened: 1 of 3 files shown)


micahflee (changes requested):
I thought the desktop tests failing were maybe a false positive, but I just ran them manually and they're still failing. Want to figure out what the issue is with the desktop tests before I merge this?

SaptakS:
Yes. I checked as well. It is failing the first test to check to see that new tab is properly rendered. I am yet to reproduce the error thrown by test by running the desktop app locally, but I wil look into it, and try and get it fuxed

micahflee:
I removed deepsource, so if you merge develop into this branch tests should pass.

SaptakS:
@micahflee done!

micahflee (approved):
I confirmed that this works great with `wget`, and I think it's good to merge!

However I did try testing with Tor Browser and I ran into the same problem as before with heartsucker's old PR, resuming a download doesn't work. To pause a download in Tor Browser, go to `about:downloads`, right-click on a download and choose "Pause". You can then right-click and choose "Resume". It still fails.

I think it would be good to fix that, but I don't think it should hold this PR back. I just opened a new issue for that specific problem: #1333

PR #1325, opened 2021-04-12
Webhook notifications for receive mode

### LINKED ISSUE #1322: Receive mode webhooks
If you run a receive mode service, especially a headless one, it's really easy to forget to check on it. Getting notified in some way when a submission comes through would be incredibly helpful. I think the simplest way to make notifications work is to allow an optional webhook URL.

When someone submits something, OnionShare can make a POST request to the webhook with minimal information such as "new submission", and possible the file size.

In the future it might make sense to build out a more robust OnionShare notification service that we run and allow people to use (supporting sending Signal messages, or whatever). But for now, people can get notifications working using something like [Keybase's @webhookbot](https://github.com/keybase/managed-bots/tree/master/webhookbot). When someone submits something, you can get a keybase message.

micahflee:
Resolves #1322.

This lets you optionally set a webhook URL. Whenever someone uploads something to your receive mode service, it makes a POST request to this URL (over Tor, unless you're running with `--local-only`) saying the number of files that were uploaded.

The simplest way to test is to use the keybase webhookbot.

## Create a webhook URL

For example, you can tell it `!webhook create onionshare-test` to generate a webhook URL:

Screenshot from 2021-04-11 21-25-53

## Test in the CLI

```
$ poetry run onionshare-cli --receive --webhook-url https://bots.keybase.io/webhookbot/Ieah6SBHq5fYhA7ZqjoryQWx4y0
```

Load the URL in Tor Browser and upload a few files. In my case, I uploaded 5. Then in keybase, I get this message from webhookbot:

Screenshot from 2021-04-11 21-29-13

## Test in the GUI

Open receive mode and check "Use webhook URL":

Screenshot from 2021-04-11 21-30-10

It has placeholder text:

Screenshot from 2021-04-11 21-30-38

Add the keybase URL, start the server, load it in Tor Browser, upload some files, and get the notification again.


micahflee:
One thing I didn't do is write a test for this. To test it, I'd basically need to start a web server, set it as the webhook URL, and wait for a POST request to come in when I download files, and this seems challenging to do in a test. I'd have to like, maybe run a flask service in a separate thread or something?

It might be worth it to add a test though anyway...

SaptakS:
Can't we mock a webhook URL?So that once it receives a request when a file is uploaded it print some text or something? I can see it will be complicated, but I would prefer mocking rather than running a flask service for testing purpose if possible.

SaptakS (commented):
Tested it via CLI and desktop, looks good to me. My only concern is the notification message and the label for the webhook might not be too verbose.

I would probably prefer "notification webhook URL" for desktop label. Otherwise it's not clear webhook for what exactly.
And in the message something more verbose like "x files uploaded to your onionshare receive instance". Right now if someone doesn't keep a proper name for their bot, it might confuse as to where the file is uploaded.

micahflee:
I fixed a bug where if you unchecked the webhook checkbox, it didn't actually unset the webhook. It does now.

I added a CLI test for webhooks (and also for uploading files at all in receive mode -- that wasn't being tested in CLI, only in GUI).

Now the webhook URL label says "Use notification webhook":

Screenshot from 2021-04-12 09-36-18

And the actual webhook message itself is more verbose too, like:

> 1 file uploaded to OnionShare

or

> 4 files uploaded to OnionShare

Ready for a final review I think.

cli/tests/test_cli_web.py
```diff
@@ -12,6 +13,21 @@
...
+webhook_data = None
+
+
+def requests_post_stub(url, data, timeout, proxies):
+    global webhook_url, webhook_data
+    webhook_url = url
+    webhook_data = data
+
+
+onionshare_cli.web.receive_mode.requests.post = requests_post_stub
```
SaptakS, line 29: This is great!

SaptakS (approved):
Looks good to me

PR #1326, opened 2021-04-12
Make verbose output prettier with terminal colors

micahflee:
When you run OnionShare with `-v` for verbose output, this makes the timestamps dark gray and the messages light gray, to more easily differentiate from the regular output which is white.

Screenshot_2021-04-12_15-42-59


SaptakS (approved):
Looks great!!

PR #1327, opened 2021-04-13
Allow custom title for each onion service

### LINKED ISSUE #1306: Allow optional customizing a title/description for each onion service
Right now all OnionShare services look the same in Tor Browser, like this:

Screenshot from 2021-02-28 11-37-46

It would be cool to allow the user to give their service a specific name, so instead of "OnionShare" at the top it could say "Micah's Anonymous Dropbox" or whatever I choose. And likewise, it would be cool to all an HTML field for a description, so that you could basically write some text explaining whatever you want to explain, include links, etc.

If you give a service a custom title, the GUI can use your custom title in the tab bar too (so "Micah's Anonymous Dropbox" instead of "Receive").

micahflee:
Resolves #1306.

For sake of UX and simplicity I decided to not include the option of setting a custom description and instead just a custom title.

This adds `--title` argument to the CLI, so like:

```sh
onionshare-cli --title "Super Secret Chatz" --chat
```

Looks like:

Screenshot_2021-04-13_18-43-22

I also updated the GUI to allow each tab to let you its title (and re-arranged the settings so title and persistent are at top):

Screenshot_2021-04-13_18-47-40

If you don't set a title, the title in Tor Browser is just "OnionShare" and the title of the tab is the mode (e.g. a receive tab is "Receive"). But if you do set a title it renames the tab (and sets a tooltip, so you can see the full name) and also uses the custom title in Tor Browser:

Screenshot_2021-04-13_18-49-58
Screenshot_2021-04-13_18-51-45

### DIFF
--- a/cli/onionshare_cli/__init__.py
+++ b/cli/onionshare_cli/__init__.py
@@ -144,6 +144,12 @@ def main(cwd=None):
     )
     # General args
     parser.add_argument(
+        "--title",
+        metavar="TITLE",
+        default=None,
+        help="Set a title",
+    )
+    parser.add_argument(
         "--public",
         action="store_true",
         dest="public",
@@ -234,6 +240,7 @@ def main(cwd=None):
     connect_timeout = int(args.connect_timeout)
     config_filename = args.config
     persistent_filename = args.persistent
+    title = args.title
     public = bool(args.public)
     autostart_timer = int(args.autostart_timer)
     autostop_timer = int(args.autostop_timer)
@@ -280,6 +287,7 @@ def main(cwd=None):
 
     if mode_settings.just_created:
         # This means the mode settings were just created, not loaded from disk
+        mode_settings.set("general", "title", title)
         mode_settings.set("general", "public", public)
         mode_settings.set("general", "autostart_timer", autostart_timer)
         mode_settings.set("general", "autostop_timer", autostop_timer)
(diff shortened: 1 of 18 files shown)


mig5:
Seems to work well!

I noticed I could set a title that is just whitespace. It didn't seem to cause any problem though, but it made the tab at the top appear as 'empty' of sorts. Should we require at least one alphanumeric character, or maybe you don't care?

I tried some XSS too just for fun but it all seems to be stripped safely :)

Is 'Title' descriptive enough for people to understand what this field will do? Should it be under 'Advanced Settings' since it's probably not something everyone wants, and we could call it 'Custom Title for Page' or is that too descriptive?

One small strange thing: I noticed that a couple of times, after setting a title, I had to click twice or even three times on either the Start or Stop button before it acted. I haven't tested recent PRs so I'm not sure if it's specific to this PR or maybe something else that's got into develop branch.

Annoyingly, I can't reliably reproduce it, I wonder if you can..

SaptakS:
> Should it be under 'Advanced Settings' since it's probably not something everyone wants

I think I agree to this as well. Right now, it looks kind of confusing for people who don't really care about setting a title. And it also takes precedence appearing before the receive mode folder select option which I think is the main thing in the receive tab.

> Should we require at least one alphanumeric character

I would say we should probably just show the default title in case it is empty (after trimming the spaces)?

micahflee:
Alright, check it out now:

- The label says `Custom title` instead of `Title`
- The title field is moved into advanced settings
- If you try setting the title to whitespace in the GUI, it resets it to blank as if it hasn't been set (of course, you can still set the title as ` ` or whatever in the CLI, but I think that's fine)

SaptakS (approved):
Looks good to me.

mig5 (approved):
Yep, LGTM too :+1:

PR #1332, opened 2021-04-24
Use socks5h instead of socks5 

whew:
With `socks5h` the proxy server resolves the domain name, which avoids leaking DNS requests.

https://github.com/urllib3/urllib3/issues/1035


SaptakS (approved):
Looks good to me! Thanks for the PR!

PR #1340, opened 2021-05-02
Allow sending messages with receive mode

### LINKED ISSUE #936: Empty POST request should not create an empty folder in Receive Mode
When an empty POST request is made in Receive mode, there is no GUI update (per design), and the website reports a Flash message 'No files uploaded'.

However, in the `ReceiveModeRequest` class, a directory is still created for the request, but it is empty since no files.

I tried to check the length of `self.files` in this overridden Request object [1], and if empty skip the folder creation, but it led to further complications: specifically, if I skipped creating the folder by checking anything relating to the Request object's `self.files` attribute, I encountered this AttributeError in the `_get_file_stream` class:

```
AttributeError: 'ReceiveModeRequest' object has no attribute 'told_gui_about_request'
[Mar 08 2019 12:27:16 PM] ReceiveModeRequest.close
```

I was able to treat it as a `self.upload_error`, however the error message that appears for the OnionShare user is hardcoded to 'Could not create the directory' because that's the only real error we envisaged arising from that logic, so it didn't seem suitable either as a message, as well as the fact that throwing an Alert() dialog about an empty POST request seemed a bit excessive (I wanted to basically ignore it)

[1] http://flask.pocoo.org/docs/0.12/api/#flask.Request.files

### LINKED ISSUE #1305: Allow sending messages with receive mode
Right now receive mode just lets you upload files. It should also let you send messages to the user.

The Tor Browser interface should be split in two, where Send Files is on one side, and Send Message is on the other. The Send Message form can have:

- Message text area, where you can write a plaintext message
- Optional reply OnionShare address, where the user can paste the URL of their receive mode OnionShare service in case they want to receive a reply

Sending a message could just essentially save a `message.txt` file, and optional `reply_address.txt` as well.

micahflee:
This PR adds support for sending text messages as well as uploading files in receive mode. Resolves #1305. Resolves #936.

There are two new CLI args for receive mode, `--disable-text` and `--disable-files`. If you just set up a receive mode service without either of these, it lets you submit both a text message or upload files, similar to SecureDrop.

Screen Shot 2021-05-02 at 3 58 42 PM

If you use `--disable-text`:

Screen Shot 2021-05-02 at 4 00 13 PM

If you use `--disable-files`, you can get a text-only receive mode (someone requested this, to use as a secure contact form):

Screen Shot 2021-05-02 at 4 01 48 PM

Before when you uploaded files in receive mode it would take the data dir (`~/OnionShare`), make a folder for the current date (like `2021-05-02`) and then make a folder for the current time (like `16.03.10`, for hours.minutes.seconds). Now it removes the periods from the time, so that folder would just be `160310`.

And if you submit a message, instead of putting it in the folder, it names `160310-message.txt`. If you also submit files in the same POST request, it will put those files in the `160310` folder.

Before if you submitted the form without uploading anything it would make an empty server, which was annoying. Now it makes sure not make an empty folder.

There are new receive mode CLI tests that test for these things.

In the GUI, there are new checkboxes in receive mode for "Disable submitting text" and "Disable uploading files".

Screen Shot 2021-05-02 at 4 07 11 PM

When someone sends a message, there is a "Read Message" button in the history widget. When you click it it opens the `HHMMSS-message.txt` file in the operating system's default text editor, or in `notepad` in Windows.

Screen Shot 2021-05-02 at 4 09 59 PM

Screen Shot 2021-05-02 at 4 10 07 PM

When you upload files and also include a message, it looks like this in the GUI:

Screen Shot 2021-05-02 at 4 14 01 PM

Screen Shot 2021-05-02 at 4 14 47 PM

Here's the directory structure of uploaded files:

Screen Shot 2021-05-02 at 4 15 56 PM

I modified the GUI receive mode test to also test submitting a message and making sure it gets saved.

And finally, it's not quite related but when I was testing this in Windows I realized the ASCII art and colored verbose logs were all borked in Windows and the color didn't work at all. So I ported the color stuff to `colorama`, so now color works great in Windows.


mig5 (changes requested):
Just bringing across my 'review' from Keybase:

1) When you disable-files, you can't submit the form when using Standard or Safer mode in Tor Browser, due to a javascript error:

```
Uncaught TypeError: filenames is undefined
    <anonymous> http://xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.onion/static_xxxxxxxxxxxxxxx/js/receive.js:115
    jQuery 2
```

2) Submitting nothing in either messages nor form, still creates an emptyfolder with today's date (known issue #936 but said to be fixed in this PR). I don't think that's a blocker for this PR unless it's easily fixed, I couldn't figure it out way back in #936 with the Request middleware.

Otherwise so far, this looks to be working great! Tested on Linux and macOS. I love that you can submit both a file and a message at once, in the default configuration.

PR #1344, opened 2021-05-04
Move the cleanup() function from Onionshare class to Web class, so that the list of files to be cleaned up is always available (needed for website temp files)

### LINKED ISSUE #1261: Temporary files aren't deleted in website mode
Before running website mode:
```
$ ls /tmp/tmp*
ls: cannot access '/tmp/tmp*': No such file or directory
```

After getting a resource in the browser:
```
$ ls /tmp/tmp*
/tmp/tmp7_f4m60kwb+
```

After closing Onionshare:
```
$ ls /tmp/tmp*
/tmp/tmp7_f4m60kwb+
```

The temporary gzip file is still there.

mig5:
This should fix #1261

The issue is that in Share mode, the 'temporary' gzipped file is created immediately when the share starts (whether or not someone has visited or fetched it), so the file is already in the OnionShare `cleanup_filenames` list at the right time (early).

But for website mode, those gzip files are generated *only* when someone actually goes to visit the site. At this point, the files do not get appended to the Onionshare class's `cleanup_filenames` list when needed.

There are a number of ways to fix it but the easiest way was to actually move the `cleanup()` function out of Onionshare class and into Web class. We were already appending the list of files to a `cleanup_filenames` list inside the Web class or rather the SendModeWeb sub class, so rather than try to append that list to the outer list, just do the cleanup here.

I hope it makes sense and that it doesn't break anything. For me it seems to fix it. To test:

1) On `develop` branch, run `onionshare-cli --local-only --website ~/git/onionshare-website --verbose` where the onionshare-website is a folder containing a website

2) Visit the Onion share, and notice that at that moment, the temporary files are generated in `/tmp`

3) Ctrl+C and you will see that the tmp files linger. You might as well manually remove them to clear the way for easily validating the next steps.

4) Switch to this feature branch. 

5) Rinse and repeat steps 1-3 and note that when you Ctrl+C, the files are removed.

Similarly, under regular Share mode, the tmp file should still get removed.

Tested in GUI as well as CLI. Possibly also fixed a bug whereby the 'tmp' folder that the zip file is made within, was not being removed (?), not sure, but needed to fix it here in any case.

### DIFF
--- a/cli/onionshare_cli/__init__.py
+++ b/cli/onionshare_cli/__init__.py
@@ -442,7 +442,6 @@ def main(cwd=None):
         print("Compressing files.")
         try:
             web.share_mode.set_file_info(filenames)
-            app.cleanup_filenames += web.share_mode.cleanup_filenames
         except OSError as e:
             print(e.strerror)
             sys.exit(1)
@@ -536,7 +535,7 @@ def main(cwd=None):
         web.stop(app.port)
     finally:
         # Shutdown
-        app.cleanup()
+        web.cleanup()
         onion.cleanup()
 
 
--- a/cli/onionshare_cli/onionshare.py
+++ b/cli/onionshare_cli/onionshare.py
@@ -19,7 +19,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 
 import os
-import shutil
 from .common import AutoStopTimer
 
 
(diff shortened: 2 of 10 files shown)


mig5:
CircleCI is not running tests anymore for my PRs and it's really annoying.. I unhooked CircleCI from my specific fork but maybe I should not have.. anyway, the GUI tests pass locally for me.

micahflee (approved):
Works great!

PR #1349, opened 2021-05-06
Pins flask-socketio to 5.0.1 in pyproject.toml

SaptakS:
Also updates the RELEASE.md to document the need to update
the socket.io.min.js file whenever we want to update
flask-socketio to ensure that they are compatible with each
other, failing which, the chat stops works.


micahflee (approved):
The failing test is a fluke, they pass locally

PR #1353, opened 2021-05-10
Register the 405 error handler properly and other method/error related stuff

### LINKED ISSUE #1253: Onionshare's own 405 page is never shown
When you send a disallowed request e.g. POST to index.html, what you get back is some default 405 page not Onionshare's 405 page. [`error405`](https://github.com/micahflee/onionshare/blob/develop/cli/onionshare_cli/web/web.py#L296) is never actually called.

```
$ curl -x socks5h://127.0.0.1:9050 -X POST -v http://[onionshare].onion/
> POST / HTTP/1.1
> Host: [onionshare].onion
> User-Agent: curl/7.68.0
> Accept: */*
> 
< HTTP/1.0 405 METHOD NOT ALLOWED
< Content-Type: text/html
< Allow: OPTIONS, GET, HEAD
< Content-Length: 178
< Server: Werkzeug/0.16.1 Python/3.8.5
< Date: Fri, 18 Dec 2020 00:25:40 GMT
< 
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>405 Method Not Allowed</title>
<h1>Method Not Allowed</h1>
<p>The method is not allowed for the requested URL.</p>
```

Edit: I tested this on version 2.2.

mig5:
Fixes #1253 

 * Registers the 405 error handler in the proper place
 * Enforces the appropriate methods on each route (note: I'm allowing POST as well as GET in website mode, in case people have a basic form or something, but maybe we could turn that off too? What do you think)
 * Explicitly disables OPTIONS method on each route (it's implicit even if you only set GET and POST etc as above). I don't think we ever really need/want OPTIONS, and when it works, it sends the Werkzeug header
 * Adds a friendlier error page for a 500 internal server error (bugs in OnionShare can surface this)
 * Adds tests for invalid methods in share, receive and website mode, which test that it's *our* 405 custom themed page that is returned, not Flask's
 
Note that HEAD seems implicitly enabled in Flask if GET is enabled. I left that as is.

Screenshot of a 500 internal server error attached (not really practical to add a test for this! But I deliberately broke the Web class to trigger it manually as a PoC)  
Screenshot_2021-05-10_11-22-18


micahflee (changes requested):
Very cool!

> Enforces the appropriate methods on each route (note: I'm allowing POST as well as GET in website mode, in case people have a basic form or something, but maybe we could turn that off too? What do you think)

I don't think website mode needs to allow POST requests since there's no way for flask itself to handle any POST input from arbitrary websites. If there's a basic form on a website mode website, it needs to set the action to some other website for it to do anything, so I think disabling POST is fine.

I think that's it, everything else looks good!

mig5:
Thanks, I've removed POST from website mode (and added it to its 405 test as an expected invalid method).

While working on this I also realised I could now add the 'HTTP 405' test against Chat mode, now that we have Chat mode tests \o/

This in turn actually uncovered a bug: our 'error handler' functions in the Web class, try to add a REQUEST_INDIVIDUAL_FILE_STARTED to the frontend to show the 40X error. This actually fails in Chat mode due to the lack of history widgets and the like.

I've added a new attribute that is evaluated in those functions before sending REQUEST_INDIVIDUAL_FILE_STARTED. I've set it to true by default, but turned it off for Chat mode. This makes it easier to turn it off for future modes too that might not have history widgets.

There might've been a cleaner way to handle this still (maybe as an attribute of the actual mode itself, not in the Web class?), but it works... And the Chat mode now has a test for 405 and invalid methods...

mig5:
Decided to make that an attribute of the modes themselves.. felt a bit cleaner? Maybe? If you preferred it as a simple attribute of the Web class instead, feel free to revert https://github.com/micahflee/onionshare/pull/1353/commits/d4d6eea500feb825d5639354b8aad1be6ba451f5

SaptakS:
> Decided to make that an attribute of the modes themselves.. felt a bit cleaner? Maybe? If you preferred it as a simple attribute of the Web class instead, feel free to revert [d4d6eea](https://github.com/micahflee/onionshare/commit/d4d6eea500feb825d5639354b8aad1be6ba451f5)

I think it makes sense to me.

mig5:
I didn't go so far as to evaluate this new boolean in all the places where we *do* send the `REQUEST_INDIVIDUAL_FILE_STARTED`  - just where it would throw exception to chat mode on 40X error - but I suppose I could evaluate it everywhere it's used just for 'completeness'.. let me know what you think

micahflee (approved):

PR #1358, opened 2021-05-13
Add security headers to every response

### LINKED ISSUE #1357: Default `Server` header reveals user's version of Python/Werkzeug
For some requests we don't override the default HTTP `Server` header that Werkzeug uses. Some of these are `OPTIONS /`, `GET /static/*`, `POST /upload`, and `POST /upload-ajax`.

e.g. `OPTIONS /`:
```
HTTP/1.0 200 OK
Content-Type: text/html; charset=utf-8
Allow: GET, HEAD, OPTIONS
Content-Length: 0
Server: Werkzeug/0.16.1 Python/3.8.5
Date: Thu, 13 May 2021 08:23:30 GMT
```

other examples: https://github.com/micahflee/onionshare/issues/1253#issuecomment-748550114

whew:
Fixes #1357.

We can override the `Server` header in all our responses by using our `add_security_headers` function with Flask's `after_request` decorator. The security headers (which includes `Server`) would be added to all our responses and we'd never need to call `add_security_headers` ourselves.


mig5 (changes requested):
Sorry for the slow review. This seems to be working fine for me, tested that the headers are present in each mode, and that the CSP can still be optionally disabled in Website Mode. GUI tests are also passing.

The only issue is that unfortunately, the PR now has conflicts due to other changes since made in develop branch. Sorry about that. Do you think you might be able to fix the conflicts up? It may not make the 2.3.2 release but hopefully the one after that.

whew:
Resolved

SaptakS (approved):
Looks good to me.

mig5 (approved):
@micahflee this looks good to me too to merge. There was a conflict on `web.py` again due to removing the basic HTTP auth recently, but I wasn't going to ask @whew to have to fix conflicts again because we're slow! :) Pushed the fix myself and tested.

PR #1359, opened 2021-05-14
Prevent usernames in Chat mode of length 128 chars or more

### LINKED ISSUE #1291: Too long usernames in chat cause cookie / state corruption
You might want to consider setting a server-side limit on username setting. Too long usernames seem to both mess up the session cookie + JS messages and trigger a gui-side exception:

    /snap/onionshare/3/lib/python3.6/site-packages/werkzeug/wrappers/base_response.py:479: UserWarning: The "b'session'" cookie is too large: the value was 11290 bytes but the header required 26 extra bytes. The final size was 11316 bytes but the limit is 4093 bytes. Browsers may silently ignore cookies larger than this.
      samesite=samesite,
      File "/snap/onionshare/3/lib/python3.6/site-packages/onionshare/tab/tab.py", line 556, in timer_callback
        mode.handle_request_individual_file_started(event)
      File "/snap/onionshare/3/lib/python3.6/site-packages/onionshare/tab/mode/__init__.py", line 479, in handle_request_individual_file_started
        self.toggle_history.update_indicator(True)
    AttributeError: 'ChatMode' object has no attribute 'toggle_history'

mig5:
Fixes #1291 . 

< 128 chars seemed reasonable to me. 

Note that it evaluates it on the POST directly (which also allowed for a test!) as well as in Javascript.

I also added some text in the template to make it clear that the form field is for a username (if it's empty, it's not clear what it's for otherwise).

Annoyingly GUI tests are failing right now on some Jinja2 version issue that is surely unrelated. I think we need to lock Jinja2 to < 3.0.0

https://app.circleci.com/pipelines/github/mig5/onionshare/186/workflows/8e38f2aa-73e3-4cc1-b311-8e29d04f2dd7/jobs/466/parallel-runs/0/steps/0-103

### DIFF
--- a/cli/onionshare_cli/resources/static/js/chat.js
+++ b/cli/onionshare_cli/resources/static/js/chat.js
@@ -88,7 +88,7 @@ var emitMessage = function (socket) {
 
 var updateUsername = function (socket) {
   var username = $('#username').val();
-  if (!checkUsernameExists(username) && !checkUsernameLength(username)) {
+  if (!checkUsernameExists(username) && !checkUsernameTooLong(username)) {
     $.ajax({
       method: 'POST',
       url: `http://${document.domain}:${location.port}/update-session-username`,
@@ -133,7 +133,7 @@ var checkUsernameExists = function (username) {
   return false;
 }
 
-var checkUsernameLength = function (username) {
+var checkUsernameTooLong = function (username) {
   $('#username-error').text('');
   if (username.length > 128) {
     $('#username-error').text('Please choose a shorter username.');
--- a/cli/onionshare_cli/resources/static/js/chat.js
+++ b/cli/onionshare_cli/resources/static/js/chat.js
@@ -88,7 +88,7 @@ var emitMessage = function (socket) {
 
 var updateUsername = function (socket) {
   var username = $('#username').val();
-  if (!checkUsernameExists(username)) {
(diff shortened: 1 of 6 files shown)


cli/onionshare_cli/resources/templates/chat.html
```diff
@@ -23,6 +23,7 @@ <h1>{% if title %}{{ title }}{% else %}OnionShare Chat{% endif %}</h1>
   <div class="chat-container no-js">
     <div class="chat-users">
       <div class="editable-username">
+        <p>Your username:</p>
```
SaptakS, line 26: Can you change this to `<label for="username">` tag instead?

cli/onionshare_cli/resources/static/js/chat.js
```diff
@@ -133,6 +133,15 @@ var checkUsernameExists = function (username) {
   return false;
 }
 
+var checkUsernameLength = function (username) {
```
SaptakS, line 136: Maybe change the function name to `checkUsernameTooLong`? from the function name it's not very clear right now what it is doing.

SaptakS (changes requested):
Works perfectly for me! Left a few knitpick comments in the PR. After that, it should be ready for merging.

SaptakS (approved):
Looks good to me!

Issue #1347, opened 2021-05-06
Add scrollbar to settings widget because it gets cut off in macOS


micahflee:
With the default window size in macOS here's what receive mode looks like now:

<img width="1152" alt="Screen Shot 2021-05-05 at 8 07 57 PM" src="https://user-images.githubusercontent.com/156128/117236464-97239000-addd-11eb-9212-0c1ec4b85fd4.png">

And when I click "Show advanced settings" it totally cuts things off making it unusable:

<img width="1152" alt="Screen Shot 2021-05-05 at 8 08 22 PM" src="https://user-images.githubusercontent.com/156128/117236494-a571ac00-addd-11eb-94ab-19142db56cbb.png">

PR #1360, opened 2021-05-16
Adds scrolling on the common settings options

SaptakS:
This adds a scrolling on the common settings for every mod. Only issue right now is, like before, it doesn't expand to the available space. It just converts to scrolling instead of taking up the available space. One solution is probably having a minimumHeight with a stretch below, so that on clicking advanced options, it can take up the space.

@micahflee let me know what you think.

Refs #1347


SaptakS:
Closing this since it will be resolved in #1363

PR #1363, opened 2021-05-26
Options are all in a scroll widget and spacing looks good

### LINKED ISSUE #1347: Add scrollbar to settings widget because it gets cut off in macOS

micahflee:
Fixes #1347.

Think it makes sense to use this PR instead of #1360 @SaptakS?

This makes the settings for each mode all live in a scroll area, where a scrollbar only appears if the height of the window is too small to fit it all. Here's receive mode, both with and without the advanced settings showing:

Screenshot at 2021-05-25 17-21-39
Screenshot at 2021-05-25 17-21-52

And if you resize the window to be a bit bigger:

Screenshot at 2021-05-25 17-22-26

This is the same for all modes. It also adds 30px of empty space at the top of the chat mode, so that "Chat Anonymously" header lines up with the header in receive mode (because receive mode has a history toggle bar thing, chat mode doesn't).

Screenshot at 2021-05-25 17-23-07


SaptakS:
Makes sense to me. Can you also add https://github.com/micahflee/onionshare/pull/1360/files#diff-fea048fb94b49043087e4e404041ab0ac7d7faaabf564733decf06f843caa51aR75-R76 in this PR? That helps to prevent the texts getting cut off in macbook in file selection widget

PR #1379, opened 2021-06-07
Set the Max-Width of the Chat-Window to 80%

BotMaster3000:
Too long single-line messages can cause the ChatUser-Panel to disappear.
Allowing the windows of the Chat to be only 80% of width will cause a automatic linebreak in such a case.

### DIFF
--- a/cli/onionshare_cli/resources/static/css/style.css
+++ b/cli/onionshare_cli/resources/static/css/style.css
@@ -202,6 +202,7 @@ ul.breadcrumbs li a:link, ul.breadcrumbs li a:visited {
 }
 
 .chat-wrapper {
+  max-width: 80%;
   display: flex;
   flex-direction: column;
   flex: 1;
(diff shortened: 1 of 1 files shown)


SaptakS (changes requested):
Thanks for opening this PR!! I think the issue you reported was text without any space between them causes the UI to break. This PR though addresses the breaking of UI, but in case there is no space, it won't wrap words to next line and instead create a horizontal scroll. To tackle that, can you also add `word-break: break-word` to the CSS rule? That should handle the wrapping of very long text and avoid the horizontal scroll

BotMaster3000:
Hello SaptakS,
thanks for the response, I changed it, so that the message-class now has the word-break-Element, and I removed the max-width again to not cause future confusion, as it has become obsolete again.
Now both cases are correctly handled, neither long-single-words cause this issue nor too long sentences.

SaptakS (approved):
Looks good to me.

Issue #1377, opened 2021-06-06
Automatic check for updates should use onionshare.org, not GitHub


micahflee:
The code that checks for available updates in Windows and macOS checks GitHub releases to see what the latest release is, then links to the GitHub release page for that release when you click download update.

Instead it should check the onion site verison of https://onionshare.org/latest-version.txt, and link to https://onionshare.org/.

BotMaster3000:
When I am not misinterpreting the Code from the update_checker.py, then the current develop-branch already checks for the current version of OnionShare via the Tor-Adress of OnionShare. 
So the only thing remaining would be to redirect to onionshare.org instead of Github.

PR #1380, opened 2021-06-09
 #1377 The Update-URL now links to the official OnionShare-Website instead of the Github-Release-Page

BotMaster3000:
As issued by micahflee in the Issue #1377


micahflee (approved):
Thank you!

Issue #1382, opened 2021-06-22
Add dark mode/light mode setting


micahflee:
Right now it automatically detects if dark mode is enabled in the OS and if so uses it. But in some operating systems it doesn't detect this accurately. We should add a setting to allow the user to explicitly chose. The options can be Automatic, Light, or Dark, and it defaults to Automatic.

siddhantdixit:
Sir, I would like to work on this issue! Can u assign me?

siddhantdixit:
![image](https://user-images.githubusercontent.com/22856752/124632389-e9e2fc00-dea1-11eb-8bfc-c568461cd62b.png)

I have added separate option to choose theme in onionshare desktop.

SaptakS:
@siddhantdixit sure! Go ahead. As options for the theme, I will recommend there be 3 different options: light, dark and system. We already have it in a way that it works (in a bit flaky way) for the system option. So basically the implementation for system is there (where it uses a not so full proof way of detecting system color mode). And light and dark options should just ignore that detection and set it to either light or dark. Let me know if you need any assistance.

siddhantdixit:
Yeah! The default option doesn't work properly. I have enabled dark mode in Windows 10, program was not detecting "dark" mode.
We can add the theme options separately for System, Light and Dark. 

For the dark mode detection I have tested darkdetect on windows and linux it is working fine we can implement this for dark theme detection.
https://github.com/albertosottile/darkdetect

Thank you!

micahflee:
Fixed in #1386 1386

PR #1386, opened 2021-07-21
Added user theme preference option in Settings

siddhantdixit:
Added an option to choose theme in settings dialog like auto, light and dark.
Fixed Dark Mode Dark Text.


SaptakS:
For making the test pass, you will need to add the theme key in the settings test file in this area: https://github.com/micahflee/onionshare/blob/develop/cli/tests/test_cli_settings.py#L20-L37

siddhantdixit:
Thank You! I added the key here and ran scripts/rebuild-cli.py and it worked.

SaptakS:
@siddhantdixit I took a brief look at the PR, and the issue with it right now is, when it's automatic/system settings, the OS itself changes certain default colors, like background, etc. in most operating systems. But when you enforce a light or dark mode, the OS itself won't update the default colors, so need to update the entire color palette (I think?) via pyside to enforce the color mode. Right now it's just updating the colors that we have enforced, while leaving the default colors to the system's current color mode. Does that make sense?

siddhantdixit:
@SaptakS  As mentioned in Issue #1382 to add an option to choose theme explicitly currently it does that only. The theme is either light or dark only. 
When set to auto it detects the system color mode if it's set to dark mode then it enables it otherwise it will set it to light. 
I can't find if the OS automatically changes default color like background etc. I tested this on Windows 10 and Linux.
It changes only the explicit theme and colors already defined. We can improve the color layout of dark mode if needed as the current one does not feels like dark mode.

Dark Theme (Windows 10)
image

Dark Theme (Ubuntu Linux)
image


Light Theme (Windows 10)
image


Light Theme (Ubuntu Linux)
image

SaptakS:
@siddhantdixit that's what I meant. You have to explicitly mention the dark mode backgrounds and all explicitly. What I mean is, let's say at automatic it takes system setting. So if the entire system is dark mode, it will apply dark mode default colors to all applications, and if in light mode, the light mode system colors. But if the system is in light mode, and you chose dark mode theme explicitly, we need to manually update the default colors, like background for example so that it is dark. So yes, when we change theme, we need to set backgrounds and other colors explicitly. I would prefer if the color's for the light and dark mode follow the default light and dark colors that the system would apply. Right now the light is applied, so it would be great if you can find the dark pyside palette and apply that when Theme is switched to dark.

siddhantdixit:
@SaptakS  I have added the fusion style palette (dark mode). The current detection of dark mode doesn't work properly (auto) so by default it sets to light. So colors in history tab are left. I will try to fix them ASAP. Is this theme fine for this?

I have tested darkdetect on Linux and Windows and it detects system color mode properly. I suggest adding this to project.
https://github.com/albertosottile/darkdetect
Is this fine here? 🤔

cli/tests/test_cli_settings.py
```diff
@@ -34,6 +34,7 @@ def test_init(self, settings_obj):
             "tor_bridges_use_meek_lite_azure": False,
             "tor_bridges_use_custom_bridges": "",
             "persistent_tabs": [],
+            "theme":0
```
SaptakS, line 37: Maybe change 0 to "auto"

desktop/src/onionshare/__init__.py
```diff
@@ -65,9 +74,38 @@ def is_dark_mode(self):
...
+        current_theme = curr_settings.get("theme")
+
+        if current_theme == 0:
+            return "dark" if self.is_dark_mode() else "light"
+        elif current_theme == 1:
+            return "light"
+        elif current_theme == 2:
+            return "dark"
+        else:
+            return "light"
```
SaptakS, line 108: Let's use a logic like:
```python
if current_theme == 1:
    return "light"
elif current_theme == 2:
    return "dark"
else
    return "dark" if self.is_dark_mode() else "light"
```

SaptakS:
@siddhantdixit this looks good to me! I think the main idea of detection was primarily for macOS. We saw that it works sometimes, but other times it doesn't in other OS, which I think is one of the main reasons to allow users to chose which color mode. I will have to check that it works cross platform following the usual dark mode styling, but else also I think this looks pretty good.

As of darkdetect, I clearly remember checking it and not deciding to go with it, though I can't remember why.  🤔  One reason is didn't want to introduce external dependencies, but don't know if there was some reason. I am not against using it. @micahflee any thoughts?

siddhantdixit:
@SaptakS I have checked darkdetect, for detecting color mode it checks registry values in Windows and a terminal command as a python subprocess in Linux. I'm not sure about Mac. But they only use python std with small functions for this task. I feel it will work properly.

SaptakS:
@siddhantdixit let's make a separate PR for the darkdetect change and let's review that separately from this PR. That way it won't be blocking the merging of this PR.

SaptakS (commented):
Tested it in different OS. Looks good to me! Only the history part is still in light mode and need some working. @siddhantdixit did you get a chance to look into that?

SaptakS (approved):
Looks good to me

micahflee (approved):
This is great, thanks!

Issue #1365, opened 2021-05-27
CLI crashes with AssertionError on Ubuntu 20.04 ARM64


Felixkruemel:
I can't get onionshare to work.
I've installed it in a python3-venv using `python -m venv venv` and activated the venv and then installed it using `pip install onionshare-cli`

The problem is that the help menu works with `onionshare-cli --help` but the actual sharing of e.g. a folder crashes with `AssertionError: View function mapping is overwriting an existing endpoint function: static`

I've got an natively running ARM64 Server running on Ubuntu 20.04. Also I want to point out that there are no snap packages for arm64 available like on many other projects. Maybe we could build them too.

Felixkruemel:
I now also tried the latest commit / dev build using the `setup.py` which still results in the same error. So it's not the pip build which is wrong ^^

esskayesss:
Same on Archlinux `5.12.4-arch1-2`

Felixkruemel:
@micahflee If you can't reproduce the error I'm open to help in regards of that. It's really crap that it won't launch at all :(

whiskey01:
I have the same issue running Linux 5.12.2-arch1-1.

whiskey01:
Here is the complete error:
`OnionShare 2.2 | https://onionshare.org/
Traceback (most recent call last):
  File "/usr/bin/onionshare", line 22, in <module>
    onionshare.main()
  File "/usr/lib/python3.9/site-packages/onionshare/__init__.py", line 181, in main
    web = Web(common, False, mode)
  File "/usr/lib/python3.9/site-packages/onionshare/web/web.py", line 77, in __init__
    self.generate_static_url_path()
  File "/usr/lib/python3.9/site-packages/onionshare/web/web.py", line 166, in generate_static_url_path
    self.app.add_url_rule(
  File "/usr/lib/python3.9/site-packages/flask/scaffold.py", line 56, in wrapper_func
    return f(self, *args, **kwargs)
  File "/usr/lib/python3.9/site-packages/flask/app.py", line 1092, in add_url_rule
    raise AssertionError(
AssertionError: View function mapping is overwriting an existing endpoint function: static`

kpcyrd:
This issue is likely due to missing flask 2.0 support in onionshare.

Danie10:
Glad I'm not the only one - identical error on Manjaro KDE 5.13.4-1

kpcyrd:
See #1385

whiskey01:
[...]

PR #1392, opened 2021-08-19
Fix issue #1365

twann4:
By browing the code, I saw that the endpoint "static" is what causes the `AssertionError`, but that this endpoint seems not used anywhere else. So I simply changed it to "onionshare-static" and it seems to work.
Not sure if this is totally safe for now, but I ran all the tests on my computer (Artix Linux x86_64 5.10.56-1-lts) and everything was successful, I didn't even get a warning for that line.


micahflee (approved):
I haven't reproduced the actual bug myself yet, but I have tested this and it works great, and in fact that `self.app.add_url_rule` call doesn't even need an `endpoint` passed in at all. I'll go ahead and approve this, then add a commit to delete that line.

PR #1402, opened 2021-08-24
Don't allow an upload if in non-public mode and there was no Authorization header in the request

### LINKED ISSUE #1396: File uploaded before checking for user's authentication
**Version:** OnionShare-cli v2.3.3 (installed via pip3)
**Start-up:** onionshare-cli --receive

**Observed behavior:**
Unauthenticated users (not passing the _Authorization: Basic_ header) are still able to upload files on the remote machine running OnionShare v2.3.3

This is probably caused by the logic of https://github.com/onionshare/onionshare/blob/develop/cli/onionshare_cli/web/receive_mode.py, in which files are uploaded and stored remotely **before** checking for user authentication.

This issue is affecting both:
* POST /upload
* POST /upload-ajax

Proof of concept images attached

mig5:
Fixes #1396 

I am still not clear on *why* the subclassed Request middleware is triggered before the HTTPAuth's `app.before_request()` decorator is applied to the request, but in any case, this should fix the issue.

I expanded the test to not only check for the 401 response but also that an uploaded message should not exist on disk.

### DIFF
--- a/cli/onionshare_cli/web/receive_mode.py
+++ b/cli/onionshare_cli/web/receive_mode.py
@@ -372,6 +372,9 @@ class ReceiveModeRequest(Request):
         if self.method == "POST":
             if self.path == "/upload" or self.path == "/upload-ajax":
                 self.upload_request = True
+                # If using a password, we should abort if there was no Authorization header sent
+                if not self.web.settings.get("general", "public") and not "Authorization" in self.headers:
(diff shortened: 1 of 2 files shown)


SaptakS (approved):
LGTM

SaptakS:
I think this can be closed since #1404 got merged.

PR #1404, opened 2021-08-27
Support ClientAuth (a.k.a 'stealth') on v3 onions with Tor 0.4.6.x

### LINKED ISSUE #697: Add support for v3 onion service client authorization
A large amount of progress has been made on the v3 onion service client authorization development, most of the code is all ready done. Adding support for v3 onion service client authorization would Improve security for end users. Look at this pull request for the v3 onion service client authorization code https://github.com/torproject/tor/pull/36 Here is the repo for the code in pull 36 https://github.com/haxxpop/tor/tree/ticket20700_01

[asn](https://trac.torproject.org/projects/tor/ticket/20700#comment:26)

> We are hoping to land this in 035.

**And the current Trac ticket tracking v3 onion service client authorization is** https://trac.torproject.org/projects/tor/ticket/20700

**Also stealth client authorization is a feature Torproject is working on here.**
https://trac.torproject.org/projects/tor/ticket/20742
@mig5 @micahflee

### LINKED ISSUE #1112: Drop support for v2 onion services
I wonder if we should drop support for v2 onion services altogether. They're not as secure as v3 onion services, and Tor Browser has supported v3 for a long time now. This could simplify the onion logic of OnionShare a lot, and remove the pycryptodome dependency (it's only used to detect if a key is v2 or v3).

The only benefit that v2 has is that the addresses are shorter, so they're easier to manually type.

### LINKED ISSUE #1343: Replace basic authentication with cookie-based authentication
Browsers are moving to completely deprecate HTTP basic authentication: https://nakedsecurity.sophos.com/2020/04/14/microsoft-and-google-delay-online-authentication-change/

It's unclear when Tor Browser will deprecate basic auth, but we might as well get ahead of it sooner.

OnionShare uses basic auth for non-public services, and the username and password are included in the URL, like this: `http://onionshare:[password]@[something].onion`

Instead, we can make non-public OnionShare services require a password and build a very simple session into the web app:

- If you load any page and you're not logged in, it can show a simple page asking for a password
- The password rules can be the same, 20 invalid guesses and the service automatically stops
- If you type the correct password, it sets a session cookie and knows that you're logged in, and then all requests should just work
- We can make an auto-login URL so that there's still a single thing to copy and paste, in the format `http://[something].onion/auth/[password]` -- if you load that URL, it will try that password and if it works, log you in and redirect to `/`, and otherwise display the password field
- Just because, we might as well have a `/auth/logout` that deletes the session

This will let us completely remote basic auth from OnionShare, and we can also drop the `flask-httpauth` dependency.

### LINKED ISSUE #1396: File uploaded before checking for user's authentication

### LINKED ISSUE #1397: Disclosure of Chat Participants to Unauthenticated Users
**Version:** OnionShare-cli v2.3.3 (installed via pip3)
**Start-up:** onionshare-cli --chat

**Observed behavior:**
It was possible, from an unauthenticated perspective, to retrieve the full list of participants in the target OnionShare chat.

This is caused by https://github.com/onionshare/onionshare/blob/02254b13bb4818745193092f2144fd83726d79e7/cli/onionshare_cli/web/chat_mode.py#L128 which discloses the connected_users list to anyone via websocket (with valid or invalid flesk session cookie).

It seems that without a valid session ID it was not possible to intercept messages between users, since the system heavily relays on the session to connect into the default room - and without a valid one, messages remain undelivered to unauthenticated users.

It is however recommended to avoid initiating a socket.io connection without prior validating the 'session' cookie.

**Proof of Concept:**
'Origin' request header should be tampered (e.g. via Burp Suite) to use the v3 onion address - otherwise the request will fail.

```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>
    </title>
  </head>
  <body>
    <h1>OnionShare Disclosure of Connected Users PoC
    </h1>
    <ul id="user-list">
    </ul>
    <script src="https://cdn.socket.io/3.1.1/socket.io.min.js" crossorigin="anonymous">
    </script>
    <script src="https://code.jquery.com/jquery-3.5.1.min.js">
    </script>
    <script> 
      $(function () {
        $(document).ready(function () {
          var socket = io.connect(
            'http://2vw6cdagn4fy2olj4mpbt6wg5vthdb4zu3odiqcluycfauvpaa7aqjid.onion/chat',
            {
              transports: ['websocket']
            }
          );
          socket.on('connect', function () {
            socket.emit('joined', {
            }
                       );
          }
                   );
          socket.on('status', function (data) {
            var userListHTML = '';
            var userslist = data.connected_users;
            for (i = 0; i < userslist.length; i++) {
              userListHTML += `<li>${userslist[i]}</li>`;
            }
            $('#user-list').html(userListHTML);
          }
                   );
        }
                         );
      }
       );
    </script>
  </body>
</html>
```

mig5:
Closes #697
Closes #1112
Closes #1343
Closes #1396
Closes #1397

Good news! We forked upstream's Stem so that we could fix some issues in the `setup.py` to make it [compatible with Poetry](https://stackoverflow.com/questions/68946064/how-do-i-make-stem-install-from-a-specific-git-branch-with-poetry/68946456#68946456). 

This means I can finally open my PR and have people test Client Auth (a.k.a stealth) with v3 onions!

Please note: Client Auth for v3 requires Tor 0.4.6.x on the *OnionShare* side. This is the major version that supports requesting ClientAuthV3 flag via the Tor control port when registering the onion service. It won't work on 0.4.5.x. 

However, the other person running Tor Browser to access your onion service can be using Tor 0.4.5.x just fine, e.g (in the latest Tor Browser, which uses Tor 0.4.5.10).

On Debian 10, I was able to install Tor 0.4.6.7 with this as my apt torproject.list file:

```
deb https://deb.torproject.org/torproject.org buster main
deb-src https://deb.torproject.org/torproject.org buster main
deb https://deb.torproject.org/torproject.org tor-experimental-0.4.6.x-buster main
deb-src https://deb.torproject.org/torproject.org tor-experimental-0.4.6.x-buster main
```

Given that the latest version of Tor Browser still uses 0.4.5.10 at time of writing, so using TB as the Tor connection won't help you either. I guess this is less of an issue for our flatpak/snap packages which can build a newer version from source..

~~Please note: I haven't removed the 'basic auth' a.k.a 'use a password'. But I really if we should just do that, although it'll not be a backwards-compatible change e.g for persistent onion shares..~~

I also think that ClientAuth kind of makes #1343 unnecessary? We don't need to manage cookies instead of a password, if we have ClientAuth.. what do you think?

Turning on Client Auth is a setting in the share options in each tab ('Use client authorization'). When active, there'll be a button on an active share 'Copy ClientAuth' which contains the private key that is needed for the other user to paste into their Tor Browser dialog when prompted for it (or for configuring manually in your tor settings, for power users..)

Happy testing!


micahflee:
Also:

- [ ] Remove the `flask_httpauth` dependency, and rip out the HTTPAuth code from https://github.com/onionshare/onionshare/blob/develop/cli/onionshare_cli/web/web.py

micahflee:
- [ ] The Web object doesn't need the `REQUEST_INVALID_PASSWORD` request type either now, or `REQUEST_RATE_LIMIT`

mig5:
>     The Web object doesn't need the REQUEST_INVALID_PASSWORD request type either now, or REQUEST_RATE_LIMIT

>     Remove the flask_httpauth dependency, and rip out the HTTPAuth code from https://github.com/onionshare/onionshare/blob/develop/cli/onionshare_cli/web/web.py

I think I already removed these in this PR, in commit https://github.com/onionshare/onionshare/pull/1404/commits/0bf8f53d30ded17dde0b3ebf66d98ea7d8e5313d

mig5:
@micahflee right, I think all that is now addressed in the latest commits.

(I still don't know why my fork hides the Circle CI tests in PRs, but you can see the tests are passing [here](https://app.circleci.com/pipelines/github/mig5/onionshare/206/workflows/cd8add47-b69a-4671-932b-802dbecefa8f/jobs/508/parallel-runs/0/steps/0-103)).

One thing I haven't done, that maybe you can tack on as a commit: the flatpak/snap yaml files will need the `pynacl` dependency, and probably the change to the Stem source location too to be from our git repo and 1.8.1 tag. I've not really done flatpak/snap stuff before so I wasn't sure if that was something you do manually or if you have some script that populates it based on the pyproject.toml or something.

I'll start working on docs but maybe that can be a separate PR as this one's getting pretty big.

BTW, perhaps also it's worth doing a 'squash' merge in case we need to revert this whole PR for some reason later..

mig5:
I pushed some documentation updates in https://github.com/onionshare/onionshare/pull/1404/commits/bca5bee2098915a94a469a8fa0532d74ad63604f after all. But there are probably screenshots that need updating too, which maybe you want to do later since yours tend to look smarter than mine on Qubes!

micahflee (approved):

micahflee:
This is great! I will work on a thorough edit of the documentation before we make the actual release, including taking new screenshots, but I think this is a great start.

And I don't think we'll want to revert this whole PR later for some reason. If we did decide to not rely on v3 client auth, it might make more sense to instead implement #1343, so I'm okay with not squashing merging.

PR #1406, opened 2021-08-28
Fixes directory listing in send mode

### LINKED ISSUE #1405: Directory listing links are broken
To reproduce, share the onionshare codebase, e.g.:

```
$ poetry run onionshare-cli --public --no-autostop-sharing .
```

Screenshot from 2021-08-27 16-43-19

The top-level links work fine, like clicking `onionshare_cli` goes to:
`http://3i4arxbysttlbt7phxru3ahhewchaslt2nxkimpsj6ygvzfsn23mdcqd.onion/onionshare_cli`

Screenshot from 2021-08-27 16-44-28

But these links are all broken. Clicking `onion.py` goes to here, which is a 404:
`http://3i4arxbysttlbt7phxru3ahhewchaslt2nxkimpsj6ygvzfsn23mdcqd.onion/onion.py`

It should go to here, which works:
`http://3i4arxbysttlbt7phxru3ahhewchaslt2nxkimpsj6ygvzfsn23mdcqd.onion/onionshare_cli/onion.py`

I seem to remember having this bug and fixing it in the past, but it's back...

SaptakS:
Fixes #1405 

Seems the changes were made in `listing.html` but send mode uses `send.html`. Just putting the basename doesn't actually work as a relative path. It's better to put the entire relative path from the base. The paths to the directory don't end with a trailing slash, hence the basename is always considered as root level by browsers. Hence providing the relative path from the root level is best and safe.

### DIFF
--- a/cli/onionshare_cli/web/share_mode.py
+++ b/cli/onionshare_cli/web/share_mode.py
@@ -425,10 +425,7 @@ class ShareModeWeb(SendBaseModeWeb):
                 # Render directory listing
                 filenames = []
                 for filename in os.listdir(filesystem_path):
-                    if os.path.isdir(os.path.join(filesystem_path, filename)):
-                        filenames.append(filename + "/")
-                    else:
-                        filenames.append(filename)
+                    filenames.append(filename)
                 filenames.sort()
                 return self.directory_listing(filenames, path, filesystem_path)
 
--- a/cli/onionshare_cli/resources/templates/send.html
+++ b/cli/onionshare_cli/resources/templates/send.html
@@ -40,7 +40,7 @@
     <div class="d-flex">
       <div>
         <img width="30" height="30" title="" alt="" src="{{ static_url_path }}/img/web_folder.png" />
-        <a href="{{ info.basename }}">
+        <a href="{{ info.link }}">
           <span>{{ info.basename }}</span>
         </a>
       </div>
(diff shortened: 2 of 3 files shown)


SaptakS:
Tests: https://app.circleci.com/pipelines/github/SaptakS/onionshare/42/workflows/fa4f56f6-e967-4d4a-84f5-e4aac575fc82

micahflee (changes requested):
Thank you, this nearly fixes it, but it looks like `info.link` is wrong for deeper subdirs. If we try sharing the `onionshare/cli` folder:

```sh
poetry run onionshare-cli --public --no-autostop-sharing .
```

- Then load the onion address
- Click `onionshare_cli`,  which loads `/onionshare_cli`
- Click `resources/`, which loads `/onionshare_cli/resources/` and is a 404, because of the trailing slash

It looks like the problem is directories get a trailing slash in the filename at some point, for some reason, when they shouldn't.

SaptakS:
Interesting. Let me dig in a bit more to see why the first directory listing isn't getting a trailing slash, while others are trying to.

cli/onionshare_cli/web/share_mode.py
```diff
@@ -425,10 +425,7 @@ def render_logic(self, path=""):
                 # Render directory listing
                 filenames = []
                 for filename in os.listdir(filesystem_path):
-                    if os.path.isdir(os.path.join(filesystem_path, filename)):
-                        filenames.append(filename + "/")
-                    else:
```
SaptakS, line 430: I am not sure why the trailing slashes were added in the first place since adding trailing space produces wrong link which always gives a 404 (at least not). And the logic was such that trailing slash where added only to directories inside a directory, and not root level. Not sure what changed, but this seems to fix the issue.

SaptakS:
@micahflee fixed the above mentioned bug

micahflee (approved):

PR #1411, opened 2021-08-30
Cli installation instructions

nyxnor:
Just so the guide contains everything.
Shoud docs.onionshare.org include this instructions or refet to them here https://docs.onionshare.org/2.3.3/en/install.html ?

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -57,7 +57,7 @@ Install OnionShare CLI:
 pip install --user onionshare-cli
 ```
 
-### Set path
+#### Set path
 
 When you install programs with pip and use the --user flag, it installs them into ~/.local/bin, which isn't in your path by default. To add ~/.local/bin to your path automatically for the next time you reopen the terminal or source your shell configuration file.
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -59,7 +59,7 @@ pip install --user onionshare-cli
 
 #### Set path
 
-When you install programs with pip and use the --user flag, it installs them into ~/.local/bin, which isn't in your path by default. To add ~/.local/bin to your path automatically for the next time you reopen the terminal or source your shell configuration file, , do the following:
+When you install programs with pip and use the --user flag, it installs them into ~/.local/bin, which isn't in your path by default. To add ~/.local/bin to your path automatically for the next time you reopen the terminal or source your shell configuration file, do the following:
 
 First, discover what shell you are using:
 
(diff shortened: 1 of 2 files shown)


cli/README.md
```diff
@@ -22,14 +22,69 @@
...
+
+Install OnionShare CLI:
+
+```sh
+pip install --user onionshare-cli
+```
+
+#### Set path
+
+When you install programs with pip and use the --user flag, it installs them into ~/.local/bin, which isn't in your path by default. To add ~/.local/bin to your path automatically for the next time you reopen the terminal or source your shell configuration file.
```
mig5, line 62: Can we add at the end of this sentence, `, do the following:`, so that it reads like this?

> To add ~/.local/bin to your path automatically for the next time you reopen the terminal or source your shell configuration file, do the following:

cli/README.md
```diff
@@ -22,14 +22,69 @@
...
+
+Fist, discover what is your shell:
```
mig5, line 64: `First`, not `Fist` :)

mig5: Can we make it:

> First, discover what shell you are using:

mig5 (changes requested):
Just some small wording changes, but it looks pretty good!

Note: the CLI/README.md is not shown at https://docs.onionshare.org. You originally wanted CLI install instructions to be on the docs site, so if you want, you could look at adding some instructions or a link to this README, in this file: https://github.com/onionshare/onionshare/blob/develop/docs/source/install.rst

It could be as simple as 'If you want to install OnionShare just for the command line, please see the README in the Git repository' with that README being a link to this file you've changed here.

Thanks!

cli/README.md
nyxnor: done

cli/README.md
nyxnor: done

cli/README.md
```diff
@@ -59,9 +59,9 @@ pip install --user onionshare-cli
 
 #### Set path
 
-When you install programs with pip and use the --user flag, it installs them into ~/.local/bin, which isn't in your path by default. To add ~/.local/bin to your path automatically for the next time you reopen the terminal or source your shell configuration file.
+When you install programs with pip and use the --user flag, it installs them into ~/.local/bin, which isn't in your path by default. To add ~/.local/bin to your path automatically for the next time you reopen the terminal or source your shell configuration file, , do the following:
```
mig5, line 62: @nyxnor are you able to remove the duplicate ` ,` here? Thanks!

nyxnor: sorry...

mig5 (approved):

micahflee (approved):

PR #1412, opened 2021-08-31
Client Auth UX improvements

### LINKED ISSUE #1409: Update v3 client auth UX
@glenn-sorrentino did some work on improving the v3 client auth UX.

Active + Activity Panel

I think we might want to change the wording a bit still, but I like the idea of splitting the address/private key into two numbered steps, and having separate QR code buttons that just show one QR code at a time.

Maybe, if public mode is disabled, it can say:

> 1. **Anyone** with this OnionShare address and private key can **(download, upload, chat, etc.)** using the **Tor Browser**
> `http://shgjd4amgl2xwa2fdmw4zztkwzuk3tsqhkolu6sn4mipv6iatq3xuaad.onion`
> [Copy Address] [Show QR Code]
> 
> 2. They will need to enter this **private key** to access your OnionShare service
> [Copy Private Key] [Show QR Code]

If public mode is enabled:

> **Anyone** with this OnionShare address can **(download, upload, chat, etc.)** using the **Tor Browser**
> `http://shgjd4amgl2xwa2fdmw4zztkwzuk3tsqhkolu6sn4mipv6iatq3xuaad.onion`
> [Copy Address] [Show QR Code]

mig5:
Closes #1409

clientauth_v3_1

I also initially added the Private Key as a white-background label like the OnionShare address itself, for consistency, as I was worried users might be confused as to why they can't 'see' the private key unless they click the Copy Private key button or use the QR code:

onionshare-clientauth-v3-ux-improvements

In the era of v3 onions, the address is arguably just as 'secret' as the private key.

But, maybe a Private Key will be seen as more sensitive to 'display' automatically to the user, in terms of shoulder-surfing concerns etc, and seeing the URL without the key is a mitigation against seeing both to gain access.. so I decided to keep it removed after all.

I'm not 100% happy because we have padding problems again, particularly in Receive Mode where there is a lot of text compared to other modes. I had to raise the minimumHeight of the main window which I am generally loathe to do, otherwise when both the URL and the Client Auth stuff is shown, it squashes the URL bar massively and you can't see the URL. Raising the minimumHeight from 700 has resolved that, but maybe someone can help with a more elegant fix.


SaptakS:
I will take a look at it for the padding stuff. I think one solution for "showing" private key would be that we show it in password mode, and then have an icon which toggle visibility.

micahflee (approved):
This is great! I've made a few commits which I'm going to include when I merge.

- I reverted your "Don't show private key" commit, and then I made it `*`'d out by default with a reveal/hide button
- I fixed the padding by putting the image in the receive mode in the right column, and set the window max height back down to 700

Screenshot from 2021-09-03 14-28-16
Screenshot from 2021-09-03 14-28-23
Screenshot from 2021-09-03 14-28-40

It's a little scrunched when the history widget is showing, but I think it's still fine:

Screenshot from 2021-09-03 14-28-55

PR #1415, opened 2021-09-02
Fixes typos about Linux as platform name

kushaldas:
It is `linux` not `Linux` in sys.platform. Now 4 more tests are running on Linux and not skipped.


SaptakS (approved):
Looks good to me

PR #1416, opened 2021-09-02
Adds exception for ConnectionError in chat mode during shutdown

### LINKED ISSUE #1319: Chat service silently prevents tor from shutting down
While testing the new onionshare release I noticed that quitting the application closes the gui but doesn't shutdown the tor process and leaves this behind, keeping the user connected to the tor network until the computer is restarted or the user manually kills the process with htop:
```
 370480 user       20   0 1269M  137M 65160 S  0.8  1.8  0:02.66 │  │  └─ /usr/bin/python /usr/bin/onionshare
 370616 user       20   0 41636 37460 16248 S  0.0  0.5  0:01.26 │  │     └─ /usr/bin/tor -f /home/user/.config/onionshare/tor_data/torrc
```
Steps to reproduce (tested on Arch Linux):
1) Start onionshare
2) Start the chat server (actually connecting with a browser isn't necessary)
3) Wait until the chat service has been fully started
4) Stop the chat service again
5) Quit onionshare by right clicking the tray icon and selecting "Quit"

Since I've started onionshare from the terminal I was able to press ^C, which causes the tor process to quit but the onionshare process seems to hang instead of reaping the child:
```
370480 user       20   0 1253M  137M 65420 S  0.0  1.8  0:03.01 │  │  └─ /usr/bin/python /usr/bin/onionshare
370616 user       20   0     0     0     0 Z  0.0  0.0  0:01.46 │  │     └─ /usr/bin/tor -f /home/user/.config/onionshare/tor_data/torrc
```

I've tried to reproduce this with regular file sharing and onionshare was shutting down correctly in that case, so I assume it's related to the websocket server.

SaptakS:
The way flask-socketio stops a connection when running using
eventlet is by raising SystemExit to abort all the processes.
Hence the connections are closed and no response is returned
So I am just catching the ConnectionError to check if it was
chat mode, in which case it's okay.

Fixes #1319

### DIFF
--- a/cli/onionshare_cli/web/web.py
+++ b/cli/onionshare_cli/web/web.py
@@ -372,9 +372,18 @@ class Web:
         # To stop flask, load http://shutdown:[shutdown_password]@127.0.0.1/[shutdown_password]/shutdown
         # (We're putting the shutdown_password in the path as well to make routing simpler)
         if self.running:
-            requests.get(
-                f"http://127.0.0.1:{port}/{self.shutdown_password}/shutdown"
-            )
+            try:
+                requests.get(
+                    f"http://127.0.0.1:{port}/{self.shutdown_password}/shutdown"
+                )
+            except requests.exceptions.ConnectionError as e:
+                # The way flask-socketio stops a connection when running using
+                # eventlet is by raising SystemExit to abort all the processes.
+                # Hence the connections are closed and no response is returned
+                # to the above request. So I am just catching the ConnectionError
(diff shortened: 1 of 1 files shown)


micahflee (approved):
I'm having trouble reproducing the original bug, but this patch looks fine to me. Have you reproduced it and you're confident it fixes it? If so I can merge it.

micahflee:
Actually I take that back. I just reproduced it and this fix seems to work great.

PR #1432, opened 2021-09-11
POSIX compliant input path on any shell

nyxnor:
`source is not posix compliant but '.' works.
Goal is less lines to read and one command that fits any shell.


micahflee (approved):
Nice

PR #1438, opened 2021-10-01
Catch KeyError on unsupported locale and fall back to en properly

### LINKED ISSUE #1437: OnionShare fails to start due to Apparmor DENIED in Ubuntu 21.04
Version: 2.4, installed via snap
OS: Ubuntu 20.04.3

Typical Behavior: OnionShare launches normally (as of previous recent version, uncertain of exactly which. Last used OnionShare successfully around 2 or 3 months ago).

Current Behavior (as of v2.4): When lauching OnionShare from GUI, process launches, hangs, then shuts down without displaying a window. When attempting to launch from CLI, the following error message is displayed:

```
qt.qpa.xcb: could not connect to display :0
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.

Aborted (core dumped)
```

Attempting to uninstall/reinstall, did not affect issue. Attempted system reboot to shake things loose, but also did not affect issue. OnionShare currently does not function.

I know of one other person who had a similar issue (which led me to try reproducing it). I was unable to find out which OS they were using at the time, but I believe it to be Windows based on screenshots. Their issue was similar to mine (unable to launch program) but the content of their error message was slightly different:

```
pythonw

This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem
```

mig5:
See #1437

This fixes installs that were running a locale (set in their existing `onionshare.json` settings file) that was supported in an earlier OnionShare version, but that has been dropped in the 2.4 release as part of https://github.com/onionshare/onionshare/commit/5de48bba6e3c17b0b45516fe536a3c500fab4fc4#diff-bad479f8b6a5c87b58ab30f427dd636f1af6461b7586362e672cddf4ede39f4eR57-R92

Not sure yet if it fixes #1437 altogether, which started out as a snap issue. But I reproduced the KeyError on macOS and Linux if using a locale we don't have in 2.4.

### DIFF
--- a/desktop/src/onionshare/strings.py
+++ b/desktop/src/onionshare/strings.py
@@ -43,9 +43,10 @@ def load_strings(common, locale_dir):
     current_locale = common.settings.get("locale")
     strings = {}
     for s in translations[default_locale]:
-        if s in translations[current_locale] and translations[current_locale][s] != "":
+        try:
             strings[s] = translations[current_locale][s]
-        else:
+        except KeyError:
+            common.log("GuiCommon", "load_strings", f"Couldn't load locale {current_locale}, falling back to {default_locale}")
             strings[s] = translations[default_locale][s]
 
 


ChanoSan:
I'm not familiar with the usual process re: localization, and I'm certainly in no position to make requests, but since I got JA back up to 100% would it be possible to petition for its re-inclusion in the next updated version?

mig5:
Yep, we'll do a 2.4.1 release for this fix, and generally any translations that are above 90% will get included.

Did modifying the locale manually fix your snap issue? And BTW, on Windows that `onionshare.json` settings file would be in something like `\Users\App_data\OnionShare`  if you wanted to test modifying the locale on 2.4 back to `en`..

ChanoSan:
> Did modifying the locale manually fix your snap issue? And BTW, on Windows that `onionshare.json` settings file would be in something like `C:\Users\$USER\AppData\OnionShare` if you wanted to test modifying the locale on 2.4 back to `en`..

Don't have access to a Windows environment, but I manually changed the locale in the onionshare.json in /.config/onionshare
Unfortunately, it didn't change anything... same error resulted. I'm not sure if a reboot is necessary or if fresh install might be different.

desktop/src/onionshare/strings.py
```diff
@@ -43,9 +43,10 @@ def load_strings(common, locale_dir):
     current_locale = common.settings.get("locale")
     strings = {}
     for s in translations[default_locale]:
-        if s in translations[current_locale] and translations[current_locale][s] != "":
+        try:
```
SaptakS, line 46: I feel we should put the if inside try anyways. At least `if translations[current_locale][s] != ""`. Else it might show up a blank translation in case we have a blank translation 🤔

ChanoSan:
In an effort to avoid clogging up the comments here with unrelated bughunting, I'm leaving further comments back in #1437

saminjon (commented):
Привет

desktop/src/onionshare/strings.py
saminjon: Какдела

Issue #1420, opened 2021-09-08
Implementation of "Enable Quickstart" window


SaptakS:
Once a user opens the OnionShare app, they will be greeted with an option to Enable Quickstart. If they select Enable Quickstart, from the next time, OnionShare tries to automatically connect the user by detecting any censorship. If not selected, they are always greeted with this screen along with a connect button and configure button.

Things to do:
- [x] basic description of the above story
- [x] A checkbox with "Enable Quickstart"
- [x] A "Start OnionShare" primary button
- [x] A "Configure" secondary button which just opens the current settings window
- [x] This screen appears before the current main screen and also before Onionshare tries to get tor started

refs: https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/29590#note_2749371

SaptakS:
Implemented in #1439

PR #1439, opened 2021-10-01
Implementation of the enable quickstart screen

SaptakS:
Refs #1420 

This is the initial structure:

Screenshot from 2021-10-02 02-19-39

Workflow right now:
- On starting OnionShare desktop, the above screen is shown
- If you click on "Start Onionshare" without clicking the checkbox to enable auto connect, it will start the tor connection and then show the new tab widget like it does right now. Then when you close and restart OnionShare again, it will still show the above screen.
- If you click the checkbox to enable auto connect, and start OnionShare, then from next time when you start OnionShare, it will start connecting to tor automatically (like it happens currently), without showing this screen again.
- If you click on configure, it opens the settings dialogue that we have right now.

Things needed to be done:
- [x] On opening settings dialogue, if we click on "Save", it tries to connect. With this flow, the wording is bit confusing I think. Maybe "Connect" makes more sense?
- [x] Need to resize the buttons to have same height
- [x] Adjust spacing between areas 
- [x] Should there be a setting to allow users to disable auto connect?
- [x] @micahflee what should be the wording for the individual sections? ~~Need to move them to strings instead of hardcoded right now~~
- [x] Once the above are done, @glenn-sorrentino some input from you to improve the designs.


micahflee:
Very cool! Here are my initial thoughts:

- [ ] I think we can just copy (mostly) the text from Tor Browser.
  - [ ] The description text should maybe be: "OnionShare relies on the Tor Network, run by thousands of volunteers around the world."
  - [ ] Then while it's connecting, "Connecting to Tor: [tor logs lines]"
  - [ ] "OnionShare has failed to establish a connection"
  - [ ] "OnionShare has failed to determine your location"
  - [ ] "OnionShare has failed to connect using a bridge"
- [ ] Right now it connects to Tor in a separate dialog, but I think having the tor logs during the connection appear directly in the description text, like it does in Tor Browser, would be nicerhere
- [ ] The settings button in the corner shouldn't be there yet, not until after Tor is connected
- [ ] The "Start OnionShare button" should say: "Connect to Tor"
- [ ] The auto-connect checkbox should say "Enable automatically connection to Tor" instead of "to OnionShare" I think
- [ ] When you click Configure it opens settings, but if you click Cancel it pops up a "Could not connect to Tor" dialog and quits that app. Instead, it should just close the settings and go back to the previous window.
- [ ] Yes, I think there should be a setting to toggle on and off automatically connecting
- [ ] In the Settings dialog, I agree that the wording "Save" is a bit confusion, and "Connect" makes more sense, but there are a few non-Tor settings here too. What if the "Save" button says "Connect to Tor and Save Settings" -- maybe that's too long though, maybe 'Connect and Save"?

glenn-sorrentino:
exploration using animation in startup
onionshare-start

SaptakS:
Thanks for the feedbacks! @glenn-sorrentino designs look great! Not sure how much of that toggle slider and animation i can implement in pyside, but let me check.

@micahflee those inputs are helpful, though some of it I may implement in a separate PR when I implement the flow of location detection and connecting to bridge UI, but rest is helpful.

> When you click Configure it opens settings, but if you click Cancel it pops up a "Could not connect to Tor" dialog and quits that app. Instead, it should just close the settings and go back to the previous window.

Should this be the behaviour with the settings dialogue in default? I am half considering implementing a separate NetworkSettingsDialog to keep it separate from the settings dialogue we have right now. What do you think?

micahflee:
> > When you click Configure it opens settings, but if you click Cancel it pops up a "Could not connect to Tor" dialog and quits that app. Instead, it should just close the settings and go back to the previous window.
> 
> Should this be the behaviour with the settings dialogue in default? I am half considering implementing a separate NetworkSettingsDialog to keep it separate from the settings dialogue we have right now. What do you think?

When you open OnionShare and connect to Tor successfully, then open Settings, you can click cancel and it closes fine. The only time it pops up the "Could not connect to Tor" dialog and quits is when you haven't been able to connect at all. So like, if you open OnionShare, connecting fails so you go to settings, tweaking the settings doesn't help and you still haven't connected, then it should quit.

I kind of like the idea of having a separate network settings dialog actually. I'm about to modify the settings dialog to add support for snowflake, so maybe I'll do that too in #1442.

SaptakS:
I like the idea of a separate network settings dialog as well.

SaptakS:
I need to still modify the tests to add the initial screen logic, but apart from that most of the UI is done. Attaching few screenshots:

Added a checkbox in the settings tab:
Screenshot from 2021-12-03 19-06-04


Screenshot from 2021-12-03 19-05-43
Screenshot from 2021-12-03 19-05-37

Showing the same tor progress bar in the widget like in the settings tab instead of a dialong. We might want to make the progress bar look a little nicer and conveying what is actually happening better in the future maybe.
Screenshot from 2021-12-03 19-05-50

micahflee:
Very cool.

A few things:

On the "Connect to Tor" tab, I think the "Network Settings" button should be renamed to "Tor Settings" because it opens the Tor Settings tab, so it's probably good to keep it consistent (and we can re-use a string).

And something I noticed:

- Run OnionShare desktop
- With the "Connect to Tor" tab open, click "Network Settings" or click the Tor icon in the corner
- Click "Save", which connects to Tor in the Tor Settings tab now
- When it finishes connecting, the Tor Settings tab closes and you're back on the "Connect to Tor" tab, but you're already connected and you're not on the New Tab tab

Instead, if the user connects to Tor from the Tor Settings tab, it should _close_ the "Connect to Tor" and open a new tab, just like when you connect to Tor from the Connect to Tor tab.

Also, I noticed that you can go to the Tor Settings tab and start connecting to Tor, and then quickly switch back to the Connect to Tor tab and start connecting to Tor there as well, and it looks like it's just trying to connecting twice, though I'm not certain. One idea to prevent this: When you open the Tor Settings tab from Connect to Tor tab, it should close the Connect to Tor tab. And if you try closing the Tor Settings tab and tor isn't actually connected, it should re-open the Connect to Tor tab.

micahflee:
Also, I need to review the figma designs and everything for this project and how Tor Browser is doing it, but should we make the option to use a bridge clear on the Connect to Tor tab? Or, is the idea that people in censored countries will likely know they need to change the settings?

Here's how it looks like when you boot Tails, FYI:

Screenshot from 2021-12-04 21-08-57

If you check "Connect to Tor automatically (easier)" it shows a bridge checkbox:

Screenshot from 2021-12-04 21-12-12

If you check configure a bridge and click connect:

Screenshot from 2021-12-04 21-12-52

If you check "Hide my location" and click connect:

Screenshot from 2021-12-04 21-13-21

Looks like they haven't implemented moat yet.

micahflee:
Here's what the latest Tor Browser looks like:

Screenshot from 2021-12-04 21-22-00

Seems like this is reasonable, let people just click Connect and if they know what they're doing go into the settings.

SaptakS:
> When it finishes connecting, the Tor Settings tab closes and you're back on the "Connect to Tor" tab, but you're already connected and you're not on the New Tab tab

I and @mig5 discussed this actually, and the implementation we were thinking is passing a boolean parameter to the TorSettingsTab if it is opened from the autoconnect tab. If it is opened from autoconnect tab, then we don't load/reload tor in the settings tab, just save the setting, close the tab, and show the autoconnect tab active. This solves the problem of being able to connect from both tabs. The other option would be like you said replacing the autoconnect tab with settings tab, and once saved and connected, show the home tab. The reason I am a bit against it is in the next part.

> Also, I need to review the figma designs and everything for this project and how Tor Browser is doing it, but should we make the option to use a bridge clear on the Connect to Tor tab? Or, is the idea that people in censored countries will likely know they need to change the settings?

So the idea I am implementing is there are 3 categories of user - Non censored country users, users in censored countries who are aware of bridge and all the intricacies of anti-censorship, users in the censored countries who are not aware of bridges.

For non-censored country users, it's up to them depending on their knowledge to either just click connect or change settings.

For users in censored country aware of bridge settings, they can go to the Tor Settings and do the bridge settings.

For users in censored country who are not accustomed to bridge settings, we guide them through the UX never needing them to go to settings and manually trying to setup bridge (even checking the Use default bridge). The idea based on [Tor Browser's mockup](https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/29590#note_2749371) is that once they click on Connect to Tor, in case they are in censored country, the connection will fail, and it will give them a prompt to try a bridge or configure themselves. If they click try a bridge, it will get location (which will again involve some UX) and then connect to the bridge settings. Once they are done with this, it's saved in their settings, and next time they just get connected. So ideally, such a user will never need to know how to set up a bridge and we guide them through UX asking few questions and setting it up for them. I will be implementing this flow as a separate PR on top of the work @mig5 did in #1482 

This is also why, I am a little skeptcial of the idea of closing the autoconnect tab, and replacing it with TorSettingsTab when the settings is clicked, since that may disrupt this flow.

So this is the Tor Browser kind of idea. Other idea is being super verbose like in tails. But I still feel many not so bridge accustomed users will want to explore the bridge options just like that.

What are your thoughts on the above?

Issue #1442, opened 2021-10-12
Support snowflake bridges, and update bridge selection UI


micahflee:
We need to add support for snowflake bridge (source code [here](https://gitweb.torproject.org/pluggable-transports/snowflake.git)), and we should also update our UI to look like Tor Browser's UI:

![Screenshot from 2021-10-11 19-47-17](https://user-images.githubusercontent.com/156128/136882759-7080158b-8e81-43ca-875b-e222d0c7b19c.png)
![Screenshot from 2021-10-11 19-47-42](https://user-images.githubusercontent.com/156128/136882760-49d21868-0a44-47a5-b3bd-50d575fa0776.png)

While not necessary, if it turns out to not be too much work it would be cool to support "Request a bridge from torproject.org" as well:

![Screenshot from 2021-10-11 19-47-56](https://user-images.githubusercontent.com/156128/136882762-5129d479-ee2c-4874-8d70-0ddbff373fbd.png)
![Screenshot from 2021-10-11 19-48-10](https://user-images.githubusercontent.com/156128/136882766-b616109c-2a03-4eb3-b6c8-fbe14f351e8b.png)

Right now during development in Linux we tell users to install `tor` and `obfs4proxy` from their package manager, but `snowflake` isn't widely packaged. So this issue will involve making a `get-tor-linux.py` that downloads the linux Tor Browser and extracts binaries, including `snowflake-client` (which is in `tor-browser_en-US/Browser/TorBrowser/Tor/PluggableTransports/snowflake-client`), for development. And then we'll build snowflake from source for Flatpak and Snapcraft.

micahflee:
I moved the packaging stuff to #1459 so that PR $1449 can finally close this issue.

mig5:
I think we can close this one out since #1449 was merged! #1459 contains the remaining packaging steps.

PR #1444, opened 2021-10-15
Snowflake support, and updated Tor Settings

micahflee:
This does much of #1442, though we'll need another PR to finish it.

Since the snowflake binary isn't included in Linux packages, this changes the desktop instructions for Linux to also download Tor Browser and extract the relevant binaries.

Now there are two settings buttons in the bottom-right, one for Tor Settings and one just for Settings:

Screenshot from 2021-10-15 08-41-52

Here's what Tor Settings looks like:

Screenshot from 2021-10-15 08-42-44

The bridge options, which are only available with "Use the Tor version built into OnionShare", now include the same description text that Tor Browser uses.

If you select "Connect using control port" or "Connect using socket file", it puts the remaining settings in a separate group box called "Tor settings":

Screenshot from 2021-10-15 08-43-38

The "Test Connection to Tor" button is now in the bottom of the window, and it now uses the same Tor Connection Dialog as when you initially connect to Tor. That has been slightly refactored to allow for this, but it's nice because it fixes the issue where the settings window was frozen while waiting to test the Tor connection.

Screenshot from 2021-10-15 08-46-01
Screenshot from 2021-10-15 08-46-16

When you first open OnionShare and it tries connecting to Tor, if it fails connecting or if you click cancel, it opens Tor Settings and instead of just Settings.

This is what Settings looks like now:

Screenshot from 2021-10-15 08-47-35

I'm not totally sold on the layout, but I do think this looks better with the version and help button at the top, than before with it in the bottom-left of the window. In Windows and macOS, there should be an additional widget for checking for updates, but I still need to test that.

Still left to do:

- [x] Make sure the pluggable transports work correctly, and that tor is actually using them when it connects
- [x] Implement moat, to let users request a bridge from torproject.org
- [ ] Use domain fronting for the BridgeDB http requests
- [ ] Update tor binaries stuff to work in Windows, macOS
- [ ] Make sure check for update in Settings dialog still works in Windows, macOS
- [ ] Make snap, flatpak build snowflake binary from source

### DIFF
--- a/desktop/src/onionshare/tor_settings_dialog.py
+++ b/desktop/src/onionshare/tor_settings_dialog.py
@@ -809,10 +809,14 @@ class TorSettingsDialog(QtWidgets.QDialog):
                         meek_lite_pattern = re.compile(
                             "(meek_lite)(\s)+([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+:[0-9]+)(\s)+([0-9A-Z]+)(\s)+url=(.+)(\s)+front=(.+)"
                         )
+                        snowflake_pattern = re.compile(
+                            "(snowflake)(\s)+([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+:[0-9]+)(\s)+([0-9A-Z]+)"
+                        )
                         if (
                             ipv4_pattern.match(bridge)
                             or ipv6_pattern.match(bridge)
                             or meek_lite_pattern.match(bridge)
+                            or snowflake_pattern.match(bridge)
                         ):
                             new_bridges.append(bridge)
                             bridges_valid = True


mig5:
I may have found another issue - after setting a custom bridge and Tor reconnecting, if I then go back to the Bridge settings and uncheck 'Use a bridge', the custom bridge, as well as `UseBridges 1`, remains in the torrc even after Tor reconnects.

mig5:
So I think even though we are setting `"no_bridges": true,` in the onionshare.json, we aren't unsetting the `  "tor_bridges_use_moat": true,` or the values of `tor_bridges_use_moat_bridges`, or `  "tor_bridges_use_custom_bridges"` when using custom bridges. 

So I think the onion.py is still injecting the custom torrc bridge settings, even if `no_bridges` is True.

Perhaps we need to also evaluate the value of `no_bridges` in each of our 'if' statements in onion.py where checking what bridges/template file to write.

PR #1447, opened 2021-10-22
Fixed send.js table sorting.

tladesignz:
This looks like a long forgotten feature, since it failed silently... I fixed it for the changed markup and made it work with non-size columns.


micahflee (approved):
Nice, thank you this works great.

PR #1475, opened 2021-11-19
Stop storing temporary files in /tmp

### LINKED ISSUE #1470: Stop storing temporary files in /tmp
There are a few places where temporary files are created in `/tmp` and they're not always cleaned up. Instead, these should go in the OnionShare config folder (like `~/.config/onionshare/tmp`), and they should always get cleaned up.

```
$ grep -r -I tempfile cli/onionshare_cli/
cli/onionshare_cli/web/receive_mode.py:import tempfile
cli/onionshare_cli/web/receive_mode.py:            self.f = tempfile.TemporaryFile("wb+")
cli/onionshare_cli/web/send_base_mode.py:import tempfile
cli/onionshare_cli/web/send_base_mode.py:                gzip_filename = tempfile.mkstemp("wb+")[1]
cli/onionshare_cli/web/share_mode.py:import tempfile
cli/onionshare_cli/web/share_mode.py:            self.gzip_filename = tempfile.mkstemp("wb+")[1]
cli/onionshare_cli/web/share_mode.py:            self.zip_temp_dir = tempfile.mkdtemp()
cli/onionshare_cli/onion.py:import tempfile
cli/onionshare_cli/onion.py:                self.tor_data_directory = tempfile.TemporaryDirectory(
```

We should make sure that when creating a temporary file or directory, to specify a prefix in the config dir, and also we should stop using `tempfile.mkstemp` and `tempfile.mkdtemp`. From the [tempfile docs](https://docs.python.org/3/library/tempfile.html):

> This module creates temporary files and directories. It works on all supported platforms. TemporaryFile, NamedTemporaryFile, TemporaryDirectory, and SpooledTemporaryFile are high-level interfaces which provide automatic cleanup and can be used as context managers. _mkstemp() and mkdtemp() are lower-level functions which require manual cleanup._

micahflee:
Fixes #1470.

This replaces `tempfile.mkstemp` and `tempfile.mkdtemp` with the higher-level `tempfile.NamedTemporaryFile` and `tempfile.TemporaryDirectory` classes.

They also now store all temporary files in OnionShare's tmp dir (`~/.config/onionshare/tmp`) and not in `/tmp` anymore. This should solve various problems, like with `/tmp` running out of disk space on systems that limit it a lot. Also, although I think this is outside of our threat model, on multi-user systems it no longer writes in the world-readable `/tmp`.

The clean up after themselves, either when the file is closed or the object is garbage collected, so it's less brittle than how cleanup was happening before.


mig5 (changes requested):
This does sound cool and that it will solve various problems.

I found a couple of issues in testing:

## Sharing a single file crashes

When I share a single file, OnionShare crashes and I see the 'compressing' widget even though it was a tiny text file. There is a trackback in my terminal:

```
  File "/home/user/git/onionshare/desktop/src/onionshare/tab/mode/share_mode/threads.py", line 47, in run
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/onionshare_cli/web/send_base_mode.py", line 131, in set_file_info
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/onionshare_cli/web/share_mode.py", line 425, in set_file_info_custom
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/onionshare_cli/web/share_mode.py", line 499, in build_zipfile_list
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/onionshare_cli/web/send_base_mode.py", line 312, in _gzip_compress
  File "/usr/lib/python3.7/gzip.py", line 57, in open
TypeError: filename must be a str or bytes object, or a file
```

## Folder/file is not removed until OnionShare exits entirely

When I shared two files (thus creating a zip), there was no crash. I was able to download the zip in Tor Browser. However, after the share-stopped when my download completed, the tmp folder in my config dir (with the zip file inside it) was still there in my filesystem.

When I closed OnionShare altogether, the tmp folder disappeared. 

If I close just the tab, the tmp dir is still there. So it seems like the entire Python process running OnionShare itself, is what is holding onto the 'temporary' nature of the file, not merely the tab or its status.

I think if we get it to remove the file on 'Server Stopped' event that is best. On the other hand it means we're back to manually cleaning up the file, which is what we wanted to avoid...

But maybe it's totally ok if you clean up all the files when OnionShare stops fully?

## Impact on 'leaving a trace'

And just noting for the record, that we may need/want to consider the implications of a 'temp' file being retained on disk due to an unexpected segfault or other kind of crash of the app (or crash of the user's computer entirely), whereby Python's garbage collection couldn't clean up. When things were on `/tmp`, probably a lot of popular OS would wipe that on reboot. I'm not sure if Windows has similar logic.

On the other hand, you could make the case that someone who wants plausible deniability maybe should be using something truly ephemeral at a lower level, like Tails. There are probably already other files we might leave on disk - such as when a user changes settings and we write the json. But just noting it's a change with subtle new implications, so good to discuss even if we decide the benefits outweigh the risk.

micahflee:
Okay, I believe I have fixed all of these now!

I don't actually have a problem with explicitly cleaning up files, I just wanted to use the higher-level tempfile classes instead of the low-level functions. So now I've added back in closing `tempfile.NamedTemporaryFile` objects and calling cleanup on `tempfile.TemporaryDirectory` objects.

I also fixed a bit of work so that it calls the tab's cleanup method at the right time. So now when a server stops, or a tab is clone, it's cleaned up, and all the files are deleted.

As far as the impact on "leaving a trace", I think that in some ways this PR actually helps with this problem because unless the python process hard crashes, the garbage collector should delete these tempfile objects, which should delete the files on disk. So even if we hit an exception somewhere, the files will get deleted. But that said, it is possible that there's a hard crash... but I think this is an edge-case and the risk is low, so I'm okay with it.

mig5:
It's looking good!

There is one problem that came up in the GUI tests in Website Mode, and I reproduced it. 

I can start and visit a Website the first time (I used a clone of https://github.com/onionshare/onionshare-website to test with).

But if I try to load the site again, I get this traceback:

```
[2021-12-02 16:59:35,575] ERROR in app: Exception on / [GET]
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/onionshare_cli/web/website_mode.py", line 43, in path_public
    return path_logic(path)
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/onionshare_cli/web/website_mode.py", line 49, in path_logic
    return self.render_logic(path)
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/onionshare_cli/web/website_mode.py", line 110, in render_logic
    return self.stream_individual_file(self.files[index_path])
  File "/home/user/git/onionshare/desktop/venv/lib/python3.7/site-packages/onionshare_cli/web/send_base_mode.py", line 207, in stream_individual_file
    self.web.cleanup_tempfiles.append(gzip_file)
UnboundLocalError: local variable 'gzip_file' referenced before assignment
127.0.0.1 - - [02/Dec/2021 16:59:35] "GET / HTTP/1.1" 500 -
```

... and the site throws a 500 internal server error.

The gzip files are all generated 'on demand' as the page loads the first time. But that 'cleanup_tempfiles' occurring on the second visit makes me wonder if it's trying to remove the existing gzipped files and recreate them? Not really sure, haven't dug under the hood yet.

micahflee:
Very good catch. This should fix it. This was the code:

```python
# gzip compress the individual file, if it hasn't already been compressed
if use_gzip:
    if filesystem_path not in self.gzip_individual_files:
        self.gzip_files.append(
            tempfile.NamedTemporaryFile("wb+", dir=self.common.build_tmp_dir())
        )
        gzip_file = self.gzip_files[-1]
        self._gzip_compress(filesystem_path, gzip_file.name, 6, None)
        self.gzip_individual_files[filesystem_path] = gzip_file.name

    file_to_download = self.gzip_individual_files[filesystem_path]
    filesize = os.path.getsize(self.gzip_individual_files[filesystem_path])

    # Cleanup this temp file
    self.web.cleanup_tempfiles.append(gzip_file)
```

Basically it lazily gzips files the first time the path is loaded, and then in the future when the same path is loaded it uses the already-gzipped file. The problem is it was trying to add the gzip file to the cleanup list every time the path is loaded, instead of just when it creates the gzip the first time.

This fixes it:

```python
# gzip compress the individual file, if it hasn't already been compressed
if use_gzip:
    if filesystem_path not in self.gzip_individual_files:
        self.gzip_files.append(
            tempfile.NamedTemporaryFile("wb+", dir=self.common.build_tmp_dir())
        )
        gzip_file = self.gzip_files[-1]
        self._gzip_compress(filesystem_path, gzip_file.name, 6, None)
        self.gzip_individual_files[filesystem_path] = gzip_file.name

        # Cleanup this temp file
        self.web.cleanup_tempfiles.append(gzip_file)

    file_to_download = self.gzip_individual_files[filesystem_path]
    filesize = os.path.getsize(self.gzip_individual_files[filesystem_path])
```

mig5 (approved):
Awesome! This is working great now.

PR #1484, opened 2021-12-10
Fetch builtin bridges in get tor scripts

mig5:
Following on from https://github.com/onionshare/onionshare/pull/1478

This updates the 'get tor' scripts to also fetch the latest built-in bridges and update our torrc templates. Meant for just prior to release when we update the Tor versions etc.


micahflee (approved):
This is great. I tested running one of the `get-tor-` scripts, and I tested connecting with all of the built-in bridge types.

PR #1489, opened 2021-12-15
Don't enqueue stdout from Meek subprocess in a Thread

mig5:
I discovered that whenever we called `meek.cleanup()`, the CPU of the python process went up to 100%, and the fan on my laptop was going wild.

Try it yourself on develop branch, by launching the Moat/BridgeDB captcha and fetch, and watch top or htop in another terminal tab? Is it just me? I'm on Debian Bullseye these days.. Python 3.9.2.

Took a while to find, but there was an infinite `read()` being called. It has to do with the way we tried to enqueue the stdout of the meek process in a non-blocking thread in order to read the random port number that Meek assigns to itself on start-up.

That work was based on https://stackoverflow.com/a/4896288

This seems to be a problem when we terminate the actual subprocess in `cleanup()` - it *does* kill the subprocess, but I think somehow creates an endless read() on this queue that now is not getting anything added to it (..?? not sure my understanding of the issue is right, but it's definitely where the issue lies)

Anyway, I've simplified this a bit so we just iterate over the stdout and break once we get the port. It... seems to work this way? And no longer causes the CPU 100% issue..

I'm slightly troubled because once I got this after I got the bridges back from Moat (and meek.cleanup() had already been called):

```
QThread: Destroyed while thread is still running
./scripts/dev.sh: line 9:  7591 Aborted                 python -c "import onionshare; onionshare.main()" $@
```

But now I can't reproduce it.. and when I got it, I was actually on the other PR branch for the Censorship work.

Anyway, if you can think of a way to simplify this even further in terms of 'catching' the Meek port number, I welcome it, but this seems pretty consistent now. Interested to hear if you reproduce the CPU issue in the first place.

### DIFF
--- a/cli/onionshare_cli/meek.py
+++ b/cli/onionshare_cli/meek.py
@@ -20,8 +20,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import os
 import subprocess
 import time
-from queue import Queue, Empty
-from threading import Thread
 
 
 class Meek(object):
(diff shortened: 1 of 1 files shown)


mig5:
Adding to the 2.4.1 milestone as that release should definitely not happen without this fix.

micahflee (approved):
I'm actually not reproducing this issue... but that said, this is shorter and cleaner code anyway. If it fixes it for you I'm happy with merging.

PR #1491, opened 2021-12-17
Force javascript MIME type to work around silly operating system MIME databases

### LINKED ISSUE #1443: Incorrect MIME setting in operating system causes Flask to share javascript files as text/plain and breaks Chat
When trying to use an onionshare chatroom, no textboxes or other visuals besides the logo and basic elements are loading.

image

From what I can tell, it looks like there is some sort of formatting error with the js.

image

I've disabled all of Tor's protections and gave it the most liberal permissions that I can on noscript, but nothing changes at all. Both Onionshare and Tor are fully up to date, so I'm not sure what the problem could be. Could be user error, but I really don't know what I could have messed up, I've tried everything that I could come up with.

mig5:
Fixes #1443 (I think)

Flask relies on the OS' MIME database to guess which Content-Type header to send with the files it serves back to the browser.

In the case of Windows, the MIME database is in the Registry.

It seems that some applications might edit the Registry to change the MIME type, for at least javascript - possibly to `text/plain`. Perhaps it's an IDE that is the culprit, so that javascript files can be opened in an editor or something.

The effect is that javascript such as the Chat mode, breaks because of our X-Content-Type-Options 'nosniff' header. The browser refuses to execute the javascript as it's seen as text/plain.

People are assuming it's OnionShare's fault without realising their registry's probably been modified. It's likely OnionShare is the first time they've served an application from their own computer as a server so haven't come across the issue before.

I reproduced the issue on my Windows machine by going into RegEdit and adding a key 'Content Type' with value 'text/plain' in the `HKEY_CLASSES_ROOT` -> `.js`. 

On my Windows machine, I didn't have any such key to begin with, but when I created one and set it to text/plain, I reproduced the issue in a Chat service that was set up on my Windows OnionShare: the javascript didn't load in my macOS Tor Browser and I got Console errors about the MIME type, exactly per the above issue. So I'm pretty confident this is the cause.

Fortunately we can import `mimetypes` and force set it to `text/javascript`, and regardless of what the registry says, it works as expected after that!

See my comment in the issue for other opensource projects on Github that have found the same issue before.

### DIFF
--- a/cli/onionshare_cli/web/web.py
+++ b/cli/onionshare_cli/web/web.py
@@ -18,6 +18,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import logging
+import mimetypes
 import os
 import queue
 import requests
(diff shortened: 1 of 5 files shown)


micahflee (approved):
What a bizarro bug. I just tested it out in Windows though and confirmed that it works. My Windows 10 didn't have that registry key set, but I tried setting it and confirmed that chat broke in the develop branch, and that your PR fixes it.

PR #1494, opened 2021-12-21
Connect to the emitted TorConnectionDialog success to trigger tab update (fixes persistent tab update)

### LINKED ISSUE #1493: Un-break persistence
Some of the new Tor Settings changes have broken persistence.

- Open OnionShare, connect to Tor, choose a mode, check persistent, start the service
- Quit OnionShare
- Open OnionShare again: the persistent tab should be open with the text, "OnionShare is not connected to the Tor network"
- Connect to Tor

After Tor connects again, the tab never gets updated and still has the text "OnionShare is not connected to the Tor network", so you can't start the service again.

Screenshot from 2021-12-20 17-11-05

Note that this is only a problem to deal with in the 2.4.1 release. I just tested and this problem actually doesn't seem to exist in the `censorship` branch.

mig5:
Fixes #1493

It is probably fixed slightly differently in censorship branch, but this should do I think.

### DIFF
--- a/desktop/src/onionshare/main_window.py
+++ b/desktop/src/onionshare/main_window.py
@@ -163,6 +163,7 @@ class MainWindow(QtWidgets.QMainWindow):
         # Start the "Connecting to Tor" dialog, which calls onion.connect()
         tor_con = TorConnectionDialog(self.common)
         tor_con.canceled.connect(self.tor_connection_canceled)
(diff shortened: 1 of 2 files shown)


SaptakS (approved):
This looks similar to the approach I took in the censorship branch with the connection tab. So I think this should work. Also, in debian, it seems to be working.

micahflee (approved):

Issue #1483, opened 2021-12-05
'Permission denied' when sharing files on Windows 10 (using develop branch)


mig5:
When I try to share a file on Windows 10 using the current `develop` branch, I get 'Permission denied' when the share goes to start.

The 'Compressing files' progress bar widget is also unexpectedly shown (given it's a tiny file).

My suspicion for now is that the 'TemporaryFile' work maybe caused this - I think the permission denied might be when creating the temporary file.

I am not sure if it's superficial but in verbose mode, there is a path logged which has forward slashes instead of backslashes: `C:/Users/migue/git/onionshare/desktop/src/setup.py`.

![43F598BB-8A04-4C7E-98DB-A3BAA9ED1606](https://user-images.githubusercontent.com/99173/144767826-27fe3921-abec-461d-8acc-ad9f232ae4df.jpg)

I didn't try Website mode, but assume it's the same.

Chat and Receive modes are working fine.

PR #1495, opened 2021-12-23
Switch to cx_Freeze

### LINKED ISSUE #1451: Updated OnionShare to 2.4 on Windows 10 but doesn't start [HELP]
Hi @,

I've just updated Onionshare to v.[2.4](https://github.com/onionshare/onionshare/releases/download/v2.4/OnionShare-2.4.msi) but after installation (without error message) it doesn't start anymore, (looping a few  seconds then nothing launches)

I'm running a Win10 machine up to date.

Tried to uninstall Onionshare and resinstalled it again with no more success.

Would you please support if you have any idea what could the issue be?

Thanks in advance.

micahflee:
This isn't quite done yet. It will fix #1451 and #1483.

This switches from briefcase to cx_Freeze for Windows and macOS packaging.

By removing briefcase, all the of the desktop code has moved from `desktop/src/onionshare` to just `desktop/onionshare`, and the desktop version is now using poetry too. This actually makes it way nicer to work with because there's no more having to run `scripts/rebuild-cli.py` -- instead it just pulls from the actual `onionshare_cli`.

Another cool thing with this change is that it makes legit separate `onionshare` and `onionshare-cli` binaries for Windows and macOS. This wasn't so much of a problem before for Mac, but this will add support for the CLI in Windows, using the bundled tor binaries.

To test freezing the code, you `poetry run python setup-freeze.py build`.

I hit a gnarly bug with cx_Freeze when trying to build binaries in macOS and Linux that involved some monkeypatching at the top of `setup-freeze.py`, but it seems to work.

What's left for Windows:

- The `desktop/package/build-windows.py` script needs to codesign the Windows binaries
- It also needs to build an msi installer. I was did see that [cx_Freeze has bdist_msi](https://cx-freeze.readthedocs.io/en/latest/setup_script.html#bdist-msi) for creating an msi installer, but it doesn't have the codesigning built-in, and didn't seem to work well, so I was thinking of using [WiX Toolset](https://wixtoolset.org/).

What's left for macOS:

- Building the binary with `setup-freeze.py build` works, but I need to test that `setup-freeze.py bdist_mac` creates a workable app bundle. It seems to, but I haven't tested it with codesigning yet
- After creating the app bundle, should decide if I want to use `setup-freeze.py bdist_dmg` to create a dmg for release or not. It seems to work and is simpler, but it's not as pretty as the dmg in earlier releases.
- I need to figure out how to hook into the build step to delete unused files from the PySide2 library, which can shave like 200mb off the package size
- Create a `desktop/package/build-macos.py` script that automates all the steps

Potential for Linux AppImage release:

- I was also looking into use cx_Freeze in Linux to make an AppImage release. It actually seems like it might not be too difficult... but maybe this should be for a later PR so that it doesn't hold back the 2.4.1 release.

And some other things to finish:

- Fix the GUI test. I think they're all broken now just because the GUI code moved from `desktop/src/onionshare` to `desktop/onionshare`
- Make sure `RELEASE.md` is all up-to-date with the new way to make Windows and macOS releases


mig5:
Finally getting around to a light review of this.

I was able to get through the `poetry install` on Python 3.9.9 on my Mac (10.15.7 Catalina) (after also upgrading my Poetry version to the latest version), and also the cxfreeze stuff. I assume one doesn't need to run the cxfreeze commands in the `cli` directory, just `desktop`? It errored out for me in the cli dir, anyway.

I could also build an .app but I couldn't run it due to the code signature stuff.

Thanks for fixing the tests, @SaptakS. I think previously we installed `pytest-xvfb` in the .circleci/config.yml which takes care of running the tests with xvfb (without needing to explicitly wrap it through `xvfb-run`). But the latter works just as well.

micahflee:
Finished all the Windows stuff, including adding code signing, and I even made the installer look pretty.

windows-installer

micahflee:
After WAY too much time troubleshooting, I finally got Mac packaging to work. We hit this issue: https://github.com/marcelotduarte/cx_Freeze/issues/594

And I worked around it by making `build-mac.py` build the cx_freeze app bundle then move files around to conform with what Apple's notary service expects, then codesign.

This PR is finally ready for review.

PR #1505, opened 2022-01-16
Install CLI dependencies in desktop test job

mig5:
The test-gui job was missing the `poetry install` in the cli folder, so the job was failing as the backend was not ready.

There's probably a way to do that as a general pre-job setup step that works for both CLI and GUI test jobs, but this seems to suffice.


micahflee (approved):

micahflee:
Actually, I don't think this is needed. The desktop `pyproject.toml` loads the CLI and all it's dependencies already:

```
onionshare_cli = {path = "../cli", develop = true}
```

The test that failed before with a segfault, I think that was just a random rare failure we get for some reason. I re-ran the same tests and it passed.

PR #1513, opened 2022-01-21
Advanced Usage documentation reworked 2

comradekingu:


docs/source/advanced.rst
```diff
@@ -6,19 +6,19 @@ Advanced Usage
...
 
 .. image:: _static/screenshots/advanced-save-tabs.png
 
-When you quit OnionShare and then open it again, your saved tabs will start opened.
-You'll have to manually start each service, but when you do they will start with the same OnionShare address and private key.
+When opening OnionShare, your saved tabs from the prior session will start opened.
+Each service then can be started manually, and will be available on the same OnionShare address and be protected by the same private key.
 
-If you save a tab, a copy of that tab's onion service secret key will be stored on your computer with your OnionShare settings.
+If you save a tab, a copy of its onion service secret key is stored on your computer with your OnionShare settings.
```
comradekingu, line 21: "with your OnionShare settings" isn't very clear. In the settings, settings for (?) is also saved alongside the key?

docs/source/advanced.rst
```diff
@@ -78,14 +83,14 @@ Then run it like this::
 
     onionshare-cli --help
 
-For information about installing it on different operating systems, see the `CLI readme file <https://github.com/onionshare/onionshare/blob/develop/cli/README.md>`_ in the git repository.
+Info about installing it on different operating systems can be found in the `CLI README file <https://github.com/onionshare/onionshare/blob/develop/cli/README.md>`_ in the Git repository.
```
comradekingu, line 86: As per https://no.wikipedia.org/wiki/Git

docs/source/advanced.rst
micahflee: What this means is that OnionShare settings are all stored in the same folder on the disk (in different places depending on the user's OS), and it's letting them know that secret keys will be stored in that folder as well. Though it could just say, "f you save a tab, a copy of its onion service secret key is stored on your computer."

docs/source/advanced.rst
```diff
@@ -6,19 +6,19 @@ Advanced Usage
 Save Tabs
 ---------
 
-Everything in OnionShare is temporary by default. If you close an OnionShare tab, its address no longer exists and it can't be used again.
-Sometimes you might want an OnionShare service to be persistent.
-This is useful if you want to host a website available from the same OnionShare address even if you reboot your computer.
+Everything in OnionShare is temporary by default. Addresses no longer exist and can't be used again when OnionShare tabs are closed.
+Your OnionShare service can also be persistent.
+If host a website, percistence means it will be available on the same OnionShare address even if you reboot your computer.
```
micahflee, line 11: Instead of, "If host a website" it should be, "If you host a website"

docs/source/advanced.rst
micahflee: "with your OnionShare settings" can just be removed.

docs/source/advanced.rst
```diff
@@ -29,35 +29,40 @@ By default, all OnionShare services are protected with a private key, which Tor
 
 When browsing to an OnionShare service in Tor Browser, Tor Browser will prompt for the private key to be entered.
 
-Sometimes you might want your OnionShare service to be accessible to the public, like if you want to set up an OnionShare receive service so the public can securely and anonymously send you files.
+You can make a OnionShare service accessible to the public. Set up an OnionShare receive service so the public can securely and anonymously send you files.
```
micahflee, line 32: The receive mode service is just a single example of when you might want to make a service public. Instead of, "Set up an OnionShare receive service", it should be something like, "For example, you can set up an OnionShare receive service"

micahflee (changes requested):
In the final "Usage" section of the advanced documentation, we actually shouldn't change any of that because that's just copied and pasted directly from the CLI interface, so changing it the documentation will actually make it wrong.

And speaking of which, the CLI version does not get localized, it's only in English. So it occurs to me that the code block in the Usage section should actually not get translated at all. I'm not sure how to convey this in Weblate.

docs/source/advanced.rst
comradekingu, line 21:
```suggestion
If you save a tab, a copy of its onion service secret key is stored on your computer.
```

docs/source/advanced.rst
comradekingu, line 32:
```suggestion
You can files accessible to the public by setting up an OnionShare receive service.
```

Issue #1519, opened 2022-01-29
Proposal: ephemeral forums


dzwdz:
The ephemeral chat feature is pretty nice, but a realtime chat isn't always ideal. It only works well when everyone can be online at the same time. When you want to convey a lot of information, having to express yourself a single line at a time is pretty limiting too. It'd be nice if OnionShare had a way to hold some more structured discussions.

I think that emphemeral textboards (or maybe even imageboards) could work well for this purpose. Something like an even simpler version of [textboard.org](textboard.org), maybe?

PR #1523, opened 2022-02-05
Add textboard mode to onionshare

sqbi-q:
Implements #1519
Allows textboard users to write text threads and reply to them.

image

### DIFF
--- a/cli/onionshare_cli/__init__.py
+++ b/cli/onionshare_cli/__init__.py
@@ -60,6 +60,9 @@ def main(cwd=None):
     parser.add_argument(
         "--chat", action="store_true", dest="chat", help="Start chat server"
     )
+    parser.add_argument(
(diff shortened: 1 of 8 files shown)


SaptakS:
If I am not wrong, you are saving the thread in a file, right? How is it ephemeral if it's saving the thread post in file? In fact I feel that could be a very big security issues, since the one hosting the thread has all the thread posts saved in a file unencrypted? So if their device gets compromised, then the entirely unencrypted thread is accessible, right?

sqbi-q:
@SaptakS Hey, thanks for replying (I've been waiting for feedback for a while). 
You are absolutely right, threads are stored as files and aren't encrypted. This isn't ephemeral, nor secure. My bad.
I think I should just close this low effort pull request, that was written while ago.

PR #1526, opened 2022-02-14
Rocket ship animation

### LINKED ISSUE #1486: Make the rocket ship animation in quickstart screen
135704687-356293ab-546e-4a06-8392-3c5b173e1d73

micahflee:
Fixes #1486.

In addition to the rocket ship animation, this also removes country shapes.


mig5 (changes requested):
Functionally this seems pretty good!

I hit one bug: I canceled my Tor connection, chose country 'Turkmenistan' and then clicked 'Use a bridge', and I got this error:

```
[Feb 14 2022 04:30:10 PM] AutoConnectTab.use_bridge_connect_clicked: Trying to automatically obtain bridges
  File "/home/user/git/onionshare/desktop/onionshare/connection_tab.py", line 228, in use_bridge_connect_clicked
    country = self.use_bridge_widget.country_code
AttributeError: 'AutoConnectUseBridgeWidget' object has no attribute 'country_code'
```

We probably just need to update whatever attribute this is now that the maps are gone.. we need to make sure we still have access to the country code names.

In terms of the visuals, I found it *slightly* distracting with the background planets being so close to the 'toggle' switch:

Screenshot_2022-02-14_16-28-20

If I stretch the window, causing all the stars/planets backdrop to be more to the right, then I find it's a bit cleaner, but it might just be personal preference. 

I also noticed it seemed like only one engine, the one on the left, gets a flame, is that deliberate @glenn-sorrentino ?

Screenshot_2022-02-14_16-29-28

Otherwise seems good. Depending on your network connection process time, the ship can be a little jerky to take off.. but I think it's fine.

glenn-sorrentino:
Good catch on the rocket, @mig5. @micahflee, the updated asset is attached. 
Ship

mig5:
@micahflee here's a patch to fix the country_code bug

```
diff --git a/desktop/onionshare/connection_tab.py b/desktop/onionshare/connection_tab.py
index c1398cbd..2eb177ee 100644
--- a/desktop/onionshare/connection_tab.py
+++ b/desktop/onionshare/connection_tab.py
@@ -225,7 +225,7 @@ class AutoConnectTab(QtWidgets.QWidget):
         if self.use_bridge_widget.detect_automatic_radio.isChecked():
             country = False
         else:
-            country = self.use_bridge_widget.country_code
+            country = self.use_bridge_widget.country_combobox.currentData().lower()
 
         self._censorship_progress_update(
             50, strings._("gui_autoconnect_circumventing_censorship_starting_meek")
```

micahflee:
Okay this fixes everything! I also moved the stars and planets image right 40px.

Screenshot from 2022-03-06 11-57-50

PR #1537, opened 2022-03-01
fix broken jinja2/markupsafe dependency chain

3nprob:
Fix #1536

Issue #1536, opened 2022-03-01
Outdated dependencies breaks onionshare


3nprob:
See https://github.com/pallets/flask/issues/4455 - subdependencies of flask have released minor updates including breaking changes causing onionshare to exit with error on startup.

Flask really should be updated but in the interim this can be resolved by pinning the right dependencies (see above issue)

Related issue: #1527 

```
Traceback (most recent call last):
  File "/usr/local/bin/onionshare-cli", line 5, in <module>
    from onionshare_cli import main
  File "/usr/local/lib/python3.10/site-packages/onionshare_cli/__init__.py", line 30, in <module>
    from .web import Web
  File "/usr/local/lib/python3.10/site-packages/onionshare_cli/web/__init__.py", line 21, in <module>
    from .web import Web
  File "/usr/local/lib/python3.10/site-packages/onionshare_cli/web/web.py", line 28, in <module>
    import flask
  File "/usr/local/lib/python3.10/site-packages/flask/__init__.py", line 14, in <module>
    from jinja2 import escape
  File "/usr/local/lib/python3.10/site-packages/jinja2/__init__.py", line 12, in <module>
    from .environment import Environment
  File "/usr/local/lib/python3.10/site-packages/jinja2/environment.py", line 25, in <module>
    from .defaults import BLOCK_END_STRING
  File "/usr/local/lib/python3.10/site-packages/jinja2/defaults.py", line 3, in <module>
    from .filters import FILTERS as DEFAULT_FILTERS  # noqa: F401
  File "/usr/local/lib/python3.10/site-packages/jinja2/filters.py", line 13, in <module>
    from markupsafe import soft_unicode
ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/usr/local/lib/python3.10/site-packages/markupsafe/__init__.py)
```

mig5:
Fixed via https://github.com/onionshare/onionshare/pull/1541

PR #1285, opened 2021-02-22
Updates socket.io client to 3.1.1 with the latest Socket.IO and Engine.IO protocols

SaptakS:
Fixes #1284 
Seems like we updated flask-socket.io to 5.x from 4.x but didn't update the socket.io.min.js file to match the latest client protocols. Hence there was a version incompatibility based on this: https://flask-socketio.readthedocs.io/en/latest/#version-compatibility. Updating the socket.io.min.js to the l[atest version](https://github.com/socketio/socket.io-client/releases/tag/3.1.1) fixes that.

PR #1541, opened 2022-03-06
Update flask to 2.0.3

### LINKED ISSUE #1527: incompatible click version
installing the latest onionshare-cli with pip:

```
flask 1.1.4 has requirement click<8.0,>=5.1, but you'll have click 8.0.3 which is incompatible
```

You may just want to upgrade to flask. I'm not sure how this ended up here.

micahflee:
This upgrades flask to 2.03, flask-socketio to 5.1.1, and also updates other pip deps. And it updates to the latest  socketio javascript version too.

This should fix #1527 and #1536, and solve #1537 by updating all of the dependencies instead of locking `markupsafe` to a lower version.

Here's one way to test it. In the `develop` branch, in the `cli` dir, run:

```sh
poetry install
poetry build
```

Then somewhere else, create a virtual env and pip install `cli/dist/onionshare_cli-2.5-py3-none-any.whl` into it.

If you try running `onionshare-cli`, it will crash with the error message in #1536. 

Now do the same thing, but this time build the .whl from the `upgrade-flask` branch. It should work.


micahflee:
Also, this PR updates Flask to a major version. I confirmed the tests all pass, and that chat mode seems to work, but it would be good to manually click through various modes to make sure the new version of Flask doesn't break anything untested.

mig5 (approved):
I could not reproduce the pip install problem in a venv, but I didn't have any problem with this branch either. 

I manually did a brief test of all modes with full Tor mode - but didn't test autostop/start timers, or Client Auth, or other features - just public mode in Share, Receive, Chat and Website modes). Everything seems ok.

It's annoying that these tests keep segfaulting. I also ran the GUI tests locally (local mode) and indeed everything passes.

3nprob:
Is there anything blocking this from merge currently?

cli/onionshare_cli/resources/static/js/socket.io.min.js
```diff
@@ -1,7 +1,7 @@
 /*!
- * Socket.IO v3.1.1
- * (c) 2014-2021 Guillermo Rauch
+ * Socket.IO v3.0.4
```
3nprob, line 2: This downgrade is not on purpose and due to being behind upstream I assume?

socket.io was upgraded in #1285

micahflee: Thanks for noticing. That was not on purpose. Let me update it to the latest version and see if it works.

Issue #1547, opened 2022-03-13
Better UI error messages in the connection flow


SaptakS:
There are several different reasons because of which a Tor Connection may fail at the initia stage. It may due to the censorship which takes the user to the circumvention flow, but also after that there can be errors like if Meek is missing in the system, or the configuration provided for circuumvention doesn't work, or no configuration was found for that specific country, etc. Better UI error message with a possible relevant solution will help users get a next step.

Also, in case a successful connection is made after the circumvention configurations, it might be a good idea to show the user what configuration they are using. For example a message like: "You are successfully connected to Tor network via snowflake"

PR #1552, opened 2022-03-22
More verbose and consistent Meek exception logging

mig5:
Works towards #1547

I concentrated on just making the logs more verbose. I don't know that we should actually throw UI alerts on MeekNotFound or MeekNotRunning because the current design is to gracefully fall back to trying built-in obfs4 bridges in those scenarios.

But we now have more clarity on what might've gone wrong. For example, move your `meek-client` binary out of the way and you'll see:

```
[Mar 22 2022 02:18:35 PM] AutoConnectTab.use_bridge_connect_clicked: Trying to automatically obtain bridges
[Mar 22 2022 02:18:35 PM] MeekNotFound.__init__: Could not find the meek binary
[Mar 22 2022 02:18:35 PM] AutoConnectTab._got_no_bridges: Could not obtain bridges, so falling back to trying built-in obfs4 bridges
```

Or, if you edit cli/onionshare_cli/meek.py and deliberately break the meek start-up parameters, e.g change:

```
         self.meek_env = {
             "TOR_PT_MANAGED_TRANSPORT_VER": "1",
             "TOR_PT_CLIENT_TRANSPORTS": "meek",
         }
```

to this:

```
         self.meek_env = {
             "TOR_PT_MANAGED_TRANSPORT_VER": "1",
             "TOR_PT_CLIENT_TRANSPORTS": "m33k",
         }
```

...you'll see:

```
[Mar 22 2022 02:18:50 PM] AutoConnectTab.use_bridge_connect_clicked: Trying to automatically obtain bridges
[Mar 22 2022 02:18:50 PM] Meek.start: Starting meek client
[Mar 22 2022 02:18:50 PM] Meek.cleanup
[Mar 22 2022 02:18:50 PM] MeekNotRunning.__init__: Meek experienced an error starting up: CMETHOD-ERROR m33k no such method

[Mar 22 2022 02:18:50 PM] AutoConnectTab._got_no_bridges: Could not obtain bridges, so falling back to trying built-in obfs4 bridges
```

We also didn't need to catch `MeekNotRunning` in `CensorshipCircumvention`'s init constructor because Meek will have already started by then, and will have itself raised that exception if it couldn't.

### DIFF
--- a/cli/onionshare_cli/censorship.py
+++ b/cli/onionshare_cli/censorship.py
@@ -19,8 +19,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import requests
 
-from .meek import MeekNotRunning
-
 
 class CensorshipCircumventionError(Exception):
     """
@@ -47,15 +45,12 @@ class CensorshipCircumvention(object):
         self.api_proxies = {}
         if meek:
             self.meek = meek
-            if not self.meek.meek_proxies:
-                raise MeekNotRunning()
-            else:
-                self.common.log(
-                    "CensorshipCircumvention",
-                    "__init__",
-                    "Using Meek with CensorshipCircumvention API",
-                )
-                self.api_proxies = self.meek.meek_proxies
+            self.common.log(
+                "CensorshipCircumvention",
+                "__init__",
+                "Using Meek with CensorshipCircumvention API",
+            )
+            self.api_proxies = self.meek.meek_proxies
         if onion:
             self.onion = onion
             if not self.onion.is_authenticated:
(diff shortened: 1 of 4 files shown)


SaptakS (approved):
The logging looks good to me. I agree that `MeekNotFound` might not be needed to be shown as an error banner in the UI given we have a fall back. I am however planning to show some kind of success message after connecting, which can have a message like "Connected successfully via built-in obfs4 bridge" or something.

micahflee (approved):
This looks good to me. I also agree `MeekNotFound` isn't needed. In fact, the only reason that error should ever show up is if the user is running from source and didn't set up their dev environment correctly. If they use the Windows, Mac, or Linux packages, they should all be bundled with meek.

PR #1560, opened 2022-03-27
Remove meek workaround in Censorship Circumvention API's built-in bridges response

mig5:
TPO recently took their Censorship Circumvention API 'into production' - see the comms around https://gitlab.torproject.org/tpo/anti-censorship/bridgedb/-/issues/40025#note_2790866

As part of this, it looks like they started 'providing real bridges' in the response (even though we observed bridges returned from the service seemed to work already).

But another side-effect, was that previously, the data coming back for the built-in Meek bridges, was in an incorrect format (coming back with key `meek` instead of `meek-azure`, and not setting `meek-lite` in the connection parameters) and/or lacking necessary values (no url/front suffix), and so we put a workaround in place for that, on our side.

Now it looks like they've fixed the response to return valid meek-lite built-in bridges, and that ironically broke our code:

```
  File "/home/user/git/onionshare/desktop/onionshare/tor_connection.py", line 288, in run
    self.parent.onion.connect(self.settings, False, self._tor_status_update)
  File "/home/user/git/onionshare/cli/onionshare_cli/onion.py", line 669, in connect
    self.update_builtin_bridges()
  File "/home/user/git/onionshare/cli/onionshare_cli/onion.py", line 957, in update_builtin_bridges
    if builtin_bridges["meek"]:
KeyError: 'meek'
```

Removing our workaround so that it Just Works...

### DIFF
--- a/cli/onionshare_cli/onion.py
+++ b/cli/onionshare_cli/onion.py
@@ -954,20 +954,6 @@ class Onion(object):
                 "update_builtin_bridges",
                 f"Obtained bridges: {builtin_bridges}",
             )
-            if builtin_bridges["meek"]:
-                # Meek bridge needs to be defined as "meek_lite", not "meek",
-                # for it to work with obfs4proxy.
-                # We also refer to this bridge type as 'meek-azure' in our settings.
-                # So first, rename the key in the dict
-                builtin_bridges["meek-azure"] = builtin_bridges.pop("meek")
-                new_meek_bridges = []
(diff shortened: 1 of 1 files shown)


micahflee (approved):

micahflee:
Hah, I was just setting up a new Windows 11 dev environment and ran into this issue, and fixed it, but didn't make a PR yet for it and you beat me to it. But this is great.

PR #1563, opened 2022-03-28
Add shortcuts for modes and new/close tab. Set accessible string for modes on mode chooser

### LINKED ISSUE #1558: Accessibility: Add Labels To Buttons That Don't Have Them
Hello,
Don't know about other platforms, but on Windows some of the buttons are missing real text labels. Adding them would be especially helpful to screen reader users, because without them screen reading software is not able to inform the user what those buttons are for.
If I'm not mistaken, there are 6 buttons in the main window when you open the main GUI -  Share, Receive, Website, Chat, Settings, Tor Settings. None of them have true labels though.
The buttons on the tabs themselves seem generally fine. That being said the Share, Receive and Website tabs do seem to contain at least one unlabelled button each. In tab order those unlabelled buttons always come after the list of open tabs. So for example, if one's currently focused on the tab switcher with a Receive tab open, pressing the tab key will bring focus to an unlabelled button.
Many thanks.

### LINKED ISSUE #1559: Add Keyboard Shortcuts For Closing Open Tabs
It would be nice if one could close open tabs by pressing keyboard shortcuts - especially Share, Receive, Website and Chat. I think that Settings and Tor Settings don't necessarily have to be closable via a keyboard shortcut, since there the Save button acts like a close button as well, but on the other hand, it could make sense for consistency reasons.
On Windows I would suggest CTRL+W or CTRL+F4. Either of these would be fine. These are the most intuitive shortcuts I can think of for this task - most programs use either the one or the other for this very purpose.

mig5:
May close #1558
May close #1559

There are now these keyboard shortcuts:

Ctrl T - New Tab
Ctrl X - Closes current tab

And from the main mode chooser screen:

Ctrl S - Share mode
Ctrl R - Receive mode
Ctrl W - Website mode (this is why I chose X for close tab, rather than W which is what browsers use)
Ctrl C - Chat mode

Ctrl H - Settings tab

I also set the [`setAccessibleName()`](https://doc.qt.io/qt-5/qwidget.html#accessibleName-prop) on the QPushButtons on the main screen, but I am not sure how to test it.

### DIFF
--- a/desktop/onionshare/main_window.py
+++ b/desktop/onionshare/main_window.py
@@ -119,6 +119,7 @@ class MainWindow(QtWidgets.QMainWindow):
                 )
             )
         )
+        self.settings_button.setAccessibleName(strings._("gui_settings_window_title"))
         self.settings_button.clicked.connect(self.open_settings)
         self.settings_button.setStyleSheet(self.common.gui.css["settings_button"])
         self.status_bar.addPermanentWidget(self.settings_button)
--- a/desktop/onionshare/main_window.py
+++ b/desktop/onionshare/main_window.py
@@ -119,6 +119,8 @@ class MainWindow(QtWidgets.QMainWindow):
                 )
             )
         )
+        sequence = QtGui.QKeySequence(QtCore.Qt.CTRL + QtCore.Qt.Key_H)
+        self.settings_button.setShortcut(sequence)
         self.settings_button.setAccessibleName(strings._("gui_settings_window_title"))
         self.settings_button.clicked.connect(self.open_settings)
         self.settings_button.setStyleSheet(self.common.gui.css["settings_button"])
--- a/desktop/onionshare/tab/tab.py
+++ b/desktop/onionshare/tab/tab.py
@@ -38,11 +38,17 @@ from ..widgets import Alert
 
 
 class NewTabButton(QtWidgets.QPushButton):
-    def __init__(self, common, image_filename, title, text):
+    def __init__(self, common, image_filename, title, text, shortcut):
         super(NewTabButton, self).__init__()
         self.common = common
 
         self.setFixedSize(280, 280)
+
+        # Keyboard shortcut, using the first letter of the mode
+        sequence = QtGui.QKeySequence(QtCore.Qt.CTRL + shortcut)
+        self.setShortcut(sequence)
+
+        self.setAccessibleName(title)
 
         # Image
         self.image_label = QtWidgets.QLabel(parent=self)
@@ -140,6 +146,7 @@ class Tab(QtWidgets.QWidget):
             "images/{}_mode_new_tab_share.png".format(self.common.gui.color_mode),
             strings._("gui_new_tab_share_button"),
             strings._("gui_main_page_share_button"),
+            QtCore.Qt.Key_S,
         )
         self.share_button.clicked.connect(self.share_mode_clicked)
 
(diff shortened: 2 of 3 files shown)


mig5:
We may need to highlight the first letter of each mode on the main screen's QPushButtons to make it apparent that there is a shortcut...

SaptakS:
I think adding shortcuts are good for the different mode selection in main window, but to make it fully accessible, all the actionable items should be focusable by tab. This PR in itself looks okay to me and definitely an improvement, but like I have mentioned before we should also do a more in-depth accessibility testing.

MatejGolian:
@SaptakS, I fully agree - ideally every actionable item should be focusable by tabbing through the interface.
I'll gladly try and help you programming guys as much as I can, but since I'm a screen reader user myself, I may not be able to always determine what needs to be done. For example, if a particular control is not included in the tab order, it's hard to say whether there are any accessibility issues with it (apart from it not being focusable).
As for this PR, it looks good, but how would I go about testing it in action - since there's no build generated for this PR?
Perhaps the quickest way of testing how accessible a given program is on Windows, is turning on Windows Moderator and than just tabbing through the interface. On more recent Windows versions (I believe starting with Windows 8), the shortcut for toggling it on/off is CTRL+Win+Enter.

SaptakS:
@MatejGolian I am going to test it using VoiceOver in MacOS which is also a widely uused screenreader for Mac users. I use it for testing accessibility in websites. I will try to ensure the things that you have mentioned as issues are resolved using VoiceOver.

MatejGolian:
Thanks, @SaptakS, I would test it on Windows, but I don't know how to obtain any builds besides the release ones. Another project I follow automagically generates builds with every PR - that way everything can be easily and immediately tested by anyone who wants. They use AppVeyor for the automated builds. Any chance OnionShare could do something similar?

mig5:
Interestingly, the `setAccessibleName()` seems to be causing OnionShare not to start properly on Qubes, at least most of the time.

The systray appears but the rest of the app doesn't, and I have to kill it. But if I start OnionShare with `QT_ACCESSIBILITY=0` environment variable set (or if I remove the `setAccessibleName()` changes, but leave in the shortcuts), it's consistently ok.

I don't know much about this stuff but I believe Qubes doesn't properly support certain accessibility/dbus features that might explain this.

Interested to hear if Micah/Saptak reproduce that on Qubes.

micahflee:
> Another project I follow automagically generates builds with every PR - that way everything can be easily and immediately tested by anyone who wants. They use AppVeyor for the automated builds. Any chance OnionShare could do something similar?

@MatejGolian We don't currently do this, but I wonder how possible it is. The Windows binary obviously wouldn't be codesigned, but I could see this being very useful. I'll open an issue for it.

SaptakS (approved):
I have tested the code with VoiceOver screen reader in MacOS and seems to work as expected. I am not sure about what the solution would be for Qubes.

SaptakS:
@mig5 I think https://github.com/QubesOS/qubes-issues/issues/5907 might be a relevant issue of why adding accessible names is causing this. Maybe we can just document somewhere that for using in Qubes you need to set `QT_ACCESSIBILITY=0 ` environment variables?

SaptakS:
@micahflee @mig5 is this ok to be merged? I don't know if there is any better solution for the Qubes OS issue

mig5:
Yes I think it's fine. I don't think this specific PR actually caused the issue either, because I still reproduced it on another branch I think, since.

PR #1564, opened 2022-03-28
Replace maps with flags

### LINKED ISSUE #1548: Add flags for country dropdown
Even though the map looked cool, it's a little bit of a nuance for the UI so had to be removed. So as suggested by @glenn-sorrentino we can replace the maps with flags. Suggested UI:

image

SaptakS:
Fixes #1548 

- Adds country flags and a script to retrieve flags
- Also replaces the try again button with a separate radio button as suggested by @glenn-sorrentino 

I still think the bridge window's text might need some reconciling to make things more clear. The error banner seems a little tricker to implement than I expected.

This is what it looks like right now:
Screenshot from 2022-03-30 20-51-54


mig5 (approved):
This looks good to me. Part of me feels like indeed it is a lot of 'text' to display, but indeed if the banner is hard to implement (I can't think of how to do it either), this is maybe good enough for now.

I definitely like the 'Try again without a bridge' as a radio box rather than its own button, that's great. 

The flags are also good, though like the rocket, I still am not 100% sold on how much value they give (all the French territories listed have a French flag same as France, so it isn't really 'distinguishable' to me, but at least it's colourful :) )

Note this branch is a bit behind `develop` so there is the meek error after connecting - but it doesn't get in the way of testing the change, so :+1: from me, nice work @SaptakS

PR #1569, opened 2022-03-31
Initial documentation for the censorship circumvention/autoconnect screen

mig5:
I took an initial pass at writing some documentation for the censorship circumvention/autoconnect screen. The file in question is `docs/source/connecting.rst`

The screenshots should be replaced later with consistent Ubuntu ones to match the other screenshots (IMO) - also, we'll probably want ones that show the country flags.

I'm not a natural at technical documentation like Micah - please feel free to improve upon it!


docs/source/tor.rst
```diff
@@ -1,7 +1,9 @@
 Connecting to Tor
 =================
 
-Pick a way to connect OnionShare to Tor by clicking the Tor onion icon in the bottom right of the OnionShare window to open the Tor Settings tab.
+.. _tor:
+
+Pick a way to connect OnionShare to Tor by clicking Network Settings from the welcome screen, or the Tor onion icon in the bottom right of the OnionShare window to open the Tor Settings tab.
```
SaptakS, line 6: Now there is no longer a separate Tor onion icon in the bottom right. Now everyone needs to click the settings icon and then go to the "Tor Settings" subtab.

docs/source/connecting.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Before clicking 'Use a Bridge', you can choose either 'Automatically determine my country from my IP address' or 'Manually select my country'.
+
+If you choose the former option, the Censorship Circumvention API will consider your IP address (yes, your real IP address) to determine what country you might reside in.
+
+Based on the country information, the API will try to automatically find bridges that suit your location.
+
+.. image:: _static/screenshots/autoconnect-trying-to-resolve-connectivity-issues.png
+
+If you choose to manually select a country, OnionShare will still make a request to the Censorship Circumvention API, but will tell the API the country code. The API will then try to fetch bridges for that country that you specified.
```
SaptakS, line 36: Maybe we can mention here that in this scenario the API doesn't check your real IP address which makes it a little better for some folks.

docs/source/connecting.rst
```diff
@@ -0,0 +1,59 @@
...
+
+If it finds any such bridges, OnionShare will try to reconnect to Tor using those bridges.
+
+If the API does not find any bridges for your location, OnionShare will ask the API for 'fallback' options. At the time of writing, this is likely to be the obfs4 built-in bridges.
+
+OnionShare will also attempt to use the obfs4 built-in bridges if for some reason it could not connect to the API itself, or the API returned an error.
+
+It's important to note that the requests to the Censorship Circumvention API do not go over the Tor Network (because if you could connect to Tor already, you wouldn't need to connect to the API).
+
+Even though it is hard for an adversary to discover where the Meek request is going, this may still be risky for some users. Therefore, it is an opt-in feature. The use of Meek and non-torified network requests are limited only to making one or two requests to the Censorship Circumvention API. Then Meek is stopped, and all further network requests happen over the Tor network.
```
SaptakS, line 46: Maybe we can reiterate over here that they can always go to the network settings and put their own bridge settings, if they think Meek requests are risky for them. The meek requests are just to make life easier for more Tor users, but if someone has high risk threat models, they should preferably use network settings.

SaptakS (commented):
I have added some thoughts for the docs. Also a lot of the particulars in the `connecting.rst` docs will change once https://github.com/onionshare/onionshare/pull/1564 is merged

docs/source/connecting.rst
mig5: The thing is that the API knows the IP address either way, it just doesn't evaluate it against a geoIP database. We have to trust that the TP's API doesn't log the IP or do anything else with it, but we can't prove it, so I don't want to give any misleading assurance

SaptakS: Agreed

SaptakS:
I think this looks good to me. We will need to update the screenshots to the more recent UI for the bridge connection tab at least. Also, it seems like @micahflee usually generates the `.po` files separately. Not sure what the workflow is right now for docs.

micahflee:
Nice, thanks @mig5. I'm working on editing this, and replacing the screenshots with new ones in Ubuntu (and installing Ubuntu 22.04 Jammy Jellyfish :)) in this branch: https://github.com/onionshare/onionshare/tree/censorship_docs

PR #1581, opened 2022-04-30
Opens new tab if the only other tab is a settings tab

### LINKED ISSUE #1578: If setting tab is open, New Tab is not opened on successfully connecting to Tor
While initial connecting to Tor, if the user opens the settings tab and leaves it open (doesn't close or click on save) and clicks on Connect to Tor, then upon successfully connecting to Tor, the autoconnect tab closes, but a New Tab with the different options doesn't open.

SaptakS:
Fixes #1578 

This PR ensures that after a tab is closed, if there is still one tab open and that tab is a settings tab, then open a new tab. It does that both for after successful connection is made, and in general later when using OnionShare. I think it preserves the user's experience like before and they are not confused about why did no new tab come and what to do now.

### DIFF
--- a/desktop/onionshare/tab_widget.py
+++ b/desktop/onionshare/tab_widget.py
@@ -314,6 +314,11 @@ class TabWidget(QtWidgets.QTabWidget):
             self.common.settings.set("persistent_tabs", persistent_tabs)
             self.common.settings.save()
 
+    def has_no_onionshare_tab(self):
+        # Check if the last tab is closed, or there is only one tab
+        # which is the settings tab
+        return self.count() == 0 or (self.count() == 1 and type(list(self.tabs.values())[0]) is SettingsParentTab)
+
     def close_tab(self, index):
         self.common.log("TabWidget", "close_tab", f"{index}")
         tab = self.widget(index)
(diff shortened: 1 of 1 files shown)


SaptakS:
I wonder why the test-gui keeps failing with segmentation fault but on rerunning it works. 🤔  Happening almost in all PRs

mig5:
I thought the segfault failing was the 'GUI app doesn't close properly' issue but I could be wrong. It seems to happen at the *end* of all the tests having (successfully) run.

mig5 (approved):
Looks good. The only weird quirk is that the effect feels like you can't 'close' the 'New Tab' via its 'X', because it's instantly creating a new tab in its place. But I guess that was the case already when there's only one 'New Tab' open (before a mode is selected).

And it's more important that the bug mentioned is fixed (which it is!)

SaptakS:
> I thought the segfault failing was the 'GUI app doesn't close properly' issue but I could be wrong. It seems to happen at the _end_ of all the tests having (successfully) run.

It doesn't happen always. Sometimes I have just re-run the CircleCI test and it just passes.

SaptakS:
> Looks good. The only weird quirk is that the effect feels like you can't 'close' the 'New Tab' via its 'X', because it's instantly creating a new tab in its place. But I guess that was the case already when there's only one 'New Tab' open (before a mode is selected).

I agree. But that was kind of always there. It's more prominent now if a settings tab is open. The only two options are either closing the last tab closes the window (whether that be new tab or settings tab or any other tab). Or this logic that we currently have of re-opening a new tab, maybe?

SaptakS:
@mig5 like I said, it passed now. So not really sure.

micahflee (approved):

PR #1582, opened 2022-04-30
Better error logs in the bridge connect widget

### LINKED ISSUE #1547: Better UI error messages in the connection flow

### LINKED ISSUE #1577: Onionshare connecting using Tor Browser shows wrong error
When a user chooses "Attempt auto-configuration with Tor Browser" and if they don't have Tor Browser running, and they click on "Connect to Tor", it takes them to the censorship circumvention tab. 

But in this case, we know the exact reason of the fail and in the logs we do show "Could not connect to the Tor controller. Is Tor Browser (available from torproject.org) running in the background?". So we should show the same error in UI so the user knows exactly what the issue is and how they might be able to fix it and not take them to the censorship tab.

A similar scenario happens if the packaged tor is missing. Though that won't happen for most users since it will already be packaged with OnionShare.

I guess we should have a generic error page which is for issues that we can absolutely be sure is not censorship related, but because something is missing.

SaptakS:
Fixes #1577 
Fixes #1547 

This PR is doing a few different things:
- If cancel is clicked when in the firstlaunchwidget, it doesn't go to the bridge widget but stays in the firstlaunchwidget. I think this makes sense, since cancel is not exactly a connection fail. (Discussion: https://github.com/onionshare/onionshare/issues/1566#issuecomment-1107424215)
- Before in the bridge connection widget, the first error message we always showed was "Are you connected to internet?". However, we are already sending better tor connection error messages giving various different reasons. For example, like I mentioned in #1577 , if someone is trying to connect using Tor browser, but they didn't start Tor browser, the tor connection thread returns an appropriate message for that. So, I have updated the widget to show the error message returned from the tor connection thread as the first error line, and then kept the second line as still suggesting that it might be a censorship issue, and you might need to use a bridge. (I think we can go a step further, and determine which errors are definitely not censorship errors, and which might be because of censorship, e.g. timeout maybe? But I wasn't sure if there is a way we can know which all errors, or if any specific error happens if it is a censorship related error).
- I have also put the radio options inside a QGroupBox. I feel this should make the options a little more accessible since screen reader users (and even other users) will be able to exactly identify what are they chosing those radio buttons for.

Screenshot from 2022-04-30 22-00-17


desktop/onionshare/connection_tab.py
```diff
@@ -326,11 +326,20 @@ def tor_con_fail(self, msg):
         """
         self.tor_con.hide()
 
-        # If we're on first launch, switch to use bridge
+        # If there is a message, update the text of the bridge widget
+        if msg:
+            self.use_bridge_widget.connection_error_message.setText(f"{msg}")
```
mig5, line 331: Is `(f"{msg}")` typical of best practice (given there's no mix of strings being interpolated, I would normally just do `(msg)` but maybe this is not right)

desktop/onionshare/connection_tab.py
```diff
@@ -326,11 +326,20 @@ def tor_con_fail(self, msg):
...
+
+        # If we're on first launch, check if wasCancelled
+        # If cancelled, stay in first launch widget and show buttons
```
mig5, line 334: Tiny nitpick but the method is named `wasCanceled()` (not `wasCancelled()`), I guess we should be consistent with US spelling even though I prefer the two l's :)

mig5 (approved):
This is working great! Confirmed that the proper error messages are appearing.

I also realise I misunderstood your question about the 'Cancel' scenario. What you've done here does make more sense, although it does make it a little bit harder for us to test the censorship circumvention if we aren't ourselves actually facing censorship.

Other than my nitpick comments I've left, LGTM.

desktop/onionshare/connection_tab.py
SaptakS: I was actually doing some string interpolation, which I then removed, but forgot to remove the f-string format. My mistake.

SaptakS:
> I also realise I misunderstood your question about the 'Cancel' scenario. What you've done here does make more sense, although it does make it a little bit harder for us to test the censorship circumvention if we aren't ourselves actually facing censorship.

I always used to change the timeout value in `cli/onionshare_cli/onion.py` to something very low.

micahflee (approved):
This is great.

Issue #1588, opened 2022-05-30
Crash when using the Censorship API


micahflee:
I came across is bug while taking screenshots for the documentation. I disconnected from the internet, opened OnionShare, and tried to connect. Once it failed, I reconnected to the internet and then clicked "Automatically determine my country from my IP address for bridge settings".

```
[May 30 2022 01:42:31 PM] Settings.__init__
[May 30 2022 01:42:31 PM] Settings.load
[May 30 2022 01:42:31 PM] Onion.__init__
[May 30 2022 01:42:31 PM] GuiCommon.get_tor_paths: using paths in resources
[May 30 2022 01:42:31 PM] Meek.__init__
[May 30 2022 01:42:31 PM] GuiCommon.get_tor_paths: using paths in resources
[May 30 2022 01:42:31 PM] MainWindow.__init__
[May 30 2022 01:42:31 PM] TabWidget.__init__
[May 30 2022 01:42:31 PM] EventHandlerThread.__init__
[May 30 2022 01:42:31 PM] TabWidget.open_connection_tab
[May 30 2022 01:42:31 PM] EventHandlerThread.run[May 30 2022 01:42:31 PM] AutoConnectTab.__init__

[May 30 2022 01:42:31 PM] Settings.__init__
[May 30 2022 01:42:31 PM] Settings.load
[May 30 2022 01:42:31 PM] AutoConnectFirstLaunchWidget.__init__
[May 30 2022 01:42:31 PM] AutoConnectUseBridgeWidget.__init__
[May 30 2022 01:42:31 PM] TorConnectionWidget.__init__
[May 30 2022 01:42:31 PM] TabWidget.tab_changed: Tab was changed to 0
[May 30 2022 01:42:31 PM] AutoConnectTab.autoconnect_checking
[May 30 2022 01:44:09 PM] AutoConnectTab.first_launch_widget_connect_clicked
[May 30 2022 01:44:09 PM] TorConnectionWidget.start
[May 30 2022 01:44:09 PM] TorConnectionThread.__init__
[May 30 2022 01:44:09 PM] TorConnectionThread.run
[May 30 2022 01:44:09 PM] Onion.connect: connection_type=bundled
[May 30 2022 01:44:09 PM] Onion.connect: tor_data_directory_name=/home/micah/.config/onionshare/tor_data
[...]

PR #1589, opened 2022-05-30
Censorship documentation

### LINKED ISSUE #1544: Add censorship circumvention documentation

micahflee:
Fixes #1544 and #1588, and replaces #1569 (it's based off of it).

- While taking screenshots for the docs I discovered a new bug, #1588, which I fixed in this branch. It looks like this bug was caused by a change on the Censorship Circumvention API side, and our code worked before that change. Now it should be less brittle.
- I also updated the poetry dependencies for the `docs` folder, which really aren't very important except one of them had a vulnerability and github was telling me about it.
- And I updated PySide2 from version 5.15.2 to 5.15.2.1, because this version works with Python 3.10, and the old version does not. Ubuntu 22.04 comes with Python 3.10, so this is the only way to get poetry deps to install in this newer OS.
- I updated the get-tor scripts to use the latest Tor Browser.

In this PR, I combined the old Tor settings page ("Connecting to Tor") with the new censorship circumvention page ("Getting connected to Tor") into just one page, and moved it up near the top of the contents.

I also updated all of the screenshots throughout all of the docs to use new ones, taking in Ubuntu 22.04.


mig5 (approved):
I built and checked the Connecting to Tor docs locally from this branch. Reads well! Thanks for the improvements. Let's merge it

PR #1616, opened 2022-10-02
Advanced Usage documentation fix

comradekingu:


docs/source/advanced.rst
```diff
@@ -25,8 +25,9 @@ Turn Off Private Key
 
 By default, all OnionShare services are protected with a private key, which Tor calls "client authentication".
 
-The Tor Browser will ask you to enter your private key when you load an OnionShare service.
-If you want allow the public to use your service, it's better to disable the private key altogether.
+The Tor Browser will ask you to enter your private key when you want to offer an OnionShare service.
```
comradekingu, line 28: Is this done for just authentication, or also encryption.

docs/source/advanced.rst
```diff
@@ -25,8 +25,9 @@ Turn Off Private Key
...
+If you want to allow the public to use your service, it's better to turn off the private key so that people you don't know
```
comradekingu, line 29: This is not good.

docs/source/advanced.rst
mig5: Authentication. Tor's onion service protocol itself takes care of encryption.

docs/source/advanced.rst
mig5: I don't understand the comment.

docs/source/advanced.rst
mig5: I don't think the proposed change is accurate, actually. Private key is prompted by the browser when visiting (not 'offering') an OnionShare service. The original sentence is correct.

docs/source/advanced.rst
```diff
@@ -25,8 +25,9 @@ Turn Off Private Key
...
+wont have to find your public key.
```
mig5, line 30: Better would be 'so that people won't have to enter it'. Also not `wont` ought to be `won't`.

mig5 (changes requested):
A few tweaks required before this can be merged.

PR #1625, opened 2022-10-10
restore the primary_action mode settings in a tab after OnionShare reconnects to Tor

### LINKED ISSUE #1506: Mode settings in open tabs disappear if Tor Connection settings are changed
If you have any mode tabs open (but without any running shares), and then go to TorSettingsTab and change your Tor connection settings (e.g enable a bridge, or stop using bridges), then once the TorSettingsTab closes and your other tabs become 'viewable' again, the per-mode settings (e.g Client Auth/public mode, autostop/autostart etc) are all missing. The only widget that is available is the green Start button.

In most cases this is probably ok if you were confident that you didn't want to change settings. But it's a blocker if the tab had previously used autostart or autostop and the 'time' has passed - you won't be able to start that share again and you can't change the time.

We need to make the mode settings widgets viewable again once the tab can be active again.

mig5:
This should fix #1506

The modes call a `handle_tor_broke_custom()` when Tor settings get changed and Tor restarts, in the TorSettingsTab. That function involves hiding the primary action (which contains the mode settings widget). It's a bit moot to do so since the entire widget is hidden in favour of the 'tor_not_connected' widget, these days, but there still might be other scenarios where a disconnection from Tor is important to hide the primary nav.

Anyway, let's just re-show the primary_settings widget once Tor is reconnected after the TorSettingsTab closes.

To test:

1) Open a 'Receive mode' tab, but don't start the service
2) Go to the Tor Settings Tab and change the Tor connection mode (e.g enable obfs4 bridges or similar), to invoke a Tor reconnection
3) Once the Tor Settings Tab closes with reconnected Tor, check that the Receive mode's settings are still visible in the Receive Mode tab. (In current main branch, they are absent, and all that is present is the 'Start' button)

### DIFF
--- a/desktop/onionshare/tab/mode/__init__.py
+++ b/desktop/onionshare/tab/mode/__init__.py
@@ -563,6 +563,7 @@ class Mode(QtWidgets.QWidget):
         """
         self.content_widget.show()
         self.tor_not_connected_widget.hide()
+        self.primary_action.show()
 
     def tor_connection_stopped(self):
         """


SaptakS (approved):
Looks good to me.

PR #1636, opened 2022-10-30
Upgrade from PySide2 to PySide6 for native compatibility with Apple Silicon

NoisyCoil:
# Features

- Upgraded from PySide2 to PySide6 to make OnionShare run natively on Apple Silicon
- Passes all tests on macos-aarch64, macos-x86_64 and linux64 (with pyside-unrelated warnings)
- MacOS bundle build scripts adapted to PySide6 and working on both Intel and Apple Silicon (though I did not try codesigning via the scripts)
- The meek, obfs4proxy and snowflake scripts build universal binaries on MacOS (Tor Browser's tor is universal since v12.0)

# To check/to be modified

- Universal binary builds should be disabled for hosts which cannot build the arm64 binaries (currently they are not, which will lead to errors if the build is attempted, say, by a developer working on an Intel MacOS with outdated tools)
- If the Python installation used to build OnionShare on MacOS is universal, then a universal OnionShare should result from the build process. Currently I do not have a universal installation, so I cannot check that this is indeed the case. However, I can see that all but a couple of inessential setuptools binaries and the entry binary built by cx_Freeze are not universal in the MacOS bundle (I added a script in desktop/scripts to check this). It looks like the entry binary will be universal as well if cx_Freeze is installed as a universal package, which should be the case for universal Python installations

# Missing

- Package build system not tested on Linux (nor Windows)
- Nothing tested on Windows (the win build scripts still contain the PySide2 code)


micahflee:
Thank you for submitting this! Sorry I've been quite busy and putting OnionShare on the backburner, but I'll start reviewing this PR now.

PR #1638, opened 2022-11-23
Ensure Tab timer starts once tor is connected

### LINKED ISSUE #1637: Changing the Tor connection stops the Tab's event timer and nothing starts it back up after Tor regains connection
Start a share in 'Receive Mode'

Upload something, and you'll see the history widgets get added.

Stop the share but don't remove the tab.

Go to Settings and do something that stops the Tor connection, like change the Tor connection mode (e.g, enable built-in obfs4 bridges). Let Tor reconnect.

Back in the same tab, start Receive mode again.

Upload something to that same onion share, and there'll be no history item widgets appear.

As far as I can tell, the `handle_request_started()` method of the Receive Mode tab is not firing (I added some log debug statements there and they do not fire when the onion is started back up).

The `handle_request_started()` method is fired by the `Tab` object in its `timer_callback()` method:

```
        if not self.common.gui.local_only:
            # Have we lost connection to Tor somehow?
            if not self.common.gui.onion.is_authenticated():
                self.timer.stop()
                self.status_bar.showMessage(strings._("gui_tor_connection_lost"))
                self.system_tray.showMessage(
                    strings._("gui_tor_connection_lost"),
                    strings._("gui_tor_connection_error_settings"),
                )
                self.get_mode().handle_tor_broke()

        # Process events from the web object
        mode = self.get_mode()

        events = []

        done = False
        while not done:
            try:
                r = mode.web.q.get(False)
                events.append(r)
            except queue.Empty:
                done = True

        for event in events:
            if event["type"] == Web.REQUEST_LOAD:
                mode.handle_request_load(event)

            elif event["type"] == Web.REQUEST_STARTED:
                mode.handle_request_started(event)
```

I think what is happening is this part:

```
            # Have we lost connection to Tor somehow?
            if not self.common.gui.onion.is_authenticated():
                self.timer.stop()
```

Back in OnionShare 2.4 we used to trigger a `settings_have_changed()` throughout the objects when the Tor connection was re-established. 

It looks like we still have `settings_have_changed()` in the Tab object, which would start the timer back up:

```
    def settings_have_changed(self):
        # Global settings have changed
        self.common.log("Tab", "settings_have_changed")

        # We might've stopped the main requests timer if a Tor connection failed. If we've reloaded
        # settings, we probably succeeded in obtaining a new connection. If so, restart the timer.
        if not self.common.gui.local_only:
            if self.common.gui.onion.is_authenticated():
                mode = self.get_mode()
                if mode:
                    if not self.timer.isActive():
                        self.timer.start(500)
                    mode.on_reload_settings()
```

but nothing connects to it:

```
onionshare/main_window.py:    def settings_have_changed(self):
onionshare/main_window.py:            tab.settings_have_changed()
onionshare/settings_tab.py:    def settings_have_changed(self):
onionshare/tor_settings_tab.py:    def settings_have_changed(self):
```

It looks like we lost it when we implemented the AutoStart widget, because previously, as soon as the Tor Connection Dialog started up when starting OnionShare, the MainWindow object did this:

```
        # Start the "Connecting to Tor" dialog, which calls onion.connect()
        tor_con = TorConnectionDialog(self.common)
        tor_con.canceled.connect(self.tor_connection_canceled)
        tor_con.open_settings.connect(self.tor_connection_open_settings)
        if not self.common.gui.local_only:
            tor_con.start()
            self.settings_have_changed()
```

mig5:
Fixes #1637

Since we have a `tor_connection_started()` method in the Mode object, I am just starting the Tab object's timer there. 

Accordingly, I don't think the old `settings_have_changed()` methods are actually needed anymore.

Likewise for the stuff that happened on `on_reload_settings()` doesn't seem to be needed in Share/Website mode anymore (perhaps we are already re-showing those widgets once we allow those Mode tabs to be visible again once Tor connects)

But I appreciate some testing - all you have to do is start a share, stop it, change the Tor Connection mode (enable or disable builtin obfs4 bridges for example), start the share again, and make sure that history widgets appear again next time you visit that share.

### DIFF
--- a/desktop/onionshare/main_window.py
+++ b/desktop/onionshare/main_window.py
@@ -244,17 +244,6 @@ class MainWindow(QtWidgets.QMainWindow):
                 break
         self.tabs.open_settings_tab(from_autoconnect, active_tab=active_tab)
 
-    def settings_have_changed(self):
-        self.common.log("OnionShareGui", "settings_have_changed")
-
-        if self.common.gui.onion.is_authenticated():
-            self.status_bar.clearMessage()
-
-        # Tell each tab that settings have changed
-        for index in range(self.tabs.count()):
-            tab = self.tabs.widget(index)
-            tab.settings_have_changed()
-
     def bring_to_front(self):
         self.common.log("MainWindow", "bring_to_front")
         self.raise_()
--- a/desktop/onionshare/settings_tab.py
+++ b/desktop/onionshare/settings_tab.py
@@ -315,10 +315,6 @@ class SettingsTab(QtWidgets.QWidget):
 
         return settings
 
-    def settings_have_changed(self):
-        # Global settings have changed
-        self.common.log("SettingsTab", "settings_have_changed")
-
     def _update_autoupdate_timestamp(self, autoupdate_timestamp):
         self.common.log("SettingsTab", "_update_autoupdate_timestamp")
 
--- a/desktop/onionshare/tab/mode/__init__.py
+++ b/desktop/onionshare/tab/mode/__init__.py
@@ -564,6 +564,8 @@ class Mode(QtWidgets.QWidget):
         self.content_widget.show()
         self.tor_not_connected_widget.hide()
         self.primary_action.show()
+        if not self.tab.timer.isActive():
+            self.tab.timer.start(500)
 
     def tor_connection_stopped(self):
         """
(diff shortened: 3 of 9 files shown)


mig5:
Now that https://github.com/onionshare/onionshare/pull/1629 is passing (at least, it should, it is locally now), if that gets merged in I can rebase this to get the build tests to pass.

mig5:
Rebased!

SaptakS (approved):
tested it locally and looks good to me.

PR #1639, opened 2022-11-27
Add terminal QR code to CLI

0-ft:
Adds option to output a QR code of the share address to the terminal in the CLI.
Uses `qrcode` python lib, added as a poetry dependency.

### DIFF
--- a/cli/onionshare_cli/__init__.py
+++ b/cli/onionshare_cli/__init__.py
@@ -31,7 +31,7 @@ from .web import Web
 from .onion import TorErrorProtocolError, TorTooOldEphemeral, TorTooOldStealth, Onion
 from .onionshare import OnionShare
 from .mode_settings import ModeSettings
-
+from qrcode import QRCode
 
 def main(cwd=None):
     """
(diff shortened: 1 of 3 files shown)


mig5 (approved):
LGTM, seems to work great.

Tests are failing because the codebase is a bit out of date, rebasing would solve that, but I already tested that CLI tests still pass, locally.

mig5 (changes requested):
Actually, @0-ft , the GUI app also shows a separate QR code for the 'Private Key' for the .onion service as well, not just the URL.

It would be good to display two QR codes here - one for the URL as you've done, but another for the private key (`app.auth_string`). 

Do you want to have a go at that? Maybe if so, it would be good to rebase this branch off the latest `main` branch upstream, to get the builds/tests to pass cleanly too, but that's not so big a deal.

Thanks!

SaptakS:
+1 to what @mig5 said.

mig5:
I'm going to close this out in favour of https://github.com/onionshare/onionshare/pull/1773 (which rebases your work against latest `main` branch to fix the conflicts, and also adds the Private Key QR code).

Thanks for your contribution!

PR #1653, opened 2023-02-05
Pyside6 package

micahflee:
Making a new PR for #1636 so I can test CI.


mig5:
@micahflee interesting that the Flatpak build is failing - it's doing so at a line where I had to recently hack to get it to work, in prep for 2.6.1. 

See https://github.com/onionshare/onionshare/pull/1629/commits/766df4d2a311fcd27e5da0f7c23172639b6ff186 - you might want to revert that, maybe it's not needed anymore with your other updates here.

The discussion/context for that was in https://github.com/onionshare/onionshare/pull/1629

micahflee:
I haven't looked at the Flatpak packaging yet, but I think one thing that will definitely need to be changed is to install pyside6 instead of pyside2. That patch is part of installing pyside2, so I'll see if pyside6 has less issues.

desktop/scripts/kounek7zrdx745qydx6p59t9mqjpuhdf
mig5, line 1: @SaptakS do you know what this strange file is?

mig5: Nevermind - I see where it is used now!

mig5 (approved):

desktop/scripts/kounek7zrdx745qydx6p59t9mqjpuhdf
SaptakS: It's the public key received using WKD. I have stored it in the repo instead of trying to retrieve it from the internet to hopefully have some measure against MITM. Reasoning being, since this is stored locally, if the other files were MITM-ed, then the signature should fail.

PR #1656, opened 2023-03-11
Include trailing '/' on directory links in file list mode.

bavier:
This allows directory contents to properly reference included resources using relative links.

My use-case is sharing several photo-galleries, each generated using a static web-page gallery tool (`fgallery`), and each in its own directory.  I can add the entire collection and onionshare will helpfully create a directory listing for the main page.  But without the trailing `/` on the directory link, the `index.html` in each gallery is not able to locate its resources via relative references.  This patch fixes that.

### DIFF
--- a/cli/onionshare_cli/web/send_base_mode.py
+++ b/cli/onionshare_cli/web/send_base_mode.py
@@ -169,7 +169,7 @@ class SendBaseModeWeb:
 
             if is_dir:
                 dirs.append(
-                    {"link": os.path.join(f"/{path}", filename), "basename": filename}
+                    {"link": os.path.join(f"/{path}", filename, ""), "basename": filename}
                 )
             else:
                 size = os.path.getsize(this_filesystem_path)


mig5:
Finally got around to testing this (with multiple copies of https://github.com/onionshare/onionshare-website as subfolders in Website mode). I see what you mean! 

Unfortunately, whilst it fixes the issue for Website mode, it also breaks the regular 'Share' mode - those folders would previously be 'navigatable' (the index.html is ignored) but, with this fix, create a 404. These two modes use the same codebase. I'll have a think about a way to avoid that happening.

Thanks for the fix!

mig5:
We ended up taking this fix and adapting it to ensure it doesn't break Share Mode, in https://github.com/onionshare/onionshare/pull/1783

I couldn't push to your branch so I made a separate branch for it upstream.

Accordingly, I'll close this out.

Thanks for the contribution!

PR #1660, opened 2023-04-09
Fixing CI builds

SaptakS:
I have updated the pyproject.toml in cli to use poetry-core instead of poetry for build-systems.

I have also updated the code for pulling Tor Browser code. I have tried to make the code a little bit more dynamic. What I am doing right now is pulling the Tor Browser version from `https://aus1.torproject.org/torbrowser/update_3/release/downloads.json` and also the individual platform download URLs. Then using the version, I am downloading the [sha256sums file](https://dist.torproject.org/torbrowser/12.0.4/sha256sums-signed-build.txt) from the Tor Browser distribution link. This allows us to always be sure that the latest Tor Browser is downloaded. Because the distribution link becomes invalid once a new release is made.

The only thing I am a little worried is having the sha256 sum downloaded online. I am not sure how risky it is. In that case, we can probably keep the sha256sums hardcoded in the code. That will still make the CI fail, everytime a new tor browser release is made though. @micahflee @mig5 thoughts?

### DIFF
--- a/cli/pyproject.toml
+++ b/cli/pyproject.toml
@@ -40,5 +40,5 @@ pytest = ">=7.2.0"
 onionshare-cli = 'onionshare_cli:main'
 
 [build-system]
-requires = ["poetry>=0.12"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core"]
+build-backend = "poetry.core.masonry.api"
--- a/desktop/scripts/get-tor.py
+++ b/desktop/scripts/get-tor.py
@@ -9,29 +9,10 @@ import subprocess
 import requests
 import click
 
-torbrowser_version = "12.0"
-expected_win32_sha256 = (
-    "a9cc0f0af2ce8ca0d7a27d65c7efa37f6419cfc793fa80371e7db73d44b4cc02"
+torbrowser_version_url = (
+    "https://aus1.torproject.org/torbrowser/update_3/release/downloads.json"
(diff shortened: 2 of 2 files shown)


SaptakS:
Seems like the build_flatpack is still failing. Not really sure why, seems like a different issue altogether. Will investigate separately.

micahflee:
We can solve the riskiness factor the same way that the Tor Browser Launcher does. The sha256sums-signed-build.txt file has a detached signature, [sha256sums-signed-build.txt.asc](https://dist.torproject.org/torbrowser/12.0.4/sha256sums-signed-build.txt.asc).

You can hardcode the [Tor Browser Developers public key](https://github.com/micahflee/torbrowser-launcher/blob/main/share/torbrowser-launcher/tor-browser-developers.asc), and then use `gpg` to verify the sig for sha256sums-signed-build.txt, and if it verifies you can trust that the checksums are legit.

I guess the code can check to make sure `gpg` is available (it won't necessary be available automatically in some platforms, like Windows) and print a warning and skip verification if that's the case? But the important part is we can make sure `gpg` is installed in CI when building the releases.

SaptakS:
@micahflee sounds good to me. I have merged this branch into https://github.com/onionshare/onionshare/pull/1653 so that the CI issues can be resolved alongside updating to PySide6. I will add the above change that you mentioned in that branch itself.

PR #1664, opened 2023-05-21
Support Python 3.11

micahflee:
This upgrades pyside6 to 6.5.0.


mig5:
Tests are possibly failing because of this? (I get the following on debian 11 when I run ` PYTHONUNBUFFERED=1 QT_DEBUG_PLUGINS=1 poetry run pytest -v -s ./tests/test_gui_*.py`)

```
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: xcb, eglfs, wayland-egl, minimalegl, minimal, vnc, vkkhrdisplay, wayland, offscreen, linuxfb.

tests/run.sh: line 6:  2268 Aborted                 poetry run pytest -v -s tests/test_gui_tabs.py
```

Don't know if that's just something specific to my machine or not as the tests don't show this error (but they aren't executed in the same way)

I read https://forum.qt.io/topic/93247/qt-qpa-plugin-could-not-load-the-qt-platform-plugin-xcb-in-even-though-it-was-found/146 but none of the packages mentioned there (libxcb-xinerama0, libxcb-util, other such libxcb packages) worked for me - they are all already installed

mig5:
Ah - setting `QT_QPA_PLATFORM=offscreen` helps! (thanks https://github.com/therecipe/qt/issues/775)

PR #1677, opened 2023-05-23
Update deps & switch to waitress for running flask servers

### LINKED ISSUE #1389: Use nginx as the web service, instead of flask's app.run
Right now when OnionShare starts a web service, it uses `app.run` to start it directly in flask. I think instead we should launch an `nginx` subprocess and use that along with gunicorn to host the flask app.

- We already have an issue to use gunicorn (#1022) instead of flask directly, and this will be much simpler if we use nginx
- If we use nginx, we can get gzip for free and don't need to gzip everything and fill up `/tmp` (#1328), and I suspect we could simplify a lot of the web server code in other ways too
- If we use nginx, we get simple working range requests (#1333)
- If we use nginx and implement the download accelerator (#1295) we get much better performance, e.g. quicker download speeds

Before implementing this, I'm not sure how a few things will work:

- Progress bars. If someone downloads a file from nginx, how can we hook into the real-time file transfer to make progress bars work?
- "Stop sharing after files have been sent". How do we know that the files are finished sending?

### LINKED ISSUE #1542: Figure out new way to shutdown flask service
When we stop an OnionShare service, this `web.force_shutdown()` function is run:

```python
    def force_shutdown(self):
        """
        Stop the flask web server, from the context of the flask app.
        """
        # Shutdown the flask service
        try:
            func = request.environ.get("werkzeug.server.shutdown")
            if func is None and self.mode != "chat":
                raise RuntimeError("Not running with the Werkzeug Server")
            func()
        except Exception:
            pass

        self.running = False

        # If chat, shutdown the socket server
        if self.mode == "chat":
            self.socketio.stop()
```

I noticed when running tests we get this warning:

```
  /home/circleci/repo/cli/onionshare_cli/web/web.py:342: UserWarning: The 'environ['werkzeug.server.shutdown']' function is deprecated and will be removed in Werkzeug 2.1.
```

I looked into this, and they give their rationale for deprecating this here: https://github.com/pallets/werkzeug/issues/1752

Also, we should probably run Flask in production instead of dev.

### LINKED ISSUE #1614: Refactoring required to upgrade Flask
I attempted to upgrade Flask to the latest version, 2.2.2, but this causes a regression bug in OnionShare. The latest version of Flask that doesn't cause this regression is version 2.1.3, so I'm upgrading to that instead. Here's the bug.

In `cli/onionshare_cli/web/web.py`, after the Flask app is first created (`self.app`), the constructor runs `self.generate_static_url_path()`, which generates a new random URL for the static URL path. This works in the CLI, and work for the first time in the desktop app.

But if you open a tab, start a service, stop the service, then start it again, it crashes with:

```
[Sep 30 2022 02:42:17 PM] Web.generate_static_url_path: new static_url_path is /static_oc2qd4qj3exlekq5ryb63dkiwq
  File "/home/micah/code/onionshare/desktop/onionshare/threads.py", line 66, in run
    self.mode.web.generate_static_url_path()
  File "/home/micah/code/onionshare/cli/onionshare_cli/web/web.py", line 200, in generate_static_url_path
    self.app.add_url_rule(
  File "/home/micah/.cache/pypoetry/virtualenvs/onionshare-BFLClbuE-py3.10/lib/python3.10/site-packages/flask/scaffold.py", line 49, in wrapper_func
    self._check_setup_finished(f_name)
  File "/home/micah/.cache/pypoetry/virtualenvs/onionshare-BFLClbuE-py3.10/lib/python3.10/site-packages/flask/app.py", line 722, in _check_setup_finished
    raise AssertionError(
AssertionError: The setup method 'add_url_rule' can no longer be called on the application. It has already handled its first request, any changes will not be applied consistently.
Make sure all imports, decorators, functions, etc. needed to set up the application are done before running it.
```

This is because the desktop code tries calling `web.generate_static_url_path()` after the app has already been set up, and there's a check to prevent this from working in newer versions of Flask. Specifically, the `_check_setup_finished` method is new:

https://github.com/pallets/flask/blob/a1c478bc93d3dc018a6e7a1ba3cf5409553c9df3/src/flask/app.py#L720-L730

The correct solution would be to regenerate the whole Flask app (`self.app`) instead of just updating the existing one, but I ran into problems with that. So instead, I'm opening an issue for the future and sticking with Flask 2.1.3.

micahflee:
Fixes #1614
Fixes #1542
Fixes #1389

There are a bunch of dependabot PRs that all have merge conflicts. I'm manually updating dependencies so I can close them all.


SaptakS:
Seems like the tests for desktop are failing and then stalling. I will test locally and make some changes tomorrow to get this PR done.

SaptakS:
Seems like `werkzeug.server.shutdown` was deprecated and completely removed in wekzeug 2.1. https://github.com/pallets/werkzeug/issues/1752. I think for the time maybe we need to consider whether we avoid updating flask and werkzeug, however I see both flask and werkzeug has vulneraibility.

There are some alternatives mentioned in the issue of how to shutdow but they sound pretty aweful while going through multiprocessing. @mitsuhiko had raised similar concerns in the issue, but seems like the issue is closed without further discussion. So not really sure.

SaptakS (approved):
This looks good to me. Like discussed, I think some testing to ensure we are getting gzip response and the shutdown works successfully is a good idea to help us catch issues in future easier.

cli/onionshare_cli/web/web.py
```diff
@@ -385,3 +385,15 @@ def cleanup(self):
             dir.cleanup()
 
         self.cleanup_tempdirs = []
+
+    def waitress_custom_shutdown(self):
+        """Shutdown the Waitress server immediately"""
+        # Code borrowed from https://github.com/Pylons/webtest/blob/4b8a3ebf984185ff4fefb31b4d0cf82682e1fcf7/webtest/http.py#L93-L104
+        self.waitress.was_shutdown = True
```
SaptakS, line 392: This `was_shutdown` property is usually used to check for exception in waitress.run part of code (https://github.com/evandroforks/anki/commit/14d4832ce8f4c026fe81c6466d81606c7e5f2d18#diff-9904faac0b4144bb89eaa2c1205913d0b1322b0476d4152c1720255f57d781c5R64-R66) (https://github.com/Pylons/webtest/blob/main/webtest/http.py#L88-L90) . Should we also put `create_server` and `waitress.run()` inside a try-except block? I am not really sure when the exception would happen which the codes are trying to handle. Honestly, if we don't need the try-except block, we might as well remove this line.

SaptakS:
I read a little bit of the waitress source code for the WSGI servers and trigger that get created, and I think it mostly makes sense to me. I am a little iffy about using `_map` since usually a variable named like that means it's an internal variable, but logically its sound since even in their source code, they use `_map` to check for all the open connections.

PR #1679, opened 2023-05-23
Cleanup translations

emmapeel2:
This strings were still on the translations, but are not on the english source file. Weblate has removed them once I installed the 'Clean up translation files'  addon, see here for more info: https://docs.weblate.org/en/latest/admin/addons.html#addon-weblate-cleanup-generic.


SaptakS (approved):
Looks good to me. @emmapeel2 This PR seems to address the cleanup only in the desktop strings. I wonder if there is some cleanup in the docs strings as well.

emmapeel2:
Ok, opened https://github.com/onionshare/onionshare/pull/1684 because weblate had problems to enable the cleanup translation files in doc/install, as there were duplicate strings (although one of them was obsolete).

PR #1692, opened 2023-06-07
Update install.rst

sudwhiwdh:


docs/source/install.rst
```diff
@@ -85,8 +85,6 @@ For Windows::
 
     gpg --verify OnionShare-win64-2.6.msi.asc OnionShare-win64-2.6.msi (Windows 64-bit)
 
-    gpg --verify OnionShare-win32-2.6.msi.asc OnionShare-win32-2.6.msi (Windows 32-bit)
```
sudwhiwdh: In case the 32-bit support will be dropped.

PR #1749, opened 2023-09-05
Getting ready for release 2.6.1 - Part 1

micahflee:
See an extremely long description of this PR here: https://micahflee.com/2023/09/what-goes-into-making-an-onionshare-release/


mig5:
Observations:

1) The issue about generating the 'poetry metadata' for flatpak with version 2.0 of poetry, I recall being fixed via this PR: https://github.com/flatpak/flatpak-builder-tools/pull/353

2) The issue with the 'Temporary failure in name resolution: /simple/poetry-core' which is causing the current failure in Flatpak, seems to have been reported very recently here: https://github.com/python-poetry/poetry/issues/8403 . Though I'm not sure if the root cause is the same for that person as for us (private repos? 'extension modules' ?), but the error is the same. A workaround is mentioned....

SaptakS (approved):
Even though this PR doesn't fully finish the release work as pointed out by @micahflee , I think it already has some essential codebase changes that can be helpful to contributors forking and working off the main branch. Hence I think it's a good idea to get this approved and merged and then finish the rest of release work in a separate PR.

PR #1781, opened 2023-10-16
Prevent disabling files as well as text (or vice versa) in receive mode

mig5:
Resolves https://github.com/onionshare/onionshare/issues/1713


SaptakS (approved):
Looks good, works for me.

PR #1783, opened 2023-10-17
Add trailing slash on directories in website mode to assist with relative asset links

mig5:
This is a replacement for https://github.com/onionshare/onionshare/pull/1656

The issue is that when navigating into subdirectories in website mode, the lack of trailing slash means that a 'relative' link such as a CSS file link, is interpreted relative to the parent dir, and so doesn't load.

To test, before and after:

1) Get a static website like https://github.com/onionshare/onionshare-website . Make 2 copies of it
2) Start a Website mode share, with the two copies of the website as subdirectories
3) Go to the front page of the Website mode share and click on either of the two subdirectories

Before the fix, there is no styling because the css has been (attempted to be) loaded at the top-level dir

After the fix, the subdirectories in the top-level directory listing, have trailing slashes, and this seems to ensure the relative links work as expected.

Website and Share mode share the same code, and the fix would break the Share mode when 'navigating the tree' (the index.html is not rendered in Share mode, and for some reason, with a trailing slash in Share mode, a 404 is rendered). It's for this reason that we add a new boolean arg to be passed down to `directory_listing()` which is only `True` for Website mode (and so adds the trailing slash). Unsure if this is the most elegant fix but it seemed nicer than overriding the `directory_listing()` method entirely for Website mode.

### DIFF
--- a/cli/onionshare_cli/web/send_base_mode.py
+++ b/cli/onionshare_cli/web/send_base_mode.py
@@ -131,7 +131,7 @@ class SendBaseModeWeb:
 
         self.set_file_info_custom(filenames, processed_size_callback)
 
-    def directory_listing(self, filenames, path="", filesystem_path=None):
+    def directory_listing(self, filenames, path="", filesystem_path=None, add_trailing_slash=False):
         # Tell the GUI about the directory listing
         history_id = self.cur_history_id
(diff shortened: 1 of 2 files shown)


SaptakS (approved):
Makes sense to me.

PR #1790, opened 2023-10-21
Raise error from waitress if not shutdown

### LINKED ISSUE #1787: "There was a problem starting the web server" when stopping a server
In the `main` branch, whenever you start a server and then stop it, an error pops up saying: "There was a problem starting the web server"

I think this may have been a regression introduced in one of the recent PRs that was merged.

Screenshot 2023-10-20 at 4 07 12 PM

SaptakS:
Fixes #1787

This seems to be a non-linux issue. I was unable to reproduce the error in linux, but got the error when testing in mac. In mac, when the server is stopped, for some reason (probably a thread waiting) waitress tries the creation function and raises error, which causes the modal to show. When reviewing the waitress PR, I and @mig5 had a similar [discussion](https://github.com/onionshare/onionshare/pull/1677#discussion_r1209821349). I think we might have misunderstood why the code example we followed was using the `was_shutdown` attribute. I think this might be exactly why, since in some cases the create server is called while it's getting shutdown, and in that time the error doesn't need to be raised/shown.

### DIFF
--- a/cli/onionshare_cli/web/web.py
+++ b/cli/onionshare_cli/web/web.py
@@ -365,7 +365,8 @@ class Web:
                 )
                 self.waitress.run()
             except Exception as e:
-                raise WaitressException(f"Error starting Waitress: {e}")
+                if not self.waitress.shutdown:
+                    raise WaitressException(f"Error starting Waitress: {e}")
 
     def stop(self, port):
         """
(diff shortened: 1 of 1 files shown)


mig5 (approved):
Solves it for me on macOS too!

Haven't tried on Windows, but if affected. hopefully sorts it there too.

PR #1865, opened 2024-02-12
Main

emmapeel2:
Solving some more conflicts from weblate.

Proposal: maybe is better to edit only the English .json files here in the repository, and leave the updating of the language files to weblate. This will prevent the .po file conflicts solved by this merge.


SaptakS:
> maybe is better to edit only the English .json files here in the repository, and leave the updating of the language files to weblate.

Makes sense to me.

desktop/onionshare/resources/locale/ga.json
```diff
@@ -137,5 +137,121 @@
...
+    "gui_close_tab_warning_close": "",
+    "gui_quit_warning_title": "",
+    "gui_color_mode_changed_notice": "",
+    "gui_main_page_chat_button": "",
+    "mode_settings_autostop_timer_checkbox": "",
+    "gui_settings_theme_dark": "",
+    "gui_autostop_timer_cant_be_earlier_than_autostart_timer": "",
+    "gui_share_url_public_description": "",
+    "gui_receive_url_public_description": "",
+    "gui_chat_url_public_description": "",
```
emmapeel2, line 154: Here there are some untranslated strings.

Do you want untranslated strings to appear like this?

We can:

- Put the English string when there is no translation
- remove the string completely (the whole line, including the key)
- Leave it as is (lots of changes on every file, even when there is no translation activity)

What do you think is best?

SaptakS (approved):

desktop/onionshare/resources/locale/ga.json
micahflee: Untranslated strings get displayed in English, but you shouldn't have to actually insert the English string here. See here: https://github.com/onionshare/onionshare/blob/main/desktop/onionshare/strings.py#L53

Basically, if a string is blank, then it automatically defaults to displaying English.

emmapeel2: > Untranslated strings get displayed in English, but you shouldn't have to actually insert the English string here. See here: https://github.com/onionshare/onionshare/blob/main/desktop/onionshare/strings.py#L53
> 
> Basically, if a string is blank, then it automatically defaults to displaying English.

OK, I want to **remove the whole line** when the string is not translated, then. This will make the translations less noisy because they will not be updated with empty translations if they are not active.

I understand that it will be ok, because is covered by https://github.com/onionshare/onionshare/blob/a01315295eaf5b2e607ebbddeca04ed034bbbfa4/desktop/onionshare/strings.py#L48  (' and s in translations[current_locale]' ).

I will change the configuration in weblate, and we can see if next weblate merge request passes the tests. If not, I will return to the current setup.

PR #1887, opened 2024-03-06
Use PyQt instead of PySide6

### LINKED ISSUE #1871: Qt6 depdenency & backport to Debian stable / Tails
Hey, 

Tails want to ship a working onionshare version. Unfortunately Debian hasn't made it to package the python Qt6 bindings for the current stable version ( as when the Debian finished the release nearly no software has finished it's Qt6 switch).

As you now switched the current head to QT6 only it makes it impossible to backport a new version for Tails/Debian Stable as the python Qt6 bindings. I looked around and found [PyQt](https://github.com/spyder-ide/qtpy), that is an abstraction layer to Qt, that would allow us to backport the current version to Tails/ Debian stable.

Would this be fine with your side to use this abstraction layer? Me can than prepare a MR.

hefee:
As described in #1871 the hard Qt6 dependency makes it very hard to backport the newer version for Debian. The PyQt abstraction layer makes Debian free to switch Qt6 if everything is packaged correctly.

Currently marked as Draft, as I only made sure that the test suite passes locally.
Closes: #1871


micahflee:
Sorry about taking so long to respond. I'd love to get OnionShare packaged in Debian, but I think we should do two things first:

1) Decide if we want to switch from PySide6 and what the pros (besides Debian support) and cons are, and

2) Much bigger: make sure this works and is packaged properly in Flatpak and Snapcraft in Linux, as well as Windows and macOS. Unfortunately swapping out a dependency like this could take a considerable amount of work for all the platforms we support.

hefee:
I'm a little bit unsure, what you want to do and what I should do. Can you 
clarify?. You are the maintainer, so you need to do the decisions.  I only can 
provide you with information...

> 1) Decide if we want to switch from PySide6 and what the pros (besides
> Debian support) and cons are, and

pros: 
* lightweight
* give maintainer a easy way to switch Qt versions and different Qt backends, 
that should make it more easy to get in more distros

cons:
* an additional abstraction layer
* an additional dependency
 
> 2) Much bigger: make sure this works and is packaged properly in Flatpak and
> Snapcraft in Linux, as well as Windows and macOS. Unfortunately swapping
> out a dependency like this could take a considerable amount of work for all
> the platforms we support.

I don't have much experience with those platforms, but the qtpy has a test 
matrix, that looks very promising:
https://github.com/spyder-ide/qtpy?tab=readme-ov-file#testing-matrix

at least Windows and MacOS should not be an issue.

kushaldas:
PyQt is great if one only wants directly `Qt -> Python` API usage. But,  `Qt for Python` (PySide6 and future versions) also provides various improvements for Python developers using Qt, [example](https://doc.qt.io/qtforpython-6/considerations.html). Older versions of Debian users can still take the benefit using Flatpak. 

I think the project should stay with PySide6.

hefee:
For Debian this issue will be solved for the upcoming stable release (named Trixie), as PySide6 has entered Debian testing.

This patch is still relevant for Tails  as it is based on Debian stable.

mig5:
Sorry for having let this PR hang around so long :( OnionShare suffers badly from human resource issues.

I think that Tails is now on Debian 13 (as of 7.0) and so the dependency issue is resolved either way, due to the sheer time that's passed.

Noting what @kushaldas said, I think we'll close this for now, but we can always revisit.

Thanks @hefee for your efforts in putting this together all the same.

PR #1907, opened 2024-07-09
Remove distutils

21CSM:
The package distutils was depcrecated in Python 3.10 and remove in Python 3.12


21CSM:
I've only updated the ```.lock``` files because I had issues building / testing without doing so (Python 3.10.12).

cli/onionshare_cli/onion.py
```diff
@@ -33,7 +33,7 @@
 import time
 import traceback
 
-from distutils.version import LooseVersion as Version
+from packaging import version as Version
```
hefee, line 36:
```suggestion
from packaging.version import Version
```

your code change is not working, we use the Version class later on, an not the version module.

desktop/onionshare/update_checker.py
```diff
@@ -22,7 +22,7 @@
 import datetime
 import re
 import socks
-from distutils.version import LooseVersion as Version
+from packaging import version as Version
```
hefee, line 25:
```suggestion
from packaging.version import Version
```

hefee:
At least for me on Python 3.12.4, I need the packageing.version.Version class.

hefee (changes requested):

cli/onionshare_cli/onion.py
21CSM: Thanks for the feedback. Addressed and added one more change to this file.

desktop/onionshare/update_checker.py
21CSM: Thanks. Addressed.

PR #1932, opened 2024-10-09
Added optional filename logging for individual file downloads

### LINKED ISSUE #1911: Add a way to log which file is downloaded when using the CLI
In the GUI, it is possible to see which files are getting downloaded but not in the CLI. This could be simply added has a flag so it can be turned off by default

lamemakes:
Fixes #1911. Adds the `--log-filenames` argument which when combined with `--no-autostop-sharing`, will log the filename of each shared file that gets downloaded.

## Screenshots

*without `--log-filenames` (appearance unchanged)*

*with `--log-filenames`*

------

Being that this is my first contrib here feel free to let me know if anything should be changed/is missing! Cheers!

### DIFF
--- a/cli/onionshare_cli/__init__.py
+++ b/cli/onionshare_cli/__init__.py
@@ -120,6 +120,13 @@ def main(cwd=None):
         help="Share files: Continue sharing after files have been sent (default is to stop sharing)",
     )
     parser.add_argument(
+        "--log-filenames",
+        action="store_true",
+        dest="log_filenames",
+        default=False,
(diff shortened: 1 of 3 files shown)


lamemakes:
@mig5 any critiques for this PR? happy to make any changes as needed!

lamemakes:
will also loop in @SaptakS

Slai1994berlin (approved):

Slai1994berlin (approved):

Slai1994berlin (approved):

PR #1935, opened 2024-11-06
add dependency setuptools

dotlambda:
It is needed for pkg_resources: https://github.com/onionshare/onionshare/blob/8102c15376f15cc0f82f7d7e8cd39522b43f1494/desktop/onionshare/gui_common.py#L23


dotlambda:
> it's in the `poetry.lock` file

Not everybody uses the lock file. Running `pip install onionshare_cli` won't install setuptools.

PR #1957, opened 2025-02-06
Dependency update and rejig Flatpak to install pluggable transports via our existing scripts, like other packaging/OS

### LINKED ISSUE #1870: Improve flatpak go and python module generators
The [flatpak builder tools](https://github.com/micahflee/flatpak-builder-tools/tree/fix-go/go) that we use right now to generate the dependency modules for our flatpak yaml are either failing or very hacky. We should work on fixing or writing our own scripts to make this process easier and smoother for the deploy.

### LINKED ISSUE #1894: Snap failed to start under ubuntu 24.04
this is what I got in terminal:

libEGL fatal: DRI driver not from this Mesa build (`'23.0.4-0ubuntu1~22.04.1'` vs `'23.2.1-1ubuntu3.1~22.04.2'`)

mig5:
Fixes #1870
Fixes #1894

~~OK, I had to tweak a few things through trial and error for building the pluggable transports in Flatpak via our own scripts, and now it's happy. Mainly, we needed to use `GO111MODULE=on` and ensure networking was available (to fetch dependencies), then I had to add `--keep-build-dirs` because otherwise during the go build it tries to delete some folders/files that it can't (I think because it's on a readonly fuse filesystem for the build)~~

~~Question is, are *we* happy with this approach? I know it's less standard to do it like this in Flatpak, which encourages network-less for reproducibility reasons, but honestly, we are already doing it this way for all the other builds (macOS and Windows are also executing these build scripts in the same way).~~

~~And not just for CI - that's what any of us would be doing from our workstations already when working on development, or when preparing the official release binaries for macOS/Windows too.~~

~~As far as I can tell, snap is also installing from source too (cloning the repos, checking out the right tag, running a `go build` to make a binary).~~

~~So I feel like maybe it's ok to make Flatpak the same as the others? And it's a lot simpler this way? But maybe the upstream FlatHub isn't happy with this approach? If so, I'll close this out..~~

As per discussions below:

1) We fixed Flatpak to build correctly from sources, and scripted the new process
2) We fixed snap to also build from hashes/sources using `poetry export` to generate requirements.txt files. This new process is also documented in RELEASE.md


SaptakS:
I personally don't mind much and this seems to make the flatpak build simpler, so that's nice. @micahflee thoughts?

SaptakS (approved):

PR #1963, opened 2025-02-10
Make it possible to automatically start a persistent share when OnionShare starts

### LINKED ISSUE #1574: Start sharing on launch
This is a great resource and so easy to use compared to other offerings. Thanks.

I think it could be improved by adding a setting to automatically start sharing folders when launched. That would take care of situations where the server crashed and rebooted while nobody was present.

mig5:
Fixes #1574

This was easier than I thought (famous last words?)

Create a 'persistent' tab and there'll be another checkbox shown to automatically start the service when OnionShare starts.

I made sure that it only tries to automatically start once Tor is connected.

Seems to work great.


SaptakS (approved):
Working for me.

PR #1978, opened 2025-02-12
Try to gracefully fall back to gevent-less SocketIO if it is not supported (#1510)

### LINKED ISSUE #1510: Tests fail for onionshare-desktop on openSUSE Tumbleweed / Invalid async_mode specified (affects ArchLinux too)
Hi, building the latest onionshare (2.5) for openSUSE Tumbleweed I ran in a couple of 'Assertion Errors', see attached.

[testresults.txt](https://github.com/onionshare/onionshare/files/7896968/testresults.txt)

Any suggestion?
Thanks!

mig5:
Fixes #1510 (I hope) - hard to test personally, but the issue sounds pretty straightforward: a ValueError is thrown if there is no gevent for SocketIO's async mode. 

Affected downstream OS have been patching it to remove the async_mode and this apparently works for them.

This shouldn't affect systems that do support the gevent async mode already. Chat is working ok for me on Debian 12 with this patch.

### DIFF
--- a/cli/onionshare_cli/web/web.py
+++ b/cli/onionshare_cli/web/web.py
@@ -176,11 +176,17 @@ class Web:
             self.website_mode = WebsiteModeWeb(self.common, self)
         elif self.mode == "chat":
             if self.common.verbose:
-                self.socketio = SocketIO(
-                    async_mode="gevent", logger=True, engineio_logger=True
-                )
+                try:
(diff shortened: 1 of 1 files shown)


SaptakS (approved):
Makes sense to me. Without specifying async_mode, it will still try the different async modes but will gracefully fall back.

PR #1979, opened 2025-02-13
Properly set the web 'done' state to True on loading the directory listing

mig5:
The share sometimes does not stop as expected in 'auto-stop timer' mode when the timer runs out, in very specific conditions:

1) CLI tool must be used (the desktop version is not affected)
2) Auto-stop-timer must be set (it's not an issue for 'stop after files are downloaded')
3) Someone must've visited the onion at least once (if no-one has ever visited, it would still stop when the timer runs out).
4) Only the 'Share' mode is affected (not receive, website, or chat)

This PR fixes the issue.


SaptakS:
@mig5 which issue is this PR fixing?

PR #1981, opened 2025-02-13
Replace use of pkg_resources.resource_filename with importlib.resources

### LINKED ISSUE #1980: pkg_resources is deprecated in python 3.12 - use of resource_filename
Use of pkg_resources is deprecated in favor of importlib.resources,
importlib.metadata and their backports (importlib_resources, importlib_metadata).
Some useful APIs are also provided by packaging (e.g. requirements and version
parsing). Users should refrain from new usage of pkg_resources and should work
to port to importlib-based solutions.

Python 3.12 in Debian/unstable provides both importlib_resources and
importlib_metadata, so no additional dependencies on those packages are needed.

the importlib_resources have created a guide how to replace resource_filename:
https://github.com/python/importlib_resources/blob/main/docs/migration.rst#pkg_resourcesresource_filename

Tracked in Debian: https://bugs.debian.org/1083519

mig5:
Fixes #1980

Seems we already are importing importlib in poetry so I don't think we need to do anything else..

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.9"
+          python-version: "3.10"
       - name: Install dependencies
         run: |
           sudo apt-get update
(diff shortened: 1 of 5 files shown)


SaptakS (approved):
Looks good to me. `importlib` is part of the cpython libraries, so we don't need to install anything.

PR #1969, opened 2025-02-12
Prepare 2.6.3

mig5:
Fixes #1914
Fixes #1990

~~Note - this is branched off of `mig/update-setuptools` from #1968. I'll rebase this once that PR gets merged.~~

PR #2001, opened 2025-02-19
Remove zh_Hans po files

mig5:
I don't really understand why, but in #1969 after I had run `poetry run ./build.sh`, it had created the LC_MESSAGES and po files for zh_Hans, apparently for the first time.

After it was merged, Weblate sent us notifications saying:

```
## Duplicated translation.
    
The component contains several translation files mapped to a single language
in Weblate. Please fix this by removing one of the translation files.

The following occurrences were found:

| Language                        | Language codes | File names                                                                                   |
|---------------------------------|----------------|----------------------------------------------------------------------------------------------|
| Chinese (Simplified Han script) | zh_Hans, zh_CN | docs/source/locale/zh_Hans/LC_MESSAGES/help.po, docs/source/locale/zh_CN/LC_MESSAGES/help.po |


[View all alerts](https://hosted.weblate.org/projects/onionshare/doc-help/#alerts)
```

I don't quite understand, maybe @emmapeel2 can help explain.

For now, I'm deleting those files to resolve it, but I'm not sure if it's the right thing to do.


emmapeel2:
> The following occurrences were found:
> 
> | Language                        | Language codes | File names                                                                                   |
> |---------------------------------|----------------|----------------------------------------------------------------------------------------------|
> | Chinese (Simplified Han script) | zh_Hans, zh_CN | docs/source/locale/zh_Hans/LC_MESSAGES/help.po, docs/source/locale/zh_CN/LC_MESSAGES/help.po |
> 
> 
> I don't quite understand, maybe @emmapeel2 can help explain.
> 
> For now, I'm deleting those files to resolve it, but I'm not sure if it's the right thing to do.

Yes, this two language codes (zh_Hans is Simplified Chinese, which is used in China, and zh_CN with CN as China country code) are the same language. 

Some applications use one name, some other. For the Tor Browser we are forced to use the country locale because that is how Firefox does it, for example.

We cannot have two files for the same language.

We should find out why poetry created the new language. It would be a pity to lose translations. Is the zh_Hans being updated while the zh_CN is still old?

Maybe we can git mv the files to the new location if that name is preferred? But we cannot keep the two names.

emmapeel2:
I think you need to replace zh_Hans with zh_CN in the configuration files, as that is the filename we are using.

PR #2004, opened 2025-02-21
Replace zh_Hans with zh_CN for docs as suggested by Weblate team

mig5:
I still don't quite understand, but as per recommendation here: https://github.com/onionshare/onionshare/pull/2001#issuecomment-2673552557


SaptakS (approved):
Based on comment by @emmapeel2 approving this PR

PR #2010, opened 2025-02-27
Build Mac ARM64 with Github Actions and then merge it with Intel to make the universal bundle

### LINKED ISSUE #2008: Build ARM64 MacOS app bundle via CI
`macos-latest` now builds on ARM64. To build on AMD64 we need to run the `macos-13` image.

We should run both, and then see if we can finish by automating the universal merge of the two artifacts.

Then it would be just a matter of codesign, package, notarize, staple.

mig5:
Fixes #2008

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -206,7 +206,7 @@ jobs:
       - name: Install cx_Freeze/PySide6 build dependencies
         run: |
           brew install libiodbc
-          curl -O -L https://github.com/PostgresApp/PostgresApp/releases/download/v2.6.5/Postgres-2.6.5-14.dmg --output ~/Downloads/Postgres-2.6.5-14.dmg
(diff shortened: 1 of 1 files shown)


mig5:
@micahflee @SaptakS ~~I wonder in fact if it is just time to sunset the intel mac build? It doesn't even seem to start (as an action) - see https://github.com/onionshare/onionshare/pull/2075~~ Cancel that, I guess we just needed to update to macos-15-intel or macos-26-intel - see [latest labels](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories)

SaptakS:
Seems like there was no conflict after rebasing. I pushed again to see if with the latest dependency updates PR, hopefully flatpak will build successfully here.

mig5:
@SaptakS small syntax issue from me in the Python version change / SHAsum check - I've pushed a fix, hopefully the macOS universal build succeeds..

SaptakS (approved):

PR #2014, opened 2025-03-04
Adds Download Mode, for downloading from another OnionShare

### LINKED ISSUE #365: Receive files from an OnionShare URL in the client
This is different from the receiver mode described in #97.

It would be nice for OnionShare to be a full-featured, end-to-end file-sending application. Users could send and receive files by sharing .onion URLs with each other, without ever having to open and perform any navigation using the Tor Browser. An OnionShare URL may be entered into a new part of the interface, and then the file will be downloaded and saved locally over Tor.

I haven't delved into the code yet, so I don't know which Python libraries are in use (I assume it's stem, etc.), but the implementation is very simple. You'd just be using the Tor SOCKS proxy to make an HTTP GET request with urllib or requests, and save it to `~/Downloads` or a custom location. 

From what I hear on Twitter, there is interest in this feature.

mig5:
Closes #365 . (Sorry, @ageis, that you're not around to see it :( )

To test: try starting an OnionShare in 'Share' mode. Then in another tab, use Download Mode to download it. 

Try with or without private key Shares, both should work.

'Polling' to re-download the share should overwrite the original file from the first download during that poll cycle (as opposed to save to a separate dir each time).

Note: only the `/download` route is supported (it doesn't support downloading individual files from a share)


mig5:
@SaptakS I've made some changes to the front page mode chooser to accomodate the presence of a 5th mode. I think it looks better than it did.

Build jobs are still gonna fail until https://github.com/onionshare/onionshare/pull/2075 gets merged in, then I could rebase.

mig5:
@SaptakS I gave this a bit more testing on the weekend and improved/hardened a few things related to Download mode.

Would love if you or @micahflee could give it a try - set up a share and then try and download it in the Download tab.

I'm thinking of adding an 'Upload' mode too that acts as the compliment to 'Receive' mode (upload files to a Receive mode OnionShare, from within the client??) What do you think? We'd also then have 'symmetry' on the front page mode chooser again, too.

(I wonder if we could then get to the 'Chat mode' in OnionShare being an actual Chat client but that might be too tricky)

Whilst working on it I also ended up fixing the wrapping/widget cutoff issue described in https://github.com/onionshare/onionshare/pull/2055 .

PR #2020, opened 2025-03-06
Use daemon = True boolean instead of setter setDaemon in CLI AutoStop thread

### LINKED ISSUE #2018: `threading.Thread.setDaemon` is deprecated
- [x] I have already searched the issue queue for my error message or symptom

Operating system: Linux
Version/distribution: Debian sid

Version of Onionshare: 2.6.3

- [x] Via my Linux distribution's package manager

## What is the problem

```
cli/onionshare_cli/common.py:617: DeprecationWarning: setDaemon() is deprecated, set the daemon attribute instead
    self.setDaemon(True)   
```

see https://docs.python.org/3/library/threading.html#threading.Thread.setDaemon

mig5:
Fixes #2018

### DIFF
--- a/cli/onionshare_cli/common.py
+++ b/cli/onionshare_cli/common.py
@@ -614,7 +614,7 @@ class AutoStopTimer(threading.Thread):
 
         self.common = common
 
-        self.setDaemon(True)
+        self.daemon = True
         self.time = time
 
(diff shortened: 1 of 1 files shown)


SaptakS (approved):
Makes sense to me.

PR #2030, opened 2025-04-01
Update get-tor.py script to use platform-specific json files

### LINKED ISSUE #2029: Update the get-tor.py script to use different json files per platform
CI is failing because Tor Project changed their json files for the Tor Browser info. It's now a separate json file per-platform (except that Windows gets to still be the exclusive platform in the `downloads.json`, for some reason)

We need to update the script.

Ref: https://forum.torproject.org/t/aus1-torproject-org-torbrowser-update-3-release-downloads-json-now-lists-only-windows-platform-links/18167/4

mig5:
Closes #2029


SaptakS (approved):
Looks good to me.

PR #2055, opened 2025-10-21
Increase minimum window height to 800 pixels

FacundoAcevedo:
Fix overlapping text on the Receive tab while the dialog box is opened.

With the previos height of 700 I get some texts overlapped and the Address box cannot display the address correctly:
image

### DIFF
--- a/desktop/onionshare/main_window.py
+++ b/desktop/onionshare/main_window.py
@@ -46,7 +46,7 @@ class MainWindow(QtWidgets.QMainWindow):
 
         # Initialize the window
         self.setMinimumWidth(1040)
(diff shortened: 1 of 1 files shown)


mig5:
Thanks @FacundoAcevedo . I have ended up fixing this a different way without needing to increase the size of the app altogether (we've heard complaints from other users that the app window is already very large and problematic on small screens). 

Instead I'm wrapping text and also hiding the image when the History pane gets opened. 

This fix is part of a separate PR which will probably get merged soon. https://github.com/onionshare/onionshare/pull/2014.

Appreciate your contribution nonetheless! Thanks for trying to improve OnionShare!

PR #2066, opened 2026-03-14
receive mode: show indeterminate progress and warn user during Tor traversal

### LINKED ISSUE #1876: Improvements to the receive mode web frontend
When sending relatively big files, the receive mode web frontend (i.e. the frontend that the sender of the file sees) has some issues which I feel can be improved.

signal-2024-02-22-232146_002

In the above screenshot, we can see that the sending of the file to the person running receive mode hasn't completed since it's traversing through the Tor network. However, the progress bar is complete. We should probably better indicate that the sending is still in progress and also mention not to close the Tor Browser till it's fully sent.

Gerijacki:
Fixes #1876

## What changed

When a file upload reaches **100% in the browser**, the data has only reached the **first Tor hop**. It still needs to traverse the rest of the onion circuit before arriving at the receiver.

Previously, the progress bar remained at **100%** (visually appearing “done”), which could mislead users into closing the tab before the transfer actually completed.

## Fix

* Switch the `<progress>` element to **indeterminate** (remove the `value` attribute) once all bytes leave the browser. This makes the bar animate as “still working” instead of appearing finished.
* Show a warning message:
  *“Do not close this tab until the submission is complete.”*
* Clarify the status text to reflect that the transfer is still in progress.

## Testing

* `cd cli && poetry run pytest tests/ -q` — **248 passed**
* Manual test:

  1. Open receive mode.
  2. Upload a large file over a slow connection.
  3. Verify the progress bar switches to **indeterminate** and the warning appears after the local send completes.

### DIFF
--- a/cli/onionshare_cli/resources/static/css/style.css
+++ b/cli/onionshare_cli/resources/static/css/style.css
@@ -369,6 +369,13 @@ div#uploads .upload progress {
   height: 20px;
 }
 
+div#uploads .upload .upload-warning {
+  color: #c97000;
+  font-size: 0.85em;
+  font-weight: bold;
+  margin-top: 4px;
+}
+
 ul.flashes {
   list-style: none;
   margin: 0;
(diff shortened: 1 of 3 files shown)


mig5 (approved):
Works great, thanks!

PR #2087, opened 2026-07-05
Store temporary files under XDG cache dir on Linux/BSD

### LINKED ISSUE #2052: Cache data should be stored in ~/.cache/onionshare per XDG spec
- [X] I have already searched the issue queue for my error message or symptom

Operating system: Linux
Version/distribution: Debian 12 Bookworm

Version of Onionshare: 2.6-5~deb12u1

- [X ] Via my Linux distribution's package manager

## What is the problem

Currently, OnionShare stores cach- and config data under ~/.config/onionshare.

According to the XDG Base Directory Specification, cache data should be splitted into
* ~/.config/onionshare
* ~/.cache/onionshare

## Anything else you'd like to tell us?

As an example:

The directory `~/.config/onionshare/tmp` is not cleaned up properly after exiting.
It should take place to `~/.cache/onionshare/tmp`

citizen204:
## Summary

OnionShare currently keeps its `tmp` working directory inside the config directory (`~/.config/onionshare/tmp`). Per the XDG Base Directory Specification, disposable cache data belongs in `$XDG_CACHE_HOME` (defaulting to `~/.cache`). This PR adds a `build_cache_dir()` helper that resolves to `~/.cache/onionshare` on Linux/BSD (honoring `$XDG_CACHE_HOME`), and moves the temporary files directory there. On Windows and macOS, cache data stays inside the existing platform data directory, so nothing changes on those platforms.

All temporary-file consumers (`onion.py`, `share_mode.py`, `send_base_mode.py`, `moat_dialog.py`) already go through `common.build_tmp_dir()`, so this single change covers them. The test-mode `-testdata` suffix behavior is preserved.

Fixes #2052

## Changes

- `cli/onionshare_cli/common.py`: add `build_cache_dir()` and base `build_tmp_dir()` on it

### DIFF
--- a/cli/onionshare_cli/common.py
+++ b/cli/onionshare_cli/common.py
@@ -430,11 +430,34 @@ class Common:
         os.makedirs(onionshare_data_dir, 0o700, True)
         return onionshare_data_dir
 
+    def build_cache_dir(self):
+        """
+        Returns the path of the OnionShare cache directory, which holds
+        non-essential data that can safely be deleted. On Linux and BSD this
+        follows the XDG Base Directory Specification (~/.cache/onionshare);
+        on Windows and macOS cache data lives inside the data directory.
+        """
+        if self.platform == "Windows" or self.platform == "Darwin":
+            onionshare_cache_dir = os.path.join(self.build_data_dir(), "cache")
+        else:
+            try:
+                xdg_cache_home = os.environ["XDG_CACHE_HOME"]
+                onionshare_cache_dir = f"{xdg_cache_home}/onionshare"
+            except Exception:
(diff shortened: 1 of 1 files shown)


mig5:
Thanks. A few issues with this one:

1. If the `XDG_CACHE_DIR` env var is set to an empty string, the resulting cache dir would be `/onionshare` which probably would result in a permission error
2. You said 'nothing changes' for Mac or Windows platform, but in fact you've moved their tmp data into a new 'cache' subfolder:
```
Before: <data directory>/tmp
After:  <data directory>/cache/tmp
```

I'm going to push a few tweaks to your branch and then merge it in.

mig5 (approved):

PR #2090, opened 2026-07-11
Warn when Receive Mode save folder is not writable

### LINKED ISSUE #2062: OnionShare should warn when the user selects a non-writable directory
- [x] I have already searched the issue queue for my error message or symptom

Operating system: Linux
Version/distribution: Debian Trixie, or Tails 7.4.2

Version of Onionshare: 2.6.3

- [ ] Native Mac/Windows package
- [ ] Flatpak
- [ ] Snap
- [ ] Pip (command line)
- [x] Via my Linux distribution's package manager
- [ ] From source

## What is the problem

When I open "Receive Files", I can select a non-writable directory (ie: `/root/`). OnionShare doesn't show any error, and allows me to start Receive Mode.
Of course, receiving files will fail whenever files are actually received.

Now, you could think that selecting `/root/` is not what any real user would do, but there are some cases in which this is less surprising. At [Tails](https://tails.net), we restrict what directories OnionShare has access to (using AppArmor), and limit it to `Downloads/`, which means that if a user selects another directory, they will suffer from this bug. See https://gitlab.tails.boum.org/tails/tails/-/issues/21425

Pitchfork-and-Torch:
## Summary

Fixes #2062.

Receive Mode allowed selecting a non-writable folder (e.g. `/root` under Tails AppArmor) and starting with no warning. Uploads then failed only when files arrived.

### Changes
- Probe writability with a real create/delete test (better than `os.access` alone for sandboxes)
- Reject non-writable folders when browsing
- Block **Start Receive Mode** until a writable folder is chosen
- English string `gui_receive_data_dir_not_writable`
- Unit tests for the probe helper

### Testing
- Unit tests for writable dir / missing path / file path
- Manual GUI check recommended on Linux/Tails if available

### DIFF
--- a/desktop/onionshare/tab/mode/receive_mode/__init__.py
+++ b/desktop/onionshare/tab/mode/receive_mode/__init__.py
@@ -209,6 +209,25 @@ class ReceiveMode(Mode):
         """
         return "receive"
 
+    @staticmethod
+    def is_data_dir_writable(path):
+        """
+        Return True if path is an existing directory we can create files in.
+
+        Uses a real write probe so AppArmor/sandbox restrictions (e.g. Tails)
+        are detected even when os.access is optimistic.
+        """
+        if not path or not os.path.isdir(path):
+            return False
+        try:
+            test_path = os.path.join(path, ".onionshare-write-test")
+            with open(test_path, "w", encoding="utf-8") as f:
+                f.write("ok")
+            os.remove(test_path)
+            return True
+        except OSError:
+            return False
+
     def data_dir_button_clicked(self):
         """
         Browse for a new OnionShare data directory, and save to tab settings
@@ -224,6 +243,14 @@ class ReceiveMode(Mode):
                 if not selected_dir.startswith(os.path.expanduser("~/OnionShare")):
                     Alert(self.common, strings._("gui_receive_flatpak_data_dir"))
                     return
+
+            if not self.is_data_dir_writable(selected_dir):
+                Alert(
+                    self.common,
+                    strings._("gui_receive_data_dir_not_writable").format(selected_dir),
+                    QtWidgets.QMessageBox.Warning,
+                )
+                return
 
             self.common.log(
                 "ReceiveMode",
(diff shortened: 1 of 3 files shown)


Pitchfork-and-Torch:
Friendly bump: this is still green on win64/mac/flatpak/snap/CodeQL + test-cli. Happy to adjust anything for maintainers if preferred.

Windows note: the writable-folder check helps when Receive Mode points at a protected or missing path — common failure mode on first setup.

SaptakS:
Hello @Pitchfork-and-Torch . Thanks for the PR and your contribution. I am going to review the changes soon but probably won't make it to the 2.6.5 release.

Pitchfork-and-Torch:
Thanks @SaptakS. No rush on 2.6.5. Happy to wait for the next train, and I will rebase if main moves before review.

desktop/onionshare/tab/mode/receive_mode/__init__.py
```diff
@@ -209,6 +209,25 @@ def get_type(self):
...
+
+        Uses a real write probe so AppArmor/sandbox restrictions (e.g. Tails)
+        are detected even when os.access is optimistic.
+        """
+        if not path or not os.path.isdir(path):
+            return False
+        try:
+            test_path = os.path.join(path, ".onionshare-write-test")
+            with open(test_path, "w", encoding="utf-8") as f:
+                f.write("ok")
```
mig5, line 225: Would you be able to make this a named temp file rather than write to a predictable, static file name? (Not that it's likely to contain any real data, but technically, we're overwriting it, so it's just good practice I think).

Something like this (make sure to add `import tempfile`):

```python
@staticmethod
def is_data_dir_writable(path):
    """Check directory creation, file writing, and cleanup."""
    if not path or not os.path.isdir(path):
        return False

    try:
        with tempfile.TemporaryDirectory(
            prefix=".onionshare-write-test-",
            dir=path,
        ) as probe_dir:
            with tempfile.NamedTemporaryFile(
                mode="wb",
                prefix="probe-",
                dir=probe_dir,
            ) as probe:
                probe.write(b"ok")
                probe.flush()

        return True
    except OSError:
        return False
```

desktop/onionshare/tab/mode/receive_mode/__init__.py
```diff
@@ -299,6 +326,20 @@ def autostop_timer_finished_should_stop_server(self):
             self.web.receive_mode.can_upload = False
             return False
 
+    def start_server(self):
+        """
+        Start Receive Mode only if the save directory is writable.
+        """
+        data_dir = self.settings.get("receive", "data_dir")
+        if not self.is_data_dir_writable(data_dir):
+            Alert(
```
mig5, line 335: Before you call `Alert` (but after you've evaluated `if not self.is_data_dir_writable(data_dir):), can you please call this?

```python
        self.server_status.stop_server_finished()
```

This might help ensure the service can't get stuck in 'Working' state on a non-writable dir.

mig5 (changes requested):

Pitchfork-and-Torch:
Added server_status.stop_server_finished() immediately before the warning Alert, after the writable check. That matches the request so the share does not stay in Working.

PR #2095, opened 2026-08-17
fix: make Test Connection to Tor result strings translatable

### LINKED ISSUE #983: Test Connection to Tor: convert True/False binary statements into strings so they can be translated
A translator pointed out that in the 'Test Connection to Tor', the various features marked as supported or not ('supports ephemeral services' etc) use True/False binary statements. These should be converted into strings so that they can be translated.

> * In the "Test Connection to Tor"» result window we have "true" (and I
guess "false" if the test fails). Both statement should be made
translatable.

citizen204:
## Summary

The "Test Connection to Tor" result dialog (`settings_test_success`) formats three Python booleans (`supports_ephemeral`, `supports_stealth`, `supports_v3_onions`) directly into the message string, so it always shows the literal English words "True"/"False" regardless of the user's locale - these can't be translated since they never go through `strings._()`.

## Changes

- `desktop/onionshare/resources/locale/en.json`: added two new translatable keys, `gui_settings_true` ("Yes") and `gui_settings_false` ("No").
- `desktop/onionshare/tor_settings_tab.py`: in `tor_con_success`, map each boolean through a small `yes_no()` helper that looks up the new translated strings instead of passing the raw Python bool into `.format()`.

Fixes #983


SaptakS (approved):
Thanks for the PR! I think this makes sense to me, especially with using a single set of string for all true/false in settings. Going to approve this and merge.

PR #2099, opened 2026-09-01
Fix crash when passing transports to the Moat circumvention settings API

dylanpulver:
`CensorshipCircumvention.request_settings()` builds its request body as a dict, but when the `transports` argument is supplied it calls `data.append(...)` on it:

```python
data = {}
if country:
    data = {"country": country}
if transports:
    data.append({"transports": transports})   # censorship.py:147
```

`dict` has no `append`, so any call that passes `transports` raises `AttributeError` before `requests.post` runs. `AttributeError` isn't a `RequestException`, so it escapes the `try` and the caller's `except CensorshipCircumventionError` too. The documented parameter has never been usable.

Nothing in the tree passes `transports` today — `connection_tab.py:246` calls with `country=` only — which is why it went unnoticed. `censorship.py` also had no test file at all.

Fix is one line: rdsys' moat spec (`doc/moat.md`, `/circumvention/settings`) says the fields "can be provided individually or as a combination in the same payload", example `{"country": "de", "transports": ["obfs4", "snowflake"]}` — so `transports` belongs in the same dict as `country`.

Verified on `main` before and after: with the change reverted, the two new `transports` tests fail with that exact `AttributeError` while the `country`-only test still passes; with it applied all three pass. Full CLI suite: 260 passed, plus one pre-existing macOS-only failure (`test_get_tor_paths_darwin`, no bundled tor locally) that is identical before and after.

Noted while in the file but deliberately left alone: `__init__` does `return False` at line 57 when `onion` is passed unauthenticated, which would be `TypeError: __init__() should return None`. Both in-repo call sites guard on `is_authenticated` first, so it's unreachable today — happy to fix it here or separately if you'd like.


mig5 (approved):
Thanks!

PR #2103, opened 2026-09-10
fix: make file size units translatable in the desktop GUI

### LINKED ISSUE #986: Make file sizes (MB, GB, etc) translatable
Terms like Mb/MiB, Gb/Gib etc, for file sizes, are different in other languages. These should be translatable strings.

A translator requested that these also be translatable in the Share/Receive web interfaces, but I think we should keep those as English to avoid leaking info, and because there's no way/reason to guarantee that the receiver uses the same language as the sender. But we can do this in the GUI itself.

citizen204:
## Summary

File size units (`B`, `KiB`, `MiB`, …) were hardcoded English in `Common.human_readable_filesize()`, which the desktop GUI used directly, so localized builds still showed English units in the file list, the transfer history, and the progress bars.

Per the scoping in #986, this only changes the **GUI**: the Share/Receive web interfaces keep their English units so the sender's locale is not leaked to the recipient, and `Common.human_readable_filesize()` (used by the CLI and the web interfaces) is unchanged in behaviour.

The size arithmetic now lives in one place: `Common.split_filesize()` returns the scaled value plus a unit index, and both the English `human_readable_filesize()` and the new GUI-side `GuiCommon.get_translated_filesize()` build their string from it. The new locale strings are full format strings (`"{} MiB"`) rather than bare unit labels, so translators can also change the spacing or put the unit before the number where that is the convention.

Fixes #986

## Changes

- `cli/onionshare_cli/common.py`: extracted `split_filesize()`; `human_readable_filesize()` now builds on it and returns exactly the same strings as before.
- `desktop/onionshare/gui_common.py`: new `GuiCommon.get_translated_filesize()`, alongside the existing `get_translated_tor_error()` / `get_translated_web_error()` helpers.
- `desktop/onionshare/resources/locale/en.json`: added `gui_filesize_bytes` … `gui_filesize_yib`.
- `desktop/onionshare/tab/mode/{file_selection,history,share_mode/__init__,website_mode/__init__}.py`: the 11 GUI call sites now use the translated helper. No web/CLI call site was touched.
- `cli/tests/test_cli_common.py`: added `TestSplitFilesize`.

## Verification

- `pytest cli/tests/test_cli_common.py` — 103 passed (the one failure, `TestGetTorPaths::test_get_tor_paths_windows`, is pre-existing on `main` in my environment and unrelated). `TestHumanReadableFilesize` + the new `TestSplitFilesize`: 20 passed.
- Checked the refactor against the old implementation over 200k random sizes plus every boundary in the existing parametrized test — byte-identical output.
- Imported the real `GuiCommon` under `QT_QPA_PLATFORM=offscreen` and confirmed `get_translated_filesize()` returns the identical English strings with `en.json` loaded, and picks up localized units when a translation is present (e.g. `3.0 Mio`, `42.0 o`).

I could not run the full `desktop/tests/test_gui_*.py` suite here (it needs Tor), but those tests make no assertions on file size strings, and the English output is unchanged.


mig5 (approved):
Thanks, seems to work well.

PR #2104, opened 2026-09-10
fix: show auto-start/auto-stop timers in the user's locale format

### LINKED ISSUE #985: Convert timer clocks to 24h time in other locales
We should see if we can use 24 hour clocks for certain timezones (perhaps anything other than English. since 24h time is common in Europe.

> On a system set with a 24 hours clock, the start and stop timers
display time as A.M./P.M. Could this be adapted to a 24 hours clock in
such cases?

citizen204:
## Summary

The auto-start / auto-stop timer widgets and their server-button tooltips used hardcoded format strings (`"hh:mm A MMM d, yy"` and `"h:mm AP, MMMM dd, yyyy"`), so every language got a 12-hour AM/PM clock and English date ordering. Because `QDateTime::toString(format)` also resolves month names in the C locale, the tooltips stayed English regardless of the selected language, and the widget mixed the two: with OnionShare set to French, the auto-stop timer read `02:05 PM mars 9, 27`.

The formats are now derived from a `QLocale` built from the language OnionShare is displayed in, and the timer widgets get that locale set on them so their month and AM/PM names match. Of the 40 languages in `available_locales`, 32 (including French, German, Norwegian, Russian, Japanese, Simplified Chinese, Brazilian Portuguese) now use a 24-hour clock, while English, Greek, Arabic, Bengali, Khmer, Tamil, Albanian and Traditional Chinese keep 12-hour — driven by CLDR data rather than a hardcoded list of "European" locales.

Fixes #985

## Changes

- `desktop/onionshare/gui_common.py`: new `GuiCommon.get_locale(common)`, returning a `QLocale` for the language OnionShare is displayed in.
- `desktop/onionshare/tab/mode/mode_settings_widget.py`: both `QDateTimeEdit`s now get `setLocale()` plus a display format built from `timeFormat(ShortFormat)` + `dateFormat(ShortFormat)`.
- `desktop/onionshare/tab/server_status.py`: new `ServerStatus.timer_datetime_string()` formats the two tooltips via `QLocale.toString()` with `timeFormat(ShortFormat)` + `dateFormat(LongFormat)`.

## Before / after

| locale | widget before | widget after | tooltip after |
| --- | --- | --- | --- |
| en | `02:05 PM Mar 9, 27` | `2:05 PM 3/9/27` | `2:05 PM, Tuesday, March 9, 2027` |
| fr | `02:05 PM mars 9, 27` | `14:05 09/03/2027` | `14:05, mardi 9 mars 2027` |
| de | `02:05 PM März 9, 27` | `14:05 09.03.27` | `14:05, Dienstag, 9. März 2027` |
| ja | `02:05 PM 3月 9, 27` | `14:05 2027/03/09` | `14:05, 2027年3月9日火曜日` |

## Verification

I could not run `desktop/tests/test_gui_*.py` here (they need Tor), so I verified with an offscreen PySide6 reproduction that mirrors the widget construction in `ModeSettingsWidget` exactly (`setLocale` -> `setDisplayFormat` -> `setDateTime` -> `setMinimumDateTime` -> `setCurrentSection(MinuteSection)`):

- every one of the 40 codes in `Settings.available_locales` parses into a real `QLocale` (none fall back to `C`);
- for `en`/`fr`/`de`/`nb_NO`/`ja`/`zh_Hans`/`ru`/`ar`/`pt_BR`, `setCurrentSection(MinuteSection)` still applies and `dateTime()` round-trips the value that was set, so the existing GUI tests (which only use `setDateTime()` / `dateTime()`) are unaffected;
- the new format exposes the same set of sections as the old one (hour, minute, am/pm where applicable, day, month, year — no seconds), so nothing is truncated that was not truncated before.

## Note

I used the language OnionShare is displayed in rather than `QLocale.system()`, since the issue frames this as "24 hour clocks ... perhaps anything other than English", and it also fixes the untranslated month names. If you would rather have the timers follow the OS regional settings instead, `GuiCommon.get_locale()` is the single place to change.


mig5:
Thanks for this change @citizen204 !

While I agree that it's an improvement re: the hour/minute being in 24h, this has the unfortunate effect for non-American English in that the month comes first and then the day, which is uncomfortable/confusing for UK, Australian, NZ users (basically anywhere other than America).

I wonder if there is a way we can retain the fix for the 24h clock but not change the month / day / year in the widget?

The alternative is supporting different types of English, I guess, but that adds more work as well (I'm not sure we want to add an en_AU, en_GB etc to Weblate for translations, but maybe we do? @micahflee @SaptakS any thoughts?).

PR #2105, opened 2026-09-12
fix: make Alert dialog standard buttons (OK, Cancel, etc) translatable

### LINKED ISSUE #984: Make 'OK' and other Qt buttons translatable
A translator pointed out that the 'OK' (and perhaps other words) in Qt buttons, could be made translatable.

> Could the OK used for the OK button (to close a window) be pushed to
Weblate since not all languages will want to use the English term.

citizen204:
## Summary

A translator pointed out that the "OK" button (and other standard Qt dialog buttons) don't get translated. Root cause: OnionShare never installs a `QTranslator` for Qt's own `qtbase` locale files, so `QMessageBox` standard buttons (`QtWidgets.QMessageBox.Ok`, etc.) always render using Qt's compiled-in English default, regardless of the OnionShare locale the user has selected in Settings. Nearly every `Alert()` dialog in the app (~10 call sites) uses the default `buttons=QtWidgets.QMessageBox.Ok`, so this affected almost every alert box.

## Changes

- `desktop/onionshare/widgets.py`: `Alert.__init__` now re-labels each standard button actually present on the dialog (`Ok`, `Cancel`, `Yes`, `No`, `Close`) using OnionShare's own locale strings via `self.button(standard_button).setText(...)`. Dialogs that pass `buttons=QtWidgets.QMessageBox.NoButton` and add their own `QPushButton`s (already translated at the call site, e.g. `MainWindow.tor_connection_canceled`) are untouched.
- `desktop/onionshare/resources/locale/en.json`: added `gui_alert_button_ok`/`_cancel`/`_yes`/`_no`/`_close` keys (English defaults matching Qt's existing text, so nothing changes for English users) for Weblate to translate into other languages.
- `desktop/tests/test_widgets.py`: new regression tests verifying standard buttons pick up a translated string while custom (non-standard) buttons are left alone.

Verified with an offscreen PySide6 `QApplication` (`QT_QPA_PLATFORM=offscreen`) since the full desktop test suite needs Tor: confirmed the button text renders as "OK" today, and updates correctly once a translation is supplied for the new keys. `git stash`/`stash pop` confirms the new tests fail on current `main` and pass with this change.

Fixes #984


mig5 (approved):
Thanks for this! I added the widget test script to the CI workflow as well.

PR #2106, opened 2026-09-12
fix: detect OS dark mode via QStyleHints.colorScheme() instead of a bare QPalette

### LINKED ISSUE #1440: Improve automatic detection of system theme
image

Currently, in auto theme settings, OnionShare uses light by default even if the system color mode is Dark.

For detecting the system dark mode, I think we can use winreg in the Python library for Windows like DarkDetect

https://github.com/albertosottile/darkdetect

I have tested darkdetect on Linux and Windows and it detects system color mode properly. 

image
image

citizen204:
## Summary

With theme set to "Automatic", OnionShare is supposed to follow the OS light/dark setting, but `Application.is_dark_mode()` checked the `Base` color of a fresh, default-constructed `QPalette()`. That palette reflects the current *widget style*, not the platform theme: on Windows and macOS the app calls `self.setStyle("Fusion")` immediately before this check runs (`__init__.py`'s `Application.__init__`), and Fusion applies its own default palette which does not track the OS's dark/light setting — so auto-detection was effectively always light on the two platforms that actually call `setStyle`.

The issue's own suggestion was to add the third-party `darkdetect` package, but Qt already ships an equivalent, purpose-built API: `QStyleHints.colorScheme()` (added in Qt 6.5) asks the platform theme directly and isn't affected by which widget style is active. OnionShare pins `PySide6 = "6.8.2.1"` (`desktop/pyproject.toml`), well above that floor, so no new dependency is needed.

Verified live on this Windows machine (currently OS light mode, `HKCU\...\Personalize\AppsUseLightTheme = 1`): `app.styleHints().colorScheme()` correctly reports `ColorScheme.Light`, matching the registry ground truth, both before and after `setStyle("Fusion")` is applied. I wasn't able to also exercise the Dark branch against a live OS setting without changing this machine's system theme, so that direction is covered by a mocked regression test instead (see below) plus reasoning about why the old code path was broken.

## Changes

- `desktop/onionshare/__init__.py`: `Application.is_dark_mode()` now returns `self.styleHints().colorScheme() == Qt.ColorScheme.Dark` instead of comparing a bare `QPalette()`'s `Base` color to white.
- `desktop/tests/test_application_color_mode.py`: new regression tests that mock `styleHints()` to return `Dark`/`Light`/`Unknown` and assert `is_dark_mode()` follows it. Fails on current `main` (mocking `Dark` still returns `False`), passes with this change.

Fixes #1440


mig5 (approved):
Thanks!

PR #2107, opened 2026-09-13
fix: make Tor bootstrap connection status messages translatable

### LINKED ISSUE #982: Try and get Tor controller messages to use user's locale
A translator pointed out that the 'Connecting to Tor' and other Tor controller messages, appear in English regardless of the user's chosen locale.

Given that Tor itself is translated, we should see if there's some way with which we can influence the Tor messages to use the same locale?

citizen204:
## Summary

The "Connecting to the Tor network" progress dialog shows Tor's own bootstrap-phase summary text (e.g. "Establishing a Tor circuit", "Loading relay descriptors", "Done") verbatim from the Tor control port, which is always in English, so it never respects the user's chosen OnionShare locale.

Tor's control protocol reports a stable per-phase "tag" alongside each human-readable summary (see [control-spec.txt Section 5.5](https://spec.torproject.org/control-spec/replies.html#bootstrap-phases-and-mnemonics)), so this passes that tag through to the status callback and uses it to look up a translated string, falling back to Tor's own summary text for any tag OnionShare doesn't recognize (e.g. one added by a future Tor version) — so nothing can regress to a crash or blank label if the mapping is ever incomplete.

Fixes #982

## Changes

- `cli/onionshare_cli/onion.py`: extract the `TAG=` field from the bootstrap-phase control response and pass it as a third argument to `tor_status_update_func`.
- `desktop/onionshare/tor_connection.py`: thread the new `tag` argument through `TorConnectionThread`'s signal, and use it in `TorConnectionWidget._tor_status_update` to fetch a translated summary via `GuiCommon`.
- `desktop/onionshare/gui_common.py`: add `get_translated_bootstrap_summary(tag, fallback)`, a small static helper following the same pattern as the existing `get_translated_tor_error`/`get_translated_web_error`.
- `desktop/onionshare/resources/locale/en.json`: add `gui_bootstrap_status_*` keys for the 25 current bootstrap phases documented in control-spec.txt Section 5.5.2–5.5.4 (covers both the plain and pluggable-transport/proxy variants of each stage), so they become available to translators in Weblate.
- `desktop/tests/test_gui_common.py`: new regression test verifying a known tag is translated and an unknown tag safely falls back to the raw summary.

## Verification

- Verified live with an offscreen import of the real `en.json` strings: known tags translate correctly, and an unmapped tag falls back to the raw fallback text with no exception.
- New `desktop/tests/test_gui_common.py` passes; confirmed it fails without the fix via `git stash`.
- `cd cli && python -m pytest tests/test_cli_common.py`: 103 passed, 2 skipped, with the one pre-existing unrelated Windows failure (`TestGetTorPaths::test_get_tor_paths_windows`, fails on clean upstream `main` too in this environment).
- Only the CLI/web-facing messages stay in English by design, per the scope the maintainer already set for a related issue (#986) — this PR only affects the desktop GUI's Tor-connection dialog.


mig5:
I'm in two minds about this one and whether we should `#wontfix` it. My main concern is that if upstream Tor changes those strings, these `tag` matches won't match anymore. I just wonder if we're making work for ourselves. What do you think @micahflee @SaptakS

